PR #1, opened 2022-06-02
Add slices for dotnet dependencies

woky:


slices/ca-certificates.yaml
```diff
@@ -0,0 +1,7 @@
+package: ca-certificates
+
+slices:
+  all:
```
cjdcordeiro, line 4: pls note that this isn't actually `all`. Things like `usr/share/man` and `usr/share/doc` are not included here.

I see that `base-files` also uses the same slice name `all`, and I'm wondering whether this won't become misleading and prone to errors in the future. Does this justify having a special keyword for "all slices"? Such that we don't need to explicitly have to maintain this `all` slice as the pkg evolves ...

woky: Agreed that `all` is not best words. I think it's better candidate for reserved slice name that signifies *all files in this package*. @cjdcordeiro How does `certificates` sound as slice name in this case?

> Does this justify having a special keyword for "all slices"?

Sounds reasonable to me. But that's different from `all` I described above. I think both would be useful.

niemeyer: Agree with earlier comments above: the problem with "all" is that it's easy to not be true, and even if it is true today, it may not be anymore when the package grows. There may be an opportunity here for us to create a strong convention which means "the standard set of things that come when you want this package". Something along the lines of "default", but that word still feels a bit too much like "all" in the sense that in theory the "default" may be seen as the upstream package choice's, and it's not clear what to put into "default" vs. what to put elsewhere. So every time we pick a "default" slice, we indeed have no clue about what is coming.

So, I guess the ideal convention for us to create is the one where not only we know what name to pick, but we know what likely comes inside it. For that, we might start with a set of names that are self explanatory. For example:

  - _pkg_data =>_ All data files from that package
  - _pkg_lib =>_ All visible libraries from that package
  - _pkg_bin =>_ All visible executables from that package (/bin/*, /usr/bin/*, etc). Most likely depends on _libs, may depend on _data too depending on functionality.
  - _pkg_doc =>_ All documentation from that package. Probably doesn't depend or is a dependency of anything.

Of course, these are stock conventions for simple cases, which don't prevent more specific names from existing.

If we take that on, the answer to the original name might be _ca-certificates_data._

cjdcordeiro: `pkg_data` sounds fine to me.

Also the proposed conventions make sense. In fact that's what we've been using, even though I had been using the plural (maybe @woky you'll want to make them all singular instead - i.e. `libc6_libs` vs `libc6_lib`)

@niemeyer wrt to `all`, I'm wondering if `chisel` could accept pkgs names without the `_sliceName`. Example: `chisel cut ... pkgX` would install all slices in pkgX

niemeyer: (small update: made _bins and _libs plural)

cjdcordeiro: (I've edited my previous comment wrt the plural vs singular slice names ;) )

yes having an explicit name (like `default`) for it sounds better. not critical for now, we can move on without it and keep thinking.

@woky I'd say _keep the `_all` for now until a better convention comes along_

niemeyer: Can we please not use "_all" for the reasons described above?

cjdcordeiro: sorry, I meant `_data` (corrected above)

woky:
I renamed `ca-certificates_{all => data}` and `base-files_{all => base}`.

cjdcordeiro (approved):
it lgtm to me. 

the following example worked just fine: 

```
chisel cut --release $PWD --root out base-files_base \
    libc6_libs \
    libgcc-s1_libs \
    libssl3_libs \
    libstdc++6_libs \
    zlib1g_libs \
    ca-certificates_data
```

+1 for merging

slices/ca-certificates.yaml
woky: Yup. Resolving...

niemeyer:
That looks like a great bootstrap indeed, thank you.

PR #2, opened 2022-06-07
Dotnet runtime slices

jssotomdz:
These slices are the same as the ones proposed for the dotnet rocks repo.
Now using globs to avoid dotnet version dependencies in paths.
Still waiting on the dotnet package publishing to universe so it can find the packages in the archive.


woky:
Incorporated into and superseded by https://github.com/canonical/chisel-releases/pull/5

PR #3, opened 2022-06-14
Add base-passwd and generate certificate bundle

woky:


slices/base-passwd.yaml
```diff
@@ -0,0 +1,14 @@
...
+
+slices:
+  data:
+    contents:
+      /usr/share/base-passwd/group.master:  {until: mutate}
+      /usr/share/base-passwd/passwd.master: {until: mutate}
+      /etc/group:  {text: FIXME, mutable: true}
+      /etc/passwd: {text: FIXME, mutable: true}
+    mutate: |
+      pw = content.read("/usr/share/base-passwd/group.master")
```
cjdcordeiro, line 11: this is probably the world most nitpciking commend :p , but the var names are swapped. `pw` should be `gr` and vice versa

cjdcordeiro (approved):
it lgtm.

I have 2 non-blocking comments though:
 1) (see inline comment about var names)
 2) in the `base-passwd` DEB, it actually is a `preinst` script who populates those files in /etc, **manually** (like printing a text snippet). The files in /usr/share are only used in `postinst`, by what I assume is a sort of backup procedure. So the comment/question is: _aren't we being untruthful to the DEB by ignoring the `preinst`?_

slices/base-passwd.yaml
woky: Good catch :-D

cjdcordeiro:
thanks for the info on the base-passwd's preinst :+1: 

no further comments then.

woky:
@niemeyer could you please merge this one if you have no objections?

slices/base-files.yaml
```diff
@@ -28,7 +28,7 @@ slices:
 
   tmp:
     contents:
-      /tmp/:
+      /tmp/: { mode: 01777 }
```
niemeyer, line 31: This shouldn't be required, as it's the actual mode of the underlying directory. I think we have a bug about umask getting in the way during creation. Is that what you're seeing? Why was the explicit mode added here?

slices/ca-certificates.yaml
```diff
@@ -3,5 +3,10 @@ package: ca-certificates
...
   data:
     contents:
-      /etc/ssl/certs/:
-      /usr/share/ca-certificates/mozilla/*.crt:
+      /etc/ssl/certs/ca-certificates.crt: {text: FIXME, mutable: true}
+      /usr/share/ca-certificates/mozilla/: {until: mutate}
+      /usr/share/ca-certificates/mozilla/*: {until: mutate}
+    mutate: |
+      certs_dir = "/usr/share/ca-certificates/mozilla/"
+      certs = [ content.read(certs_dir + path) for path in content.list(certs_dir) ]
```
niemeyer, line 11: Nitpick: I think it's more conventional in Python to not have spaces padding the edges of list comprehensions. Is that true? I don't mind too much the details of our conventions, as long as it's consistent, but since we're just starting we should try to borrow some typical Python style.

niemeyer (changes requested):
Just some nitpicking, and one relevant question:

slices/ca-certificates.yaml
woky: Ack. Fixed.

slices/base-files.yaml
woky: I though there wasn't bug but that the directory had wrong permissions in the archive since the postinst script explicitly adjusted its permissions. But I've just checked it's correct and indeed it sounds like a bug with umask handling. Reported here: https://github.com/canonical/chisel/issues/9

PR #4, opened 2022-07-08
Select arch-specific paths by globs

woky:
Tested with chisel with commits from PR
https://github.com/canonical/chisel/pull/12 applied:

    slices="base-files_base libc6_libs libgcc-s1_libs libssl3_libs libstdc++6_libs zlib1g_libs ca-certificates_data"
    for a in amd64 arm64 armhf i386 ppc64el riscv64 s390x; do
        mkdir output-$a
        chisel cut --release ~/devel/chisel-releases/ --root $PWD/output-$a --arch $a $slices
    done

### DIFF
--- a/slices/libc6.yaml
+++ b/slices/libc6.yaml
@@ -3,28 +3,28 @@ package: libc6
 slices:
   config:
     contents:
-      /etc/ld.so.conf.d/x86_64-linux-gnu.conf:
+      /etc/ld.so.conf.d/*-linux-*.conf:
 
   libs:
     contents:
-      /lib/x86_64-linux-gnu/ld-linux-x86-64.so.*:
-      /lib/x86_64-linux-gnu/libBrokenLocale.so.*:
-      /lib/x86_64-linux-gnu/libanl.so.*:
-      /lib/x86_64-linux-gnu/libc.so.*:
-      /lib/x86_64-linux-gnu/libc_malloc_debug.so.*:
-      /lib/x86_64-linux-gnu/libdl.so.*:
-      /lib/x86_64-linux-gnu/libm.so.*:
-      /lib/x86_64-linux-gnu/libmemusage.so:
-      /lib/x86_64-linux-gnu/libmvec.so.*:
-      /lib/x86_64-linux-gnu/libnsl.so.*:
-      /lib/x86_64-linux-gnu/libnss_compat.so.*:
-      /lib/x86_64-linux-gnu/libnss_dns.so.*:
-      /lib/x86_64-linux-gnu/libnss_files.so.*:
-      /lib/x86_64-linux-gnu/libnss_hesiod.so.*:
-      /lib/x86_64-linux-gnu/libpcprofile.so:
-      /lib/x86_64-linux-gnu/libpthread.so.*:
-      /lib/x86_64-linux-gnu/libresolv.so.*:
-      /lib/x86_64-linux-gnu/librt.so.*:
-      /lib/x86_64-linux-gnu/libthread_db.so.*:
-      /lib/x86_64-linux-gnu/libutil.so.*:
-      /lib64/ld-linux-x86-64.so.*:
+      /lib/*-linux-*/ld*.so.*:
+      /lib/*-linux-*/libBrokenLocale.so.*:
+      /lib/*-linux-*/libanl.so.*:
+      /lib/*-linux-*/libc.so.*:
+      /lib/*-linux-*/libc_malloc_debug.so.*:
+      /lib/*-linux-*/libdl.so.*:
+      /lib/*-linux-*/libm.so.*:
+      /lib/*-linux-*/libmemusage.so:
+      /lib/*-linux-*/libmvec.so.*: {arch: amd64}
+      /lib/*-linux-*/libnsl.so.*:
+      /lib/*-linux-*/libnss_compat.so.*:
+      /lib/*-linux-*/libnss_dns.so.*:
+      /lib/*-linux-*/libnss_files.so.*:
+      /lib/*-linux-*/libnss_hesiod.so.*:
+      /lib/*-linux-*/libpcprofile.so:
+      /lib/*-linux-*/libpthread.so.*:
+      /lib/*-linux-*/libresolv.so.*:
+      /lib/*-linux-*/librt.so.*:
+      /lib/*-linux-*/libthread_db.so.*:
+      /lib/*-linux-*/libutil.so.*:
+      /lib*/ld*.so.*:
--- a/slices/libcom-err2.yaml
+++ b/slices/libcom-err2.yaml
@@ -5,4 +5,4 @@ slices:
     essential:
       - libc6_libs
     contents:
-      /lib/x86_64-linux-gnu/libcom_err.so.*:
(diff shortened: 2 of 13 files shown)


slices/libc6.yaml
```diff
@@ -3,28 +3,28 @@ package: libc6
...
-      /lib64/ld-linux-x86-64.so.*:
+      /lib/*-linux-*/ld*.so.*:
+      /lib/*-linux-*/libBrokenLocale.so.*:
+      /lib/*-linux-*/libanl.so.*:
+      /lib/*-linux-*/libc.so.*:
+      /lib/*-linux-*/libc_malloc_debug.so.*:
+      /lib/*-linux-*/libdl.so.*:
+      /lib/*-linux-*/libm.so.*:
+      /lib/*-linux-*/libmemusage.so:
+      #/lib/*-linux-*/libmvec.so.*: # FIXME: libmvec is only available on amd64
```
niemeyer, line 18: This is the only bit concerning. If this was a stable release and we had more people using Chisel by now, we could not afford to drop a library like this, even if temporarily. Assuming the use cases we do have at hand don't depend on it for now, it's probably okay for now though. We need to sort this out via some sort of conditional though.

slices/libc6.yaml
```diff
@@ -3,28 +3,28 @@ package: libc6
 slices:
   config:
     contents:
-      /etc/ld.so.conf.d/x86_64-linux-gnu.conf:
+      /etc/ld.so.conf.d/*-linux-*.conf:
```
niemeyer, line 6: It's a bit unfortunate that x86_64 is not an actual architecture as far as dpkg is concerned. It would be a bit nicer to have this as ${ARCH} or similar. Maybe we can still do this in a similar way.

slices/libc6.yaml
cjdcordeiro: but without it, aren't we breaking the installation of libc6? I understand we cannot have it atm cause it doesn't belong in any other archs, but for our current use cases, when we chisel this package, we won't get this library in - isn't that a problem? I'd assume common apps making use of this lib for simple mathematical purposes would not work anymore...

I know there's no directive or conditionals in `chisel` at the moment to support this use case, but what if, just for the time being and to avoid breaking this use case, you create an `amd64`-specific slice, which we can explicitly refer to when installing the amd64 version of this package?

cjdcordeiro (changes requested):

slices/libc6.yaml
woky: > I know there's no directive or conditionals in chisel at the moment to support this use case, but what if, just for the time being and to avoid breaking this use case, you create an amd64-specific slice, which we can explicitly refer to when installing the amd64 version of this package?

That'd require completely separate slice tree, as almost every slice depends on libc6.

cjdcordeiro: the suggestion isn't a permanent fix. you wouldn't need to update the entire slice tree, but instead explicitly call for `libc6_amd64` whenever we're chiseling lib6-dependant slices for amd64.

Worst case scenario, we forget to explicitly chisel `libc6_amd64`, and the result will be the same as it is not in the PR -> we don't get that lib.

woky: > but instead explicitly call for libc6_amd64 whenever we're chiseling lib6-dependant slices for amd64.

I don't understand what you mean. If I want to install `libkrb5_libs` slice on arm64, I'm also installing `libc6_libs`. Specifying `libc6_amd64_libs` on command-line won't replace this dependency. It'll result in conflict.

cjdcordeiro: let me try to break it down:

let's say you are running the following:

`chisel cut --release ubuntu-22.04 --root "${PWD}/output" libssl3_libs`

both for amd64 and arm64.

with the SDF from this PR, you will not get `libmvec`, which is correct for arm64, but not for amd64.

---

now let's say we have an additional, and temporary (note: I don't like creating tmp things that we know we're gonna trash later on, so that's why this is just a suggestion), slice called `libc6_amd64` pointing to the amd64 version of `libmvec`. 

In this case, what we can do to avoid an incomplete amd64 installation is to:
 1. run the aforementioned `chisel` command, as is, for arm64
 2. run `chisel cut --release ubuntu-22.04 --root "${PWD}/output" libssl3_libs libc6_amd64` for amd64

---

bottom-line, the options are:
 1) merge as is, knowing that until a `chisel` patch is released for this, we'll be chiseling systems which wouldn't have `libmvec`
 2) exceptionally add this amd64-specific slice to libc6 just for the time being, to cover our dotnet-runtime-deps use case, until a `chisel` patch is released

I'm fine with either decision. I wanted to present option 2) just in case it hadn't been considered, and make sure whatever option is chosen, it is chosen deliberately. @niemeyer should also weigh in on it

woky: Now I understand. So `libc6_amd64` is a slice that contains only `/lib/*-linux-*/libmvec.so.*`. I've pushed PoC commit that splits `libc6` like this, except I call the slice `_mvec` (with underscore prefix to make it temporary), so your example for `amd64` would be:
```
chisel cut --release ubuntu-22.04 --root "${PWD}/output" libssl3_libs libc6__mvec
```

woky: OK, underscore doesn't work...

woky:
I've updated the PR to use `{arch: amd64}` conditional for `/lib/*-linux-*/libmvec.so.*` path.

slices/libc6.yaml
```diff
@@ -3,28 +3,28 @@ package: libc6
...
-      /lib/x86_64-linux-gnu/libnss_files.so.*:
-      /lib/x86_64-linux-gnu/libnss_hesiod.so.*:
-      /lib/x86_64-linux-gnu/libpcprofile.so:
-      /lib/x86_64-linux-gnu/libpthread.so.*:
-      /lib/x86_64-linux-gnu/libresolv.so.*:
-      /lib/x86_64-linux-gnu/librt.so.*:
-      /lib/x86_64-linux-gnu/libthread_db.so.*:
-      /lib/x86_64-linux-gnu/libutil.so.*:
-      /lib64/ld-linux-x86-64.so.*:
+      /lib/*-linux-*/ld*.so.*:
```
niemeyer, line 10: Do we want a dash here, as in ld-*.so.*?

slices/libc6.yaml
```diff
@@ -3,28 +3,28 @@ package: libc6
...
+      /lib/*-linux-*/libnss_dns.so.*:
+      /lib/*-linux-*/libnss_files.so.*:
+      /lib/*-linux-*/libnss_hesiod.so.*:
+      /lib/*-linux-*/libpcprofile.so:
+      /lib/*-linux-*/libpthread.so.*:
+      /lib/*-linux-*/libresolv.so.*:
+      /lib/*-linux-*/librt.so.*:
+      /lib/*-linux-*/libthread_db.so.*:
+      /lib/*-linux-*/libutil.so.*:
+      /lib*/ld*.so.*:
```
niemeyer, line 30: Same as above.

niemeyer (commented):
One question, and still LGTM otherwise.

slices/libc6.yaml
woky: It's `/lib/ld64.so.1` :-)

PR #5, opened 2022-07-20
Add dotnet6 to new 22.10 (kinetic) branch

woky:
dotnet6 is currently in kinetic, but not yet in jammy. Add dotnet6 slices to
ubuntu-22.10 branch.

PLEASE DO NOT squash merge this PR. The commit that adds dotnet6 shall be
cherry-picked to ubuntu-22.04 branch once dotnet6 packages hit jammy archive.

### DIFF
--- a/slices/aspnetcore-runtime-6.0.yaml
+++ b/slices/aspnetcore-runtime-6.0.yaml
@@ -1,0 +1,7 @@
+package: aspnetcore-runtime-6.0
+slices:
+  libs:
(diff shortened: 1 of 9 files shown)


cjdcordeiro (approved):
I see that the kinetic branch already exists and already has the same base packages as 22.04. so this lgtm

niemeyer (commented):
No knowledge of the details of these packages, but LGTM in general.

PR #8, opened 2022-08-01
dotnet-hostfxr-6.0_libs: Add dotnet-host_bins dependency

woky:


niemeyer (approved):
Can speak for the semantics, but looks okay.

PR #9, opened 2022-08-02
Make kinetic slices arch-independent

woky:


cjdcordeiro (approved):
lgtm

niemeyer (approved):
LGTM

As we discussed, it would be nice to find a more automated way to keep things in sync, or to know when they diverge unintentionally.

PR #14, opened 2022-10-03
[22.04] add release-info slice to base-files slice definition file

cjdcordeiro:
Create a new slice in `base-files` to install all content related to the underlying linux distribution (eg. /etc/os-release, /etc/issue, etc.)


woky (approved):
LGTM

PR #15, opened 2022-10-03
[22.10] add release-info slice to base-files slice definition file

cjdcordeiro:
Same as in https://github.com/canonical/chisel-releases/pull/14, but for Kinetic


woky (approved):
LGTM

PR #16, opened 2022-10-04
[22.04] create SDF for tzdata

cjdcordeiro:
Introduce a new slice definition file for `tzdata`.

The slicing logic is based on:
 - different regions can be installed separately;
    - Europe and Asia have intertwined files, so they belong to the same slice
 - standalone time zones (without location) can also be installed separately via the `zoneinfo-base`

Tested with a simple Golang app that runs in a container with and without the `zoneinfo-base` slice. With the slice, the EST printout works.

```golang
func main() {
    tm := time.Now()
    fmt.Printf("%s\n", tm.Local())
    location, err := time.LoadLocation("EST")
    if err != nil {
        fmt.Println(err)
    }
    fmt.Println("ZONE : ", location, " Time : ", tm.In(location)) // EST
}
```

### DIFF
--- a/slices/tzdata.yaml
+++ b/slices/tzdata.yaml
@@ -1,0 +1,300 @@
+package: tzdata
+
+slices:
+  zoneinfo:
+    essential:
+      - tzdata_africa
+      - tzdata_america
+      - tzdata_antarctica
+      - tzdata_arctic
+      - tzdata_atlantic
+      - tzdata_australia
+      - tzdata_brazil
+      - tzdata_canada
+      - tzdata_chile
+      - tzdata_etc
+      - tzdata_eurasia
+      - tzdata_indian
+      - tzdata_mexico
+      - tzdata_pacific
+      - tzdata_united-states
+      - tzdata_zoneinfo-base
+
+  zoneinfo-icu:
+    contents:
+      /usr/share/zoneinfo-icu/44/be/*:
+      /usr/share/zoneinfo-icu/44/le/*:
+
+  zoneinfo-base:
+    contents:
+      /usr/share/zoneinfo/CET:
+      /usr/share/zoneinfo/CST6CDT:
+      /usr/share/zoneinfo/EET:
+      /usr/share/zoneinfo/EST:
+      /usr/share/zoneinfo/EST5EDT:
+      /usr/share/zoneinfo/Factory:
+      /usr/share/zoneinfo/HST:
+      /usr/share/zoneinfo/iso3166.tab:
+      /usr/share/zoneinfo/leapseconds:
+      /usr/share/zoneinfo/leap-seconds.list:
+      /usr/share/zoneinfo/localtime:
+      /usr/share/zoneinfo/MET:
+      /usr/share/zoneinfo/MST:
+      /usr/share/zoneinfo/MST7MDT:
+      /usr/share/zoneinfo/posix/CET:
+      /usr/share/zoneinfo/posix/CST6CDT:
+      /usr/share/zoneinfo/posix/EET:
+      /usr/share/zoneinfo/posix/EST:
+      /usr/share/zoneinfo/posix/EST5EDT:
+      /usr/share/zoneinfo/posix/Factory:
+      /usr/share/zoneinfo/posix/HST:
+      /usr/share/zoneinfo/posix/MET:
+      /usr/share/zoneinfo/posix/MST:
+      /usr/share/zoneinfo/posix/MST7MDT:
+      /usr/share/zoneinfo/posix/PST8PDT:
+      /usr/share/zoneinfo/posix/WET:
+      /usr/share/zoneinfo/PST8PDT:
+      /usr/share/zoneinfo/right/CET:
+      /usr/share/zoneinfo/right/CST6CDT:
+      /usr/share/zoneinfo/right/EET:
+      /usr/share/zoneinfo/right/EST:
+      /usr/share/zoneinfo/right/EST5EDT:
+      /usr/share/zoneinfo/right/Factory:
+      /usr/share/zoneinfo/right/HST:
+      /usr/share/zoneinfo/right/MET:
+      /usr/share/zoneinfo/right/MST:
+      /usr/share/zoneinfo/right/MST7MDT:
+      /usr/share/zoneinfo/right/PST8PDT:
+      /usr/share/zoneinfo/right/WET:
+      /usr/share/zoneinfo/tzdata.zi:
+      /usr/share/zoneinfo/WET:
+      /usr/share/zoneinfo/zone1970.tab:
+      /usr/share/zoneinfo/zone.tab:
+
+  africa:
+    contents:
+      /usr/share/zoneinfo/Africa/*:
+      /usr/share/zoneinfo/posix/Africa/*:
+      /usr/share/zoneinfo/right/Africa/*:
+      /usr/share/zoneinfo/Egypt:
+      /usr/share/zoneinfo/Iceland:
+      /usr/share/zoneinfo/Libya:
+      /usr/share/zoneinfo/posix/Egypt:
+      /usr/share/zoneinfo/right/Egypt:
+      /usr/share/zoneinfo/posix/Iceland:
+      /usr/share/zoneinfo/right/Iceland:
+      /usr/share/zoneinfo/posix/Libya:
+      /usr/share/zoneinfo/right/Libya:
+
+  america:
+    contents:
+      /usr/share/zoneinfo/America/Argentina/*:
+      /usr/share/zoneinfo/America/Indiana/*:
+      /usr/share/zoneinfo/America/Kentucky/*:
+      /usr/share/zoneinfo/America/North_Dakota/*:
+      /usr/share/zoneinfo/America/*:
+      /usr/share/zoneinfo/posix/America/*:
+      /usr/share/zoneinfo/posix/America/Argentina/*:
+      /usr/share/zoneinfo/posix/America/Indiana/*:
+      /usr/share/zoneinfo/posix/America/Kentucky/*:
+      /usr/share/zoneinfo/posix/America/North_Dakota/*:
+      /usr/share/zoneinfo/right/America/*:
+      /usr/share/zoneinfo/right/America/Argentina/*:
+      /usr/share/zoneinfo/right/America/Indiana/*:
+      /usr/share/zoneinfo/right/America/Kentucky/*:
+      /usr/share/zoneinfo/right/America/North_Dakota/*:
+      /usr/share/zoneinfo/Cuba:
+      /usr/share/zoneinfo/Jamaica:
+      /usr/share/zoneinfo/Navajo:
+      /usr/share/zoneinfo/posix/Cuba:
+      /usr/share/zoneinfo/right/Cuba:
+      /usr/share/zoneinfo/posix/Jamaica:
+      /usr/share/zoneinfo/right/Jamaica:
+      /usr/share/zoneinfo/posix/Navajo:
+      /usr/share/zoneinfo/right/Navajo:
+      /usr/share/zoneinfo/posixrules:
+
+  antarctica:
+    essential:
+      - tzdata_eurasia
+      - tzdata_pacific
+    contents:
+      /usr/share/zoneinfo/Antarctica/*:
+      /usr/share/zoneinfo/posix/Antarctica/*:
+      /usr/share/zoneinfo/right/Antarctica/*:
+
+  arctic:
+    essential:
+      - tzdata_eurasia
+    contents:
+      /usr/share/zoneinfo/Arctic/*:
+      /usr/share/zoneinfo/posix/Arctic/*:
+      /usr/share/zoneinfo/right/Arctic/*:
+
+  atlantic:
+    essential:
+      - tzdata_africa
+      - tzdata_eurasia
+    contents:
+      /usr/share/zoneinfo/Atlantic/*:
+      /usr/share/zoneinfo/posix/Atlantic/*:
+      /usr/share/zoneinfo/right/Atlantic/*:
+
+  australia:
+    contents:
+      /usr/share/zoneinfo/Australia/*:
+      /usr/share/zoneinfo/posix/Australia/*:
+      /usr/share/zoneinfo/right/Australia/*:
+
+  brazil:
+    essential:
+      - tzdata_america
+    contents:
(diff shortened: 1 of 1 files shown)


woky (approved):
I can't imagine a use case where someone would like to install timezones for specific regions only. It makes sense to install single (or few) timezone(s) to format local time on your workstation, but for containers I can see two use cases:
1. Requests for clients from anywhere
2. Single timezone for database

For 1. this split is not needed. For 2. this split is too coarse.

Anyway, I don't know much about timezone use cases and as long as it's possible to install complete package, it's all good.

cjdcordeiro:
thanks for the comment. I don't mind using `**`. IMO it feels way too permissive, given the amount of files within. But I can understand both perspectives. @niemeyer what do you think?

lbussell:
Hello @cjdcordeiro and @woky, are there plans to merge this `tzdata` slice any time soon? We would like to include `tzdata` in a version of our Ubuntu Chiseled Distroless images for .NET. We would probably prefer to install all of the region/timezone data together. Thanks.

richlander:
Context: https://github.com/dotnet/dotnet-docker/discussions/4821

cjdcordeiro:
Hi @lbussell ! Noted. We'll try to get it in b4 EOW.

slices/tzdata.yaml
```diff
@@ -0,0 +1,300 @@
...
+      /usr/share/zoneinfo/posix/UTC:
+      /usr/share/zoneinfo/right/UCT:
+      /usr/share/zoneinfo/right/Universal:
+      /usr/share/zoneinfo/right/UTC:
+      /usr/share/zoneinfo/posix/Zulu:
+      /usr/share/zoneinfo/right/Zulu:
+      /usr/share/zoneinfo/UCT:
+      /usr/share/zoneinfo/Universal:
+      /usr/share/zoneinfo/UTC:
+      /usr/share/zoneinfo/Zulu:
```
rebornplusplus, line 205: I would merge this with `base`. This contains `UTC` among other things. And I noticed `CET` is in `base`. It feels logical to group them together.

slices/tzdata.yaml
```diff
@@ -0,0 +1,300 @@
...
+      /usr/share/zoneinfo/right/Canada/*:
+
+  chile:
+    essential:
+      - tzdata_america
+      - tzdata_pacific
+    contents:
+      /usr/share/zoneinfo/Chile/*:
+      /usr/share/zoneinfo/posix/Chile/*:
+      /usr/share/zoneinfo/right/Chile/*:
```
rebornplusplus, line 172: As `canada`, I would put `chile` in `america`.

slices/tzdata.yaml
```diff
@@ -0,0 +1,300 @@
...
+      /usr/share/zoneinfo/zone1970.tab:
+      /usr/share/zoneinfo/zone.tab:
+
+  africa:
+    contents:
+      /usr/share/zoneinfo/Africa/*:
+      /usr/share/zoneinfo/posix/Africa/*:
+      /usr/share/zoneinfo/right/Africa/*:
+      /usr/share/zoneinfo/Egypt:
+      /usr/share/zoneinfo/Iceland:
```
rebornplusplus, line 80: Iceland is probably in Europe.

slices/tzdata.yaml
```diff
@@ -0,0 +1,300 @@
...
+      /usr/share/zoneinfo/posix/America/*:
+      /usr/share/zoneinfo/posix/America/Argentina/*:
+      /usr/share/zoneinfo/posix/America/Indiana/*:
+      /usr/share/zoneinfo/posix/America/Kentucky/*:
+      /usr/share/zoneinfo/posix/America/North_Dakota/*:
+      /usr/share/zoneinfo/right/America/*:
+      /usr/share/zoneinfo/right/America/Argentina/*:
+      /usr/share/zoneinfo/right/America/Indiana/*:
+      /usr/share/zoneinfo/right/America/Kentucky/*:
+      /usr/share/zoneinfo/right/America/North_Dakota/*:
```
rebornplusplus, line 105: How about `.../America/**:`?

woky:
FYI #60 introduced tzdata as well

slices/tzdata.yaml
```diff
@@ -0,0 +1,300 @@
...
+      - tzdata_brazil
+      - tzdata_canada
+      - tzdata_chile
+      - tzdata_etc
+      - tzdata_eurasia
+      - tzdata_indian
+      - tzdata_mexico
+      - tzdata_pacific
+      - tzdata_united-states
+      - tzdata_zoneinfo-base
```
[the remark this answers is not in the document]
rebornplusplus: Oh wait, is the list in this page faulty?
https://packages.ubuntu.com/jammy/all/tzdata/filelist

I later downloaded the deb and extracted myself. Upon comparison I only get the following minimal diff instead of the late one before:
```diff
1d0
< /usr/sbin/tzconfig
3,12d1
< /usr/share/doc/tzdata/changelog.Debian.gz
< /usr/share/doc/tzdata/README.Debian
< /usr/share/zoneinfo-icu/44/be/timezoneTypes.res
< /usr/share/zoneinfo-icu/44/be/windowsZones.res
< /usr/share/zoneinfo-icu/44/be/zoneinfo64.res
< /usr/share/zoneinfo-icu/44/be/metaZones.res
< /usr/share/zoneinfo-icu/44/le/timezoneTypes.res
< /usr/share/zoneinfo-icu/44/le/windowsZones.res
< /usr/share/zoneinfo-icu/44/le/zoneinfo64.res
< /usr/share/zoneinfo-icu/44/le/metaZones.res
```

cjdcordeiro:
> I took a brief look and will be thorough next time. Although I do understand the reasons for splitting this into zone-specific slices, I do wonder how much those will be used. In any case, as long as there is a slice which contains all of them, I don't see any harm in having those; only pros.
> 
> But I would probably not group the timezones into country-specific slices. Continents seem like a good choice. America, Africa, Antartica, Australia, Eurasia -- seems good. I wonder if we should merge the `Atlantic`, `Arctic` and `Pacific` slices into one. They are fine on their own as they are right now, however.

Thanks for the feedback.

I'm not too sure about this TBH. For multiple reasons:
 - the `tzdata` package is actually organized this way, so by respecting it we're taking the most conservative approach possible, which IMO is the best way to start
 - certain factoring choices seem logical to me. For example:
    - while country X may be part of continent Y, the geographical span of X may deem it unfit (tz-wise) to be part of Y
    - certain countries do not abide by the same tz conventions as their neighbouring ones (eg. single TZ, fractional TZs, etc)
 - while I also struggle to identify the rationale behind `tzdata`'s internal structuring, I could easily see cases where a country's TZ span overlaps with another country's TZ, from a different continent.

I'll proceed to address your comments, but I'd rather stick with the package's original structure, and later on, if deemed appropriate, create additional slices that group the geographical continents as you're suggesting (e.g. `americas-continent` or similar)

slices/tzdata.yaml
cjdcordeiro: So, `Etc` is meant to provide "timezones" that don't fit with the standard timezones. UTC isn't actually a timezone, but a standard. Like Zulu and others, most of these can be used for time information, but derive from different domains (like the military). Same for others.

This [table](https://en.wikipedia.org/wiki/List_of_tz_database_time_zones#List) is actually useful as it notes that Zulu, UTC, and others within this slice are actually links and not Canonical timezones. 

I'll add comments to these slices to provide more info.

slices/tzdata.yaml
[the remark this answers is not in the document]
cjdcordeiro: That sounds about right. The `zoneinfo-icu` slice was intentionally not included in `zoneinfo` since it is very specific to the ICU library.

slices/tzdata.yaml
cjdcordeiro: it is a symlink to `Africa/Abidjan`. Having it in the `eurasia` slice would create a dependency between the two slices

cjdcordeiro: I'll move it anyway. If we could handle slice conflicts, I'd just add the `Africa/Abidjan` to `eurasia`...but atm that won't work

slices/tzdata.yaml
```diff
@@ -0,0 +1,300 @@
...
+      /usr/share/zoneinfo/Iceland:
+      /usr/share/zoneinfo/Libya:
+      /usr/share/zoneinfo/posix/Egypt:
+      /usr/share/zoneinfo/right/Egypt:
+      /usr/share/zoneinfo/posix/Iceland:
+      /usr/share/zoneinfo/right/Iceland:
+      /usr/share/zoneinfo/posix/Libya:
+      /usr/share/zoneinfo/right/Libya:
+
+  america:
```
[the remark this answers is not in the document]
cjdcordeiro: `America` is the right TZ designation:

image

slices/tzdata.yaml
cjdcordeiro: Sure

slices/tzdata.yaml
```diff
@@ -0,0 +1,300 @@
...
+      /usr/share/zoneinfo/posix/Brazil/*:
+      /usr/share/zoneinfo/right/Brazil/*:
+
+  canada:
+    essential:
+      - tzdata_america
+    contents:
+      /usr/share/zoneinfo/Canada/*:
+      /usr/share/zoneinfo/posix/Canada/*:
+      /usr/share/zoneinfo/right/Canada/*:
```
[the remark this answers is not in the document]
cjdcordeiro: (same as https://github.com/canonical/chisel-releases/pull/16/files#r1297030775)

slices/tzdata.yaml
cjdcordeiro: (same as https://github.com/canonical/chisel-releases/pull/16/files#r1297030775)

slices/tzdata.yaml
```diff
@@ -0,0 +1,300 @@
...
+      /usr/share/zoneinfo/posix/Indian/*:
+      /usr/share/zoneinfo/right/Indian/*:
+
+  mexico:
+    essential:
+      - tzdata_america
+    contents:
+      /usr/share/zoneinfo/Mexico/*:
+      /usr/share/zoneinfo/posix/Mexico/*:
+      /usr/share/zoneinfo/right/Mexico/*:
```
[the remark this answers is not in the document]
cjdcordeiro: (same as https://github.com/canonical/chisel-releases/pull/16/files#r1297030775)

slices/tzdata.yaml
```diff
@@ -0,0 +1,300 @@
...
+      /usr/share/zoneinfo/right/NZ-CHAT:
+
+  united-states:
+    essential:
+      - tzdata_america
+      - tzdata_pacific
+    contents:
+      /usr/share/zoneinfo/US/*:
+      /usr/share/zoneinfo/posix/US/*:
+      /usr/share/zoneinfo/right/US/*:
```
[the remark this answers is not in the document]
cjdcordeiro: (same as https://github.com/canonical/chisel-releases/pull/16/files#r1297030775)

cjdcordeiro:
@rebornplusplus can you please have another look. 

Extra info - size per slice:

```
base - 320K tests
africa - 712K tests
america - 1.8M tests
antarctica - 2.9M tests
arctic - 2.4M tests
atlantic - 2.5M tests
australia - 220K tests
brazil - 1.8M tests
canada - 1.8M tests
chile - 2.3M tests
config - 68K tests
etc - 384K tests
eurasia - 2.4M tests
indian - 2.5M tests
mexico - 1.8M tests
pacific - 544K tests
united-states - 2.3M tests
zoneinfo - 5.7M tests
zoneinfo-icu - 512K tests
```

rebornplusplus (approved):
Looks good. :world_map: :clock10: 

nit: There's no newline at the end.

Question: should every slice require `base`?

cjdcordeiro:
I'll add the newlines before merging, thanks ;)

> Question: should every slice require base?

I don't think so. I've split the `.tab` files into a new slice `config` precisely for that, so I think `base` is now a sort of generic standalone slice.

woky:
> I'm not too sure about this TBH. For multiple reasons:
> 
> * the `tzdata` package is actually organized this way, so by respecting it we're taking the most conservative approach possible, which IMO is the best way to start
> * certain factoring choices seem logical to me. For example:
>   
>   * while country X may be part of continent Y, the geographical span of X may deem it unfit (tz-wise) to be part of Y
>   * certain countries do not abide by the same tz conventions as their neighbouring ones (eg. single TZ, fractional TZs, etc)
> * while I also struggle to identify the rationale behind `tzdata`'s internal structuring, I could easily see cases where a country's TZ span overlaps with another country's TZ, from a different continent.

As I said, in my comment, I don't know of any use case where you'd want timezone data for specific region only, so this partitioning is of dubious value. Users are interested in _full_ tzdata slice. As long as this file provides it, it's fine. But since we're picking directory by directory, it's hard to tell, and it'll be hard to sync with the package. Hopefully countries won't get created/destroyed too often for this to be a problem.

lbussell:
> before I merge, @lbussell @richlander since you're about to use these slices, can you please double-check if the current slice design makes sense for your use case?

It seems OK. We would probably just install `zoneinfo` and `zoneinfo-icu` since the purpose is to provide full globalization capabilities to our users.

cjdcordeiro:
Alright thanks for the info.

In that case, and to @woky 's comment, there's very little gain in installing those two slices together vs the whole deb, since those two slices bring in 99% of the deb (only `./usr/sbin/tzconfig` and `./usr/share/doc/` are left out).

Nonetheless, the current design fits your needs, which is good, and it also accommodates other possible use cases where the whole package might not be needed, and where Chisel becomes more valuable.

richlander:
Thanks for this effort and all the info.

> `zoneinfo-icu - 512K tests`

Honest q. What's the relevance of the # of tests? I want to make sure I'm not missing anything.

If we take just the `zoneinfo*` slices, what are we missing?

@tarekgh

cjdcordeiro:
@richlander np ;) happy to help.

> Honest q. What's the relevance of the # of tests? I want to make sure I'm not missing anything.

oh nvm that. That's just the name of the folder where I was installing the slices :smile: 

> If we take just the zoneinfo* slices, what are we missing?

You'll be missing `./usr/sbin/tzconfig` and `./usr/share/doc/tzdata/*Debian*`, which account for ~0.25% of the whole package (<4kb).

richlander:
Thanks. I appreciate the help.

PR #18, opened 2022-10-04
[22.04] add slice definition file for libc-bin, with nsswitch slice only

cjdcordeiro:
This new slice definition file encompasses only the `nsswitch.conf` file of the libc-bin package.

### DIFF
--- a/slices/libc-bin.yaml
+++ b/slices/libc-bin.yaml
@@ -1,0 +1,6 @@
+package: libc-bin
+
+slices:
+  nsswitch:
+    contents:
+      /etc/nsswitch.conf: {copy: /usr/share/libc-bin/nsswitch.conf}


woky (approved):
Approved, but I wonder, in the name of being tiny, can't we drop the first file and keep just the one in `/etc` (copied from `/usr/share/...`)? I think that libc only looks in etc when doing name resolution. (I've just fiddled wit it yesterday.)

cjdcordeiro:
at 1st glance, I'd agree. But can we really make sure that `/usr/share/libc-bin/nsswitch.conf` is never needed anymore? Ok, libc doesn't seem to use it, what about other packages? And what about other maintainer scripts? Given its size, I'd sat we should rather stay true to the DEB...

woky:
I get that we cannot know which files are used by what, but by this rationale we can quickly go back to Ubuntu base image. I think that this kind of file is kept only because admin might have its own version in `/etc`, so at least a backup is kept. That doesn't apply to containers. Since we do not have any analysis I'd say that if it's unlikely to be used, just drop it and wait for bug report. Otherwise we will be adding unnecessary files that will people will begin to depend on.

slices/libc-bin.yaml
```diff
@@ -0,0 +1,7 @@
+package: libc-bin
+
+slices:
+  nsswitch:
+    contents:
+      /usr/share/libc-bin/nsswitch.conf:
+      /etc/nsswitch.conf: {copy: /usr/share/libc-bin/nsswitch.conf}
```
rebornplusplus, line 7: nit: no newline at the end-of-file.

rebornplusplus (commented):
I agree with Tomas on being minimalistic here. I think we can just keep the `/etc/nsswitch.conf` file here and wait to see if anyone complains of not having the file at `/usr/share/libc-bin/`. 

Otherwise the PR looks good.

woky (approved):
LGTM

cjdcordeiro:
> I agree with Tomas on being minimalistic here. I think we can just keep the `/etc/nsswitch.conf` file here and wait to see if anyone complains of not having the file at `/usr/share/libc-bin/`.
> 
> Otherwise the PR looks good.

But that is not what the deb does. This is:

```bash
install_from_default() {
  if [ ! -f $2 ]; then
    cp -p $1 $2
  fi
}
...
install_from_default /usr/share/libc-bin/nsswitch.conf /etc/nsswitch.conf
```

So why should we steer away from the deb? If we were knowledgeable enough about the deb, such that we could say: "we don't need that file", then y, let's remove it. That's not the case. Sure we could just build things from the ground up, with our own assumptions, and wait for people to complain. But that doesn't seem very sustainable or a nice experience. 

Anyway, in this case, we do know a little bit about the pkg, so, to get this merged, I'll just remove it. But I urge you to stay true to the debs when slicing them, **unless** you are a subject matter expert on what the deb

rebornplusplus (approved):

PR #25, opened 2022-11-29
fix: dotnet* pkgs new content structure (22.10)

cjdcordeiro:
(same as https://github.com/canonical/chisel-releases/pull/23, but for Kinetic)


woky (approved):
Checked the package. LGTM.

niemeyer (approved):
If the package changed this way, there's not much to be done here. But it's a curious choice in the sense the prior scheme supported parallel installations, and this doesn't. Thanks for the fix.

PR #26, opened 2022-12-01
fix: aspnetcore-runtime-6.0 new content structure (22.04)

cjdcordeiro:
Same reason as for https://github.com/canonical/chisel-releases/pull/23. In that PR, ASPNET was forgotten, but apparently, it has also suffered the same restructuring as all other .NET runtime packages.


woky (approved):

niemeyer (approved):
Haven't seen the packages, and the same comment applies about it moving from a setup that could tolerate multiple versions towards one that doesn't, but assuming the package has already moved that way there's not much we can do here.

PR #27, opened 2022-12-01
fix: aspnetcore-runtime-6.0 new content structure (22.10)

cjdcordeiro:
analogous to https://github.com/canonical/chisel-releases/pull/26


woky (approved):

niemeyer (approved):
Haven't seen the packages, and the same comment applies about it moving from a setup that could tolerate multiple versions towards one that doesn't, but assuming the package has already moved that way there's not much we can do here.

PR #35, opened 2023-05-06
[IMP] new slices to jammy: libgmp10 libhogweed6 libidn2-0 libnettle8 libp11-kit0 libtasn1-6 libunistring2...

grzs:
New slices:
libexpat1_libs
libffi8_libs
libgmp10_libs
libgnutls30_libs
libhogweed6_libs
libidn2-0_libs
libldap-2.5-0_libs
libnettle8_libs
libp11-kit0_libs
libpq-dev_bins
libpq5_libs
libsasl2-2_libs
libtasn1-6_libs
libunistring2_libs


grzs:
Sorry, I forgot to switch to my dev branch in the forked repo while pushing new changes. It's become a fat pull request now.

grzs:
@cjdcordeiro thanks for the reply. I don't want to add more for a while and what I added worked for me.

woky:
We have three branches now and soon we may have fourth for focal. This PR is against `ubuntu-22.04`. Nothing against this PR but we need to decide out how to maintain diverging branches in this repository and keep our sanity.

@grzs Do you have any example application for which you need these slices?

grzs:
@woky do you have any suggestions to consolidate the diverging branches? Are there guidelines to follow? Or should I check out the other branches for paradigms to follow, or copy existing slices from there? I think the libs for python3.10 are important and needed for many projects.

cjdcordeiro (changes requested):
Many thanks once more for this contribution. It looks good and I only have a few minor remarks:
 1. `libhogweed6` shouldn't depend `libnettle8` shouldn't it?
 2. many of the suggested contents are abusing the globs. An example is `libnettle8`, which only has one file (and a symlink) - `./usr/lib/x86_64-linux-gnu/libnettle.so.8.4` - but still, the slice definition is using `/usr/lib/*-linux-*/libnettle.so.*`. The best path here would be `/usr/lib/*-linux-gnu/libnettle.so.8.4` + `/usr/lib/*-linux-gnu/libnettle.so.8`, or at least `/usr/lib/*-linux-gnu/libnettle.so.8*`. As a general rule, this makes chiselling more deterministic and avoid potential conflicts coming from the introduction of the version of the same package into the same release. I suggest going through all the submitted slices and making the contents a bit more explicit.
 3. doesn't `libpq-dev` depend on `libssl-dev`?
 4. `libpq-dev` does not seem functional, and is missing a couple of libs and a symlink
 5. `libpython3.10-stdlib` should depend on `libpython3.10-minimal`
 6. should `libsasl2-2` depend on `libsasl2-modules-db`?
 7. as a best practice, I'd rename `libtirpc-common_etc` to `libtirpc-common_config`, since that's a configuration file ("etc" sounds too generic)
 8. for the same reason as 7., I'd rename `media-types_etc` to `media-types_data`
 9. shouldn't `python3.10-minimal` depend on `libpython3.10-minimal`?
 10. `python3.10-minimal` seems to me missing some mutation scripts, as there is additional content that is installed in the filesystem via maintainer scripts...so the question is how relevant that content is in order to achieve minimum functionality?
 11. `python3.10` should also depend on `libpython3.10-stdlib` and `media-types`
 12. `readline-common` has an empty `essential` field

Wrt 10., as @woky said, if you have a use case that we can reproduce as a validation test, that would be better, because Python is a complicated slicing exercise, so we need to make sure we make it future-proof as more complex cases come in.

Sorry for the long bullet list :) most are nitpicks and simple fixes anyway. It's the Python* slices that required a bit more testing and validation.

woky (changes requested):
Please squash the changes into single commit. The title could be something like _Add dependencies for psycopg2_.

grzs:
@woky , @cjdcordeiro thank you for the clarifications. I think the best is to close this PR and start separate ones for python3.10, psycopg2, and wkhtmltopdf. I am new to chisel and couldn't find detailed documentation so I would appreciate it if you could direct me to a page where I can learn more about mutation scripts.

setharnold:
re `/usr/lib/*-linux-*/libnettle.so.*` this is a common pattern in the existing definitions:

https://github.com/canonical/chisel-releases/blob/ubuntu-22.04/slices/libc6.yaml
https://github.com/canonical/chisel-releases/blob/ubuntu-22.04/slices/libgssapi-krb5-2.yaml

Should the existing ones be changed? Or is there a reason for the difference?

Thanks

cjdcordeiro:
whenever there are version-specific elements of the pkg within the paths, then yes, it is best practice to include them. That applies to all new and old slices.

rebornplusplus:
Hi @grzs, it's been a while since the last activity here. I see that some of the slices you wanted to add have been added by others already. I reckon this PR could use a rebase, there are a few conflicts already. If you want to close this PR (and start a new one), that's fine too. Let me know, thank you!

clay-lake:
Hi @grzs, Im just checking in. based on your earlier comment, is it ok if I close this PR?

> ... I think the best is to close this PR and start separate ones for python3.10, psycopg2, and wkhtmltopdf....

grzs:
@clay-lake thanks for asking, yes, it's better to close it

PR #37, opened 2023-06-19
feat: busybox binary slice

vpa1977:
This PR adds a busybox binaries slice that is useful for building small shell-enabled Java containers. Applications such as Apache Tomcat and Apache Kafka rely on the common shell commands in the launcher scripts and having a busybox slice simplifies their deployment. 

Note: the list of symlinks was generated by running ``busybox --list``, so please advice if there is a better option to create it.

Test Docker [file](https://github.com/vpa1977/slice-tests/blob/main/busybox/Dockerfile). 

Testing:
```
$ docker build -t busybox .
[+] Building 1.0s (17/17) FINISHED                                                                                                                                                            
 => [internal] load build definition from Dockerfile                                                                                                                                     0.1s
 => => transferring dockerfile: 32B                                                                                                                                                      0.0s
 => [internal] load .dockerignore                                                                                                                                                        0.0s
 => => transferring context: 2B                                                                                                                                                          0.0s
 => [internal] load metadata for public.ecr.aws/ubuntu/ubuntu:22.04@sha256:2c1168b31e636c7ab22598bfaefa6de63d1806a908d0c39bd402277881356fab                                              0.0s
 => [internal] load metadata for docker.io/library/golang:1.18                                                                                                                           0.8s
 => [builder 1/3] FROM public.ecr.aws/ubuntu/ubuntu:22.04@sha256:2c1168b31e636c7ab22598bfaefa6de63d1806a908d0c39bd402277881356fab                                                        0.0s
 => [chisel 1/4] FROM docker.io/library/golang:1.18@sha256:50c889275d26f816b5314fc99f55425fa76b18fcaf16af255f5d57f09e1f48da                                                              0.0s
 => CACHED [builder 2/3] RUN apt-get update     && DEBIAN_FRONTEND=noninteractive apt-get install -y         binutils     && apt-get clean -y     && rm -rf /var/lib/apt/lists/*         0.0s
 => CACHED [chisel 2/4] RUN git clone -b busybox-slice https://github.com/vpa1977/chisel-releases /opt/chisel-releases     && git clone --depth 1 -b main https://github.com/canonical/  0.0s
 => CACHED [chisel 3/4] WORKDIR /opt/chisel                                                                                                                                              0.0s
 => CACHED [chisel 4/4] RUN go generate internal/deb/version.go     && go build ./cmd/chisel                                                                                             0.0s
 => CACHED [builder 3/3] COPY --from=chisel /opt/chisel/chisel /usr/bin/                                                                                                                 0.0s
 => CACHED [sliced-deps 1/3] COPY --from=chisel /opt/chisel-releases /opt/chisel-releases                                                                                                0.0s
 => CACHED [sliced-deps 2/3] RUN mkdir -p /rootfs     && chisel cut --release /opt/chisel-releases --root /rootfs     busybox_bins                                                       0.0s
 => CACHED [sliced-deps 3/3] RUN install -d -m 0755 -o 101 -g 101 /rootfs/home/app     && mkdir -p /rootfs/etc     && echo -e "root:x:0:\napp:x:101:" >/rootfs/etc/group     && echo -e  0.0s
 => CACHED [stage-3 1/2] COPY --from=sliced-deps /rootfs /                                                                                                                               0.0s
 => CACHED [stage-3 2/2] COPY --from=sliced-deps --chown=101:101 /rootfs/home/app /home/app                                                                                              0.0s
 => exporting to image                                                                                                                                                                   0.0s
 => => exporting layers                                                                                                                                                                  0.0s
 => => writing image sha256:55ee7d9f6af145e67ff7a447ddd0aad278b83becc06928c9d67dc3080076847e                                                                                             0.0s
 => => naming to docker.io/library/busybox                                                    

$ docker run -it busybox

BusyBox v1.30.1 (Ubuntu 1:1.30.1-7ubuntu3) built-in shell (ash)
Enter 'help' for a list of built-in commands.

/ $ ls
bin    dev    etc    home   lib    lib64  proc   sys    usr
/ $ 
````

Thanks for considering the MR!!!

### CODE REVIEW

**Review by cjdcordeiro — CHANGES_REQUESTED** [review]
Thanks!

While I get the point of the symlinks, this strategy only works well if `busybox` is being installed by itself. However:
 - the symlinks are not in the deb nor any of its maintainer scripts. We should stay true to that.
 - this will potentially conflict with other slices which actually deliver those binaries (like `coreutils`).

It feels safer to drop all the symlinks, and in your Dockerfile you'd instead have:

```dockerfile
FROM scratch
ARG USER
ARG UID
ARG GID
USER $UID:$GID
# Workaround for https://github.com/moby/moby/issues/38710
COPY --from=sliced-deps --chown=$UID:$GID /rootfs/home/$USER /home/$USER
ENTRYPOINT ["/bin/busybox"]
CMD ["sh"]
```

vpa1977:
> Thanks!
> 
> While I get the point of the symlinks, this strategy only works well if `busybox` is being installed by itself. However:
> 
> * the symlinks are not in the deb nor any of its maintainer scripts. We should stay true to that.
> * this will potentially conflict with other slices which actually deliver those binaries (like `coreutils`).
> 
> It feels safer to drop all the symlinks, and in your Dockerfile you'd instead have:
> 
Thank you!!! 
Updated as per comment.

cjdcordeiro (approved):
Short and sweet. thanks

slices/busybox.yaml
```diff
@@ -0,0 +1,8 @@
+package: busybox
+
+slices:
+  bins:
+    essential:
+      - libc6_libs
```
woky, line 6: Why? :thinking: 

```
# ldd /usr/bin/busybox
	not a dynamic executable
```

cjdcordeiro: That would probably need digging around the source code, cause:
```bash
$ apt info busybox | grep Depends

Depends: libc6 (>= 2.34)
```

vpa1977: Needs it for compiling, but not for runtime. Thanks @woky !!!! =)

woky: Oh, loader, you are right. Definitely a good idea to make a loader only libc slice.

woky (approved):

PR #40, opened 2023-07-12
feat: Node.js 18 LTS in Lunar (23.04)

liushuyu:
This pull request adds the Node.js 18 (LTS) chisel definitions for 23.04.

The reason why I chose 23.04 is that Node.js 18 is currently the latest LTS release (as of writing), and 23.04 is the first Ubuntu release that comes with Node.js 18.

## What's inside

- Node.js 18
- Library dependencies for Node.js
- JavaScripts dependencies for Node.js REPL

## Explanation

This container image aims for server-backend applications, so it has no extra NPM dependencies (not even `npm`).

For the users of this image, it's recommended to use an SDK container image (like [this one](https://hub.docker.com/layers/library/node/18-bullseye/images/sha256-db5230f47146c54cdcc52988fea502fb581997107c710e6e5700cce97f97f049?context=explore)) and also a bundler (like [`esbuild`](https://esbuild.github.io/), [`parcel`](https://parceljs.org/) or [`rspack`](https://www.rspack.dev/)) to build minified and/or bundled JavaScript application bundles and then copy the final production-ready JS bundles into this chiselled image for the best result. If the application contains Node.js native addons, it's recommended to verify if they are indeed used by the application (since it may introduce ABI issues).


valentincanonical:
Hey @liushuyu , many thanks for your contribution. That's a very exciting and valuable set of slices for chiselled Ubuntu to have! I'll let the team review the PR and make suggestions. Great job! 🎉

slices/node-acorn.yaml
cjdcordeiro, line 1: is there a way these libs could be split by functional unit? I.e., for library packages, we often simply create a `libs` slice, and that's ok, cause it's a lib pkg.

In this case though (very much like [this](https://github.com/canonical/chisel-releases/pull/31/files#diff-842f11960fe3c1253fde54d36da727142140e78e5b6baae453b536eb3685ecaa)) it would be nice to design slices by functionality. I don't know much about the pkg, so it might not be possible, but please consider it. The goal is that one can still have a very minimal node slice that is functional, and then other slices with paths that are only required for very specific purposes (e.g. for SSL-driven operation, for profiling, for mathematical computations, etc.)

liushuyu: > In this case though (very much like [this](https://github.com/canonical/chisel-releases/pull/31/files#diff-842f11960fe3c1253fde54d36da727142140e78e5b6baae453b536eb3685ecaa)) it would be nice to design slices by functionality. I don't know much about the pkg, so it might not be possible, but please consider it. The goal is that one can still have a very minimal node slice that is functional, and then other slices with paths that are only required for very specific purposes (e.g. for SSL-driven operation, for profiling, for mathematical computations, etc.)

Unfortunately, that is impossible for Node.js.
Node.js only has one binary and one essential library (v8 JavaScript Engine and its Node interop code) so they can't be split any further.

Node.js does not have much built-in stuff, most of the functionality is coming from user code (and/or NPM).

liushuyu:
@woky I am unsure which Ubuntu series you are trying to chisel on, but here is my Dockerfile for generating the image:

```dockerfile
ARG UBUNTU_RELEASE=23.04
ARG USER=ubuntu UID=101 GROUP=ubuntu GID=101

FROM golang:1.18 AS chisel
ARG UBUNTU_RELEASE
RUN git clone -b ubuntu-${UBUNTU_RELEASE} https://github.com/liushuyu/chisel-releases /opt/chisel-releases \
    && git clone --depth 1 -b main https://github.com/canonical/chisel /opt/chisel
WORKDIR /opt/chisel
RUN go generate internal/deb/version.go \
    && go build ./cmd/chisel

FROM ubuntu:$UBUNTU_RELEASE AS builder
SHELL ["/bin/bash", "-oeux", "pipefail", "-c"]
RUN apt-get update \
    && DEBIAN_FRONTEND=noninteractive apt-get install -y ca-certificates \
    && apt-get clean -y \
COPY --from=chisel /opt/chisel/chisel /usr/bin/

FROM builder AS rootfs-prep
ARG USER UID GROUP GID
RUN mkdir -p /rootfs/etc \
    && echo "$GROUP:x:$GID:" >/rootfs/etc/group \
    && echo "$USER:x:$UID:$GID::/nohome:/noshell" >/rootfs/etc/passwd

FROM scratch AS image-prep
ARG UID GID
USER $UID:$GID

### BOILERPLATE END ###

FROM rootfs-prep AS sliced-deps
COPY --from=chisel /opt/chisel-releases /opt/chisel-releases
RUN chisel cut --release /opt/chisel-releases --root /rootfs \
    base-files_base \
    base-files_release-info \
    tzdata_zoneinfo \
    ca-certificates_data \
    libgcc-s1_libs \
    libc6_libs \
    nodejs_bins

FROM image-prep
```

I have tested that the generated image is functional.

woky:
>```
> ...
> RUN chisel cut --release /opt/chisel-releases --root /rootfs \
>     base-files_base \
>     base-files_release-info \
>     tzdata_zoneinfo \
>     ca-certificates_data \
>     libgcc-s1_libs \
>     libc6_libs \
>     nodejs_bins
> 
> FROM image-prep
> COPY --from=sliced-deps /rootfs /
> ```
> 
> I have tested that the generated image is functional.

It works with the slices you have on the run command line. The slice that is buggy is `tzdata_zoneinfo-icu` which is also introduced by this PR. Try to add it to your command line.

slices/node-acorn.yaml
woky: I agree with @liushuyu. There are no sensible cuts possible for nodejs.

woky (changes requested):
As noted above.

woky:
Hey @liushuyu. Works now. Thanks!

woky (approved):

slices/node-acorn.yaml
```diff
@@ -0,0 +1,35 @@
+package: node-acorn
+
+slices:
+  libs:
+    essential:
+      - node-xtend_libs
+    contents:
+      # basically avoiding all the .map files (line tables) and unittest drivers
+      /usr/share/nodejs/acorn/**:
```
cjdcordeiro, line 9:
```suggestion
      # do you need /usr/share/nodejs/acorn/dist/bin.js? If not, the following pattern needs to be fine tuned
      /usr/share/nodejs/acorn/dist/*:
      /usr/share/nodejs/acorn/package.json:
```

This glob is bringing in `/usr/share/nodejs/acorn/bin/acorn`, which I guess you don't want? At least not in this slice.

slices/node-acorn.yaml
cjdcordeiro: I see. So, basically, there's no way one could run a nodejs app that only uses only the `/usr/share/nodejs/acorn-numeric-separator*` and `/usr/share/nodejs/acorn-bigint/*` libs right?

slices/node-acorn.yaml
```diff
@@ -0,0 +1,35 @@
...
+      /usr/share/nodejs/acorn-walk/**:
+      /usr/share/nodejs/acorn-static-class-features/index.js:
+      /usr/share/nodejs/acorn-static-class-features/package.json:
+      /usr/share/nodejs/acorn-private-methods/index.js:
+      /usr/share/nodejs/acorn-private-methods/package.json:
+      /usr/share/nodejs/acorn-private-methods/dist/*.js:
+      /usr/share/nodejs/acorn-private-methods/dist/*.mjs:
+      /usr/share/nodejs/acorn-private-class-elements/**:
+      /usr/share/nodejs/acorn-numeric-separator/dist/*.js:
+      /usr/share/nodejs/acorn-numeric-separator/dist/*.mjs:
```
cjdcordeiro, line 19: nit: I **think** the `mjs` and `js` lines could be merged into just 1, like:

`/usr/share/nodejs/acorn-numeric-separator/dist/*.*js:`

but I haven't tested...

slices/node-busboy.yaml
```diff
@@ -0,0 +1,7 @@
+package: node-busboy
+
+slices:
+  libs:
+    contents:
+      /usr/share/nodejs/busboy/lib/**:
```
cjdcordeiro, line 6: I think you're missing `/usr/share/nodejs/busboy/package.json`?

slices/node-busboy.yaml
cjdcordeiro, line 1: what about `/usr/share/nodejs/@types/busboy/*`?

slices/node-undici.yaml
```diff
@@ -0,0 +1,13 @@
+package: node-undici
+
+slices:
+  libs:
+    essential:
+      - node-busboy_libs
+    contents:
```
cjdcordeiro, line 7: you've kept some files out (like `/usr/share/nodejs/undici/index.d.ts`, `/usr/share/nodejs/undici/types/*`). If that was the intention, could you please add a comment explaining the rationale?

cjdcordeiro: P.s. I noticed that most of these left-out files are TypeScript, but at the same time, both in this slice and other lib slices above, .ts files are often included alongside.

slices/nodejs.yaml
```diff
@@ -0,0 +1,12 @@
+package: nodejs
+
+slices:
+  bins:
+    essential:
```
cjdcordeiro, line 5: I know it already comes from `libnode108`, but still, when it comes to dependencies I believe we should stay true to the deb, and in this case, `nodejs` needs `libc6` apparently

slices/nodejs.yaml
```diff
@@ -0,0 +1,12 @@
...
+      - libnode108_libs
+      - node-cjs-module-lexer_libs
+      - node-acorn_libs
+      - node-undici_libs
```
cjdcordeiro, line 9: right, so these are the dependencies I flagged missing, above in https://github.com/canonical/chisel-releases/pull/40/files#diff-142d24a2227a0d5f6f093889d9e9eed94ec886e90242dbf2c7c56bc391775959. 

Why have you included them here instead of `libnode108`? According to the deb, they should come with the latter and not `nodejs`.

slices/nodejs.yaml
cjdcordeiro, line 1: no need for `/usr/share/nodejs/@types/**`?

cjdcordeiro: what about `/usr/lib/x86_64-linux-gnu/node_modules` and `/usr/share/node_modules`?

cjdcordeiro (changes requested):
Thanks again for all of this. It will be an awesome addition to the portfolio of chiselled images. I'm leaving a few comments, and after this pass, I'll try to run some sort of hello world just to get acquainted with it :)

slices/node-busboy.yaml
liushuyu: > what about `/usr/share/nodejs/@types/busboy/*`?

Types definitions are not needed during the runtime. They are only required during development (and only if you are using TypeScript or `flow`).

slices/node-undici.yaml
liushuyu: Node.js can not run TypeScript files. First, they must be compiled into JavaScript files, which will strip all the type information. So typing data is not required during the runtime.

slices/nodejs.yaml
liushuyu: I have explained in the descriptions that:

> This container image aims for server-backend applications, so it has no extra NPM dependencies (not even npm).

It's expected that the user should bundle their dependencies (which is usually how Node.js developers deploy their apps). So extra dependencies are not included.

Typing information and TypeScript sources are not required for the application to run correctly after deployment. They are only needed when you run `esbuild` or `tsc` during the development and/or the bundle generation/deployment phase.

slices/node-acorn.yaml
liushuyu: > I see. So, basically, there's no way one could run a nodejs app that only uses only the `/usr/share/nodejs/acorn-numeric-separator*` and `/usr/share/nodejs/acorn-bigint/*` libs right?

Those are for Node.js itself (`acron` is the de-facto JavaScript parser in the Node.js ecosystem at the moment), and not for the user applications.

slices/node-busboy.yaml
liushuyu: > I think you're missing `/usr/share/nodejs/busboy/package.json`?

I have audited the code, and `busboy` does not need the `package.json` to function correctly.
Since the user application will not (and should not) use any Node modules from the system, `package.json` is not required.

slices/nodejs.yaml
liushuyu: Done

slices/node-acorn.yaml
liushuyu: Done

slices/nodejs.yaml
liushuyu: Done

slices/libnode108.yaml
```diff
@@ -0,0 +1,17 @@
+package: libnode108
+
+slices:
+  libs:
+    essential:
```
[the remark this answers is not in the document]
liushuyu: `libatomic1` is only needed for `mips` and `riscv64`.
I think the chisel tool does not support architecture-specific dependency. So do we want to support that CPU architecture for our image?

slices/node-acorn.yaml
liushuyu: I think I will just keep the pattern more specific. Since `*.*js` may also match some potentially unneeded files in the future.

slices/tzdata.yaml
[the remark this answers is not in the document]
liushuyu: I have attempted that, and unfortunately, in 23.04, the `tzdata` package has undergone some significant changes, so the slice for 22.04 can't be applied even after some modifications.

liushuyu:
By the way, I have opened a pull request (https://github.com/SchemaStore/schemastore/pull/3149) to add the Chisel slices schema to http://schemastore.org/.

slices/node-undici.yaml
cjdcordeiro: Ok thanks. So then can you pls either remove all `.ts` files from slices, or add a comment to the slices have have or don't have them, explaining the rationale?

slices/nodejs.yaml
cjdcordeiro: Why `libgcc-s1_libs` too? It's not listed in the deb...

slices/tzdata.yaml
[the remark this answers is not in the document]
cjdcordeiro: The overall slice design must still be consistent, otherwise it will be a mess for people migrating their chiselled applications between Ubuntu releases. Examples:
 - `zoneinfo-base` doesn't exist in 22.04
 - .tab files are required by most slices and don't belong to the `base` slice
 - ...etc.

I recommend starting from the 22.04 slice and adjusting it so it works for 23.04

slices/libc-ares2.yaml
```diff
@@ -0,0 +1,8 @@
+package: libc-ares2
+
+slices:
+  libs:
+    essential:
+      - libc6_libs
+    contents:
+      /usr/lib/*-linux-*/libcares.so.*:
```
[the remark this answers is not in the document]
cjdcordeiro: This is marked as "Resolved" but nothing has changed

slices/libnghttp2-14.yaml
```diff
@@ -0,0 +1,8 @@
+package: libnghttp2-14
+
+slices:
+  libs:
+    essential:
+      - libc6_libs
+    contents:
+      /usr/lib/*-linux-*/libnghttp2.so.*:
```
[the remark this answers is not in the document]
cjdcordeiro: Also resolved but unchanged

slices/libnode108.yaml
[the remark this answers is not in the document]
cjdcordeiro: True, good point. Can you please add a comment in the slice to point that out?

slices/libnode108.yaml
```diff
@@ -0,0 +1,17 @@
...
+      - libstdc++6_libs
+      - zlib1g_libs
+      - libssl3_libs
+      - libuv1_libs
+      - libbrotli1_libs
+      - libicu72_libs
+      - libc-ares2_libs
+      - libnghttp2-14_libs
+    contents:
+      /usr/lib/*-linux-*/libnode.so.*:
```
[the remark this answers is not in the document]
cjdcordeiro: Also not resolved

slices/libuv1.yaml
```diff
@@ -0,0 +1,8 @@
+package: libuv1
+
+slices:
+  libs:
+    essential:
+      - libc6_libs
+    contents:
+      /usr/lib/*-linux-*/libuv.so.*:
```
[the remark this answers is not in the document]
cjdcordeiro: Also not resolved

slices/node-acorn.yaml
cjdcordeiro: What was done? I don't see any diffs

slices/node-busboy.yaml
cjdcordeiro: Ok, so what about all the other `package.json` files being included from different modules?

slices/libc-ares2.yaml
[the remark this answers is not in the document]
cjdcordeiro: Nvm, the force-push made it hard to track the diff

cjdcordeiro (changes requested):
(see previous exchange)

slices/nodejs.yaml
liushuyu: > Why `libgcc-s1_libs` too? It's not listed in the deb...

It's not listed in the binary package, but if you do a `ldd` on the binary, you will see the binary executable requires `libgcc` runtime.

slices/node-busboy.yaml
liushuyu: > Ok, so what about all the other `package.json` files being included from different modules?

Some other packages require `package.json` to work (certain logic depends on definitions in `package.json`).
This practice is usually not recommended since it increases the bundle size during the deployment process.
However, many packages on NPM still do this.

slices/node-undici.yaml
liushuyu: I will remove all the `.ts` files.

slices/nodejs.yaml
cjdcordeiro: yes but it is a nested dependency. `libnode108` should bring it in.

slices/node-busboy.yaml
cjdcordeiro: alright thanks for the explanation. Given the unpredictable rationale behind the inclusion of such files, I'll ask you to add comments on where those `package.json` files are "missing", such that we can remember the reasoning in the future if someone proposes adding those.

slices/node-undici.yaml
cjdcordeiro: ok thanks. feel free to resolve this once you've committed the changes ;)

slices/libnode108.yaml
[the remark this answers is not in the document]
liushuyu: Done

slices/nodejs.yaml
liushuyu: Done

slices/node-acorn.yaml
```diff
@@ -0,0 +1,35 @@
+package: node-acorn
+
+slices:
+  libs:
+    essential:
```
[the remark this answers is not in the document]
liushuyu: `acorn` is required by `libnode`, where the internal REPL requires it (REPL is inside `libnode`).

cjdcordeiro (changes requested):
thanks for the update @liushuyu . Some of the previous comments (like the `tzdata` slices) are still unresolved.

liushuyu:
I have found a bug in the `slices/openssl.yaml` slice and fixed it in 5f1dfe0f00b5b0dd19a2c583808fbf6e72460949.
This caused Node.js to be unable to initiate any SSL/TLS connections.

liushuyu:
Now I present to you a simple testcase:

`index.js`:
```javascript
async function test_fetch() {
  const { default: fetch } = await import("node-fetch");
  const resp = await fetch("https://snapcraft.io/");
  console.log(`Response: ${resp.url} returned HTTP ${resp.status} ${resp.statusText}`);
}

async function test_temporal() {
  const dayjs = require("dayjs");
  const utc = require("dayjs/plugin/utc");
  dayjs.extend(utc);

  const now = new Date();
  const d_now = dayjs(now);
  console.log(d_now.utc().format());
}

Promise.all([test_fetch(), test_temporal()]).then(() => {});
```

`package.json`:
```json
{
  "name": "test-app",
  "version": "1.0.0",
  "main": "index.js",
  "license": "MIT",
  "scripts": {
    "build": "esbuild --target=node18 --bundle --minify --platform=node index.js --outfile=dist/index.js"
  },
  "dependencies": {
    "dayjs": "^1.11.9",
    "esbuild": "^0.19.2",
    "node-fetch": "^3.3.2"
  }
}
```
Make sure you have `nodejs` and `npm` installed on your host system and execute the following:

```bash
npm ci
npm run build
```

This will build the JavaScript to `dist/index.js`. Please copy this file to the container and run it.
The expected result should look like:
```
2023-09-05T18:59:24Z
Response: https://snapcraft.io/ returned HTTP 200 
```

slices/node-undici.yaml
liushuyu: All TypeScript stubs removed.

slices/node-busboy.yaml
liushuyu: I have now included all the `package.json` files for consistency.
Also, it might make more sense for unbuilt/non-bundled JavaScript modules to have a `package.json` file for the entry point information.

slices/tzdata.yaml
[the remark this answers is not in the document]
liushuyu: Done.

liushuyu:
The slices I used in my `Dockerfile`:
```dockerfile
FROM rootfs-prep AS sliced-deps
COPY --from=chisel /opt/chisel-releases /opt/chisel-releases
RUN chisel cut --release /opt/chisel-releases --root /rootfs \
    base-files_base \
    base-files_release-info \
    tzdata_zoneinfo \
    ca-certificates_data \
    openssl_config \
    libssl3_libs \
    nodejs_bins
```

cjdcordeiro:
> I have found a bug in the `slices/openssl.yaml` slice and fixed it in [5f1dfe0](https://github.com/canonical/chisel-releases/commit/5f1dfe0f00b5b0dd19a2c583808fbf6e72460949). This caused Node.js to be unable to initiate any SSL/TLS connections.

Thanks for the fix @liushuyu 

I've edited the file to place `/usr/lib/ssl/cert.pem` in a `data` slice instead, since that's a symlink to a data file from ca-certificates. Hope that's ok

cjdcordeiro:
fyi @liushuyu , during the next sprint, we'll be creating a project for this chiselled image in https://github.com/ubuntu-rocks, alike others (chiselled-python, chiselled-jre, etc).

Do you want to be added as a contributor?

liushuyu:
> fyi @liushuyu , during the next sprint, we'll be creating a project for this chiselled image in https://github.com/ubuntu-rocks, alike others (chiselled-python, chiselled-jre, etc).
> 
> Do you want to be added as a contributor?

I can maintain this image as a side-project, so yes.

liushuyu:
> I've edited the file to place `/usr/lib/ssl/cert.pem` in a `data` slice instead, since that's a symlink to a data file from ca-certificates. Hope that's ok

I think in this case, we will need a to modify the slice to something like this:

```yaml
package: openssl

slices:
  libs:
      - libc6_libs
      - libc6_config
      - libssl3_libs
      - openssl_config
      - openssl_data
  bins:
    essential:
      - openssl_libs
    contents:
      /usr/bin/c_rehash:
      /usr/bin/openssl:

  config:
    contents:
      /etc/ssl/private/:
      /etc/ssl/openssl.cnf:
      /usr/lib/ssl/certs:
      /usr/lib/ssl/openssl.cnf:
      /usr/lib/ssl/private:

  data:
    contents:
      /usr/lib/ssl/cert.pem:
```

... so that if the applications only need the libraries and data, they can include libraries and data without the OpenSSL tools.

cjdcordeiro:
that sounds reasonable @liushuyu -> https://github.com/canonical/chisel-releases/issues/60

PR #41, opened 2023-07-12
feat: add basic slice definitions files for Focal

cjdcordeiro:
This PR backports (from 22.04) and fits a few (the most basic and essential) slice definitions files into the new Focal Chisel release.

The affected packages are:
 - base-files
 - base-passwd
 - ca-certificates
 - libc6
 - libgcc-s1

The choice of packages was driven by our [Chiselled Ubuntu base image](https://ubuntu.com/blog/craft-custom-chiselled-ubuntu-distroless).

### DIFF
--- a/slices/libc6.yaml
+++ b/slices/libc6.yaml
@@ -1,0 +1,51 @@
+package: libc6
+
+slices:
+  config:
+    contents:
+      /etc/ld.so.conf.d/*-linux-*.conf:
+
+  libs:
+    contents:
+      /lib/*-linux-*/ld-*.*.so:
+      /lib/*-linux-*/ld-*.so.*:
+      /lib/*-linux-*/libBrokenLocale-*.*.so:
+      /lib/*-linux-*/libBrokenLocale.so.*:
+      /lib/*-linux-*/libSegFault.so:
+      /lib/*-linux-*/libanl-*.*.so:
+      /lib/*-linux-*/libanl.so.*:
+      /lib/*-linux-*/libc-*.*.so:
+      /lib/*-linux-*/libc.so.*:
+      /lib/*-linux-*/libdl-*.*.so:
+      /lib/*-linux-*/libdl.so.*:
+      /lib/*-linux-*/libm-*.*.so:
+      /lib/*-linux-*/libm.so.*:
+      /lib/*-linux-*/libmemusage.so:
+      /lib/*-linux-*/libmvec-*.*.so: {arch: amd64}
+      /lib/*-linux-*/libmvec.so.*: {arch: amd64}
+      /lib/*-linux-*/libnsl-*.*.so:
+      /lib/*-linux-*/libnsl.so.*:
+      /lib/*-linux-*/libnss_compat-*.*.so:
+      /lib/*-linux-*/libnss_compat.so.*:
+      /lib/*-linux-*/libnss_dns-*.*.so:
+      /lib/*-linux-*/libnss_dns.so.*:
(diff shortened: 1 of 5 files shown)


woky (commented):
@cjdcordeiro  Did you just copy it from other release? AFAIK some libc6 libraries in jammy are not in focal?

cjdcordeiro:
> @cjdcordeiro Did you just copy it from other release? AFAIK some libc6 libraries in jammy are not in focal?

@woky copied and changed. `libc6` had to be completely changed yes. The libraries diff isn't that significant, but the main difference is the naming convention.

slices/libc6.yaml
```diff
@@ -0,0 +1,32 @@
...
+    contents:
+      /lib/*-linux-*/ld*.so*:
+      /lib/*-linux-*/libBrokenLocale*.so*:
+      /lib/*-linux-*/libSegFault.so:
+      /lib/*-linux-*/libanl*.so*:
+      /lib/*-linux-*/libc*.so*:
+      /lib/*-linux-*/libdl*.so*:
+      /lib/*-linux-*/libm*.so*:
+      /lib/*-linux-*/libmemusage.so:
+      /lib/*-linux-*/libmvec*.so*: {arch: amd64}
```
rebornplusplus, line 18: This will cause some problems. Notice how `libm*.so*` is written. As I mentioned in https://github.com/canonical/chisel/pull/62#pullrequestreview-1432159069, if two globs (or, two path definitions) refer to the same path, chisel may raise an error saying it didn't find any content at one of the paths.

In this case, I have the following error. Note that this error changes due to the behaviour I mentioned in https://github.com/canonical/chisel/pull/62#pullrequestreview-1432159069, and sometimes passes as well if it can properly map a package path to the globs mentioned in slice definition file.

```
error: cannot extract from package "libc6": no content at:
- /lib/*-linux-*/libmemusage.so
- /lib/*-linux-*/libmvec*.so*
```

I suppose for now, until this random behaviour is fixed, the rule of thumb could be not to write two paths who may be same. In this case, `/lib/*-linux-*/libm*.so*` encapsulates `/lib/*-linux-*/libmemusage.so` and `/lib/*-linux-*/libmvec*.so*`.

rebornplusplus (changes requested):
I have only briefly taken a look at `libc6`. Will be thorough with the new changes. :) 

PS. This reminded me. Perhaps we should have a basic test action in this repo which runs Chisel on the proposed slices and gives out a preliminary verdict.

slices/libc6.yaml
rebornplusplus:
```sh
[PASS]  /lib/x86_64-linux-gnu/libmemusage.so : /lib/*-linux-*/libm*.so*
[PASS]  /lib/x86_64-linux-gnu/libmvec-2.31.so : /lib/*-linux-*/libm*.so*
[PASS]  /lib/x86_64-linux-gnu/libmvec.so.1 : /lib/*-linux-*/libm*.so*
```
These are additional log messages I had while debugging. As I described above, Chisel thinks the `libmemusage.so` and `libmvec*.so*` files come from `libm*.so*`.

cjdcordeiro: @rebornplusplus I've updated `libc6`. Can y pls have another look?

P.S. wrt the tests - definitely. we have that in our backlog, we just need to find the time for it. In the meantime, if while reviewing these PRs you come up with ideas or scripts for useful checks to have, pls put them in our backlog ticket for future reference.

PR #42, opened 2023-07-13
feat: add basic slice definitions files for Focal

cjdcordeiro:
**Duplicate of https://github.com/canonical/chisel-releases/pull/41, with a different base branch.**

This PR backports (from 22.04) and fits a few (the most basic and essential) slice definitions files into the new Focal Chisel release.

The affected packages are:
 - base-files
 - base-passwd
 - ca-certificates
 - libc6
 - libgcc-s1

The choice of packages was driven by our [Chiselled Ubuntu base image](https://ubuntu.com/blog/craft-custom-chiselled-ubuntu-distroless).

### DIFF
--- a/slices/base-files.yaml
+++ b/slices/base-files.yaml
@@ -1,0 +1,61 @@
+package: base-files
+
+slices:
+  base:
+    essential:
+      - base-files_etc
+      - base-files_bin
+      - base-files_lib
+      - base-files_tmp
+      - base-files_var
+      - base-files_home
+
+  etc:
+    contents:
+      /etc/:
+
+  bin:
+    contents:
+      /bin/:
+      /sbin/:
+      /usr/bin/:
+      /usr/sbin/:
+
+  lib:
+    contents:
+      /lib/:
+      /usr/lib/:
+
+  tmp:
+    contents:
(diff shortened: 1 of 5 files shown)


rebornplusplus (approved):
Looks good to me now. Thanks.

Although I would have wished the `libc6_libs` slice definition could be a bit shortened; some of the lines may be merged together. But that'd just increase the likelihood of missed errors. It's better to be explicit, for now at least.

cjdcordeiro:
> Looks good to me now. Thanks.
> 
> Although I would have wished the `libc6_libs` slice definition could be a bit shortened; some of the lines may be merged together. But that'd just increase the likelihood of missed errors. It's better to be explicit, for now at least.

Yeah I think we should change a bit our approach towards the usage of globs. I don't think we should be using them as a way to  be "lazy" about writing the full paths or trying to merge multiple paths into one, but rather as a way to reduce path duplication due to metadata properties of the path (like the architecture or version). For example:
 - `/usr/lib/*/libfoo*`: IMO this is **bad**
 - `/usr/lib/*-linux-*/libfoo-2.*.so`: this is better

woky (approved):

PR #46, opened 2023-08-02
jammy: slices for chiselled OVN

javacruft:
Add required slices to produce a chiselled OVN ROCK.

### DIFF
--- a/slices/dash.yaml
+++ b/slices/dash.yaml
@@ -8,4 +8,3 @@ slices:
       - base-files_base
     contents:
       /bin/dash:
-      /bin/sh:
--- a/slices/dash.yaml
+++ b/slices/dash.yaml
@@ -1,7 +1,7 @@
 package: dash
 
 slices:
-  base:
+  bins:
     essential:
       - libc6_libs
       - base-passwd_data
--- a/slices/dash.yaml
+++ b/slices/dash.yaml
@@ -1,0 +1,11 @@
+package: dash
+
+slices:
+  base:
+    essential:
+      - libc6_libs
+      - base-passwd_data
+      - base-files_base
+    contents:
+      /bin/dash:
+      /bin/sh:
--- a/slices/libcap-ng0.yaml
+++ b/slices/libcap-ng0.yaml
@@ -5,4 +5,4 @@ slices:
     essential:
       - libc6_libs
     contents:
-      /lib/*-linux-*/libcap-ng.so.*:
+      /lib/*-linux-*/libcap-ng.so.0*:
--- a/slices/libcap-ng0.yaml
+++ b/slices/libcap-ng0.yaml
@@ -1,0 +1,8 @@
+package: libcap-ng0
+
+slices:
+  libs:
+    essential:
+      - libc6_libs
+    contents:
+      /lib/*-linux-*/libcap-ng.so.*:
--- a/slices/openvswitch-common.yaml
+++ b/slices/openvswitch-common.yaml
@@ -3,8 +3,9 @@ package: openvswitch-common
 slices:
   core:
     essential:
+      - coreutils_bins
       - sed_bins
-      - dash_base
+      - dash_bins
       - libc6_libs
       - libcap-ng0_libs
       - libssl3_libs
--- a/slices/openvswitch-common.yaml
+++ b/slices/openvswitch-common.yaml
@@ -1,0 +1,22 @@
+package: openvswitch-common
+
+slices:
+  core:
+    essential:
+      - sed_bins
+      - dash_base
+      - libc6_libs
+      - libcap-ng0_libs
+      - libssl3_libs
+      - libunbound8_libs
+      - openssl_bins
+    contents:
+      /usr/bin/ovsdb-tool:
+      /usr/sbin/ovsdb-server:
+      /usr/share/openvswitch/scripts/ovs-lib:
+  bins:
+    essential:
+      - openvswitch-common_core
+    contents:
+      /usr/bin/ovs-appctl:
+      /usr/bin/ovs-ofctl:
--- a/slices/ovn-central.yaml
+++ b/slices/ovn-central.yaml
@@ -1,0 +1,10 @@
+package: ovn-central
+
+slices:
+  base:
+    essential:
+      - openvswitch-common_bins
+      - ovn-common_bins
+    contents:
+      /usr/bin/ovn-northd:
+      /usr/share/ovn/*.ovsschema:
--- a/slices/ovn-central.yaml
+++ b/slices/ovn-central.yaml
@@ -1,10 +1,14 @@
 package: ovn-central
 
 slices:
-  base:
+  config:
+    contents:
+      /etc/default/ovn-central:
+      /usr/share/ovn/*.ovsschema:
+  core:
     essential:
       - openvswitch-common_bins
+      - ovn-central_config
       - ovn-common_bins
     contents:
       /usr/bin/ovn-northd:
-      /usr/share/ovn/*.ovsschema:
--- a/slices/ovn-common.yaml
+++ b/slices/ovn-common.yaml
@@ -1,0 +1,25 @@
+package: ovn-common
+
+slices:
+  core:
+    essential:
+      - coreutils_bins
+      - sed_bins
+      - dash_base
+      - libc6_libs
+      - libcap-ng0_libs
+      - libssl3_libs
+      - libunbound8_libs
+    contents:
+      /usr/share/ovn/scripts/ovn-ctl:
+      /usr/share/ovn/scripts/ovn-lib:
+      /var/run/ovn/: {make: true, mode: 02775}
+      /var/log/ovn/: {make: true, mode: 02775}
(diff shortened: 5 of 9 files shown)


slices/openvswitch-common.yaml
```diff
@@ -0,0 +1,23 @@
+package: openvswitch-common
+
+slices:
+  core:
+    essential:
+      - coreutils_bins
+      - sed_bins
+      - dash_bins
```
rebornplusplus, line 8: Are these needed by the files in this slice?

slices/openvswitch-common.yaml
rebornplusplus, line 1: Hmm, this is interesting... :thinking:
Is `openvswitch-common` package available for `i386` arch? I get the following error:
```
error: slice package "openvswitch-common" missing from archive
```

slices/ovn-common.yaml
```diff
@@ -0,0 +1,25 @@
+package: ovn-common
+
+slices:
+  core:
+    essential:
+      - coreutils_bins
+      - sed_bins
+      - dash_bins
```
rebornplusplus, line 8: Same comment as above. Why are they needed in this slice?

slices/ovn-common.yaml
```diff
@@ -0,0 +1,25 @@
...
+      - libc6_libs
+      - libcap-ng0_libs
+      - libssl3_libs
+      - libunbound8_libs
+    contents:
+      /usr/share/ovn/scripts/ovn-ctl:
+      /usr/share/ovn/scripts/ovn-lib:
+      /var/run/ovn/: {make: true, mode: 02775}
+      /var/log/ovn/: {make: true, mode: 02775}
+      /var/lib/ovn/: {make: true, mode: 02775}
```
rebornplusplus, line 18: Are these required? I checked out the package maintainer scripts and didn't notice anything similar.

slices/ovn-common.yaml
```diff
@@ -0,0 +1,25 @@
+package: ovn-common
+
+slices:
+  core:
```
rebornplusplus, line 4: Again asking for info because I don't have much info on the package. What is the rationale behind putting these files in core?

slices/ovn-common.yaml
```diff
@@ -0,0 +1,25 @@
...
+      /var/run/ovn/: {make: true, mode: 02775}
+      /var/log/ovn/: {make: true, mode: 02775}
+      /var/lib/ovn/: {make: true, mode: 02775}
+  bins:
+    essential:
+      - ovn-common_core
+    contents:
+      /usr/bin/ovn-appctl:
+      /usr/bin/ovn-nbctl:
+      /usr/bin/ovn-sbctl:
```
rebornplusplus, line 25: Some binaries in [this list](https://packages.ubuntu.com/jammy/amd64/ovn-common/filelist) seems missing:
```
/usr/bin/ovn-appctl
/usr/bin/ovn-detrace
/usr/bin/ovn-nbctl
/usr/bin/ovn-sbctl
/usr/bin/ovn-trace
```

slices/openvswitch-common.yaml
rebornplusplus: The `ovn-common` package seems missing as well, for `i386`.

slices/ovn-central.yaml
```diff
@@ -0,0 +1,10 @@
+package: ovn-central
+
+slices:
+  base:
+    essential:
+      - openvswitch-common_bins
+      - ovn-common_bins
+    contents:
+      /usr/bin/ovn-northd:
+      /usr/share/ovn/*.ovsschema:
```
rebornplusplus, line 10: Are these config files? If so, perhaps it's better to create a `config` slice for these and include it as an `essential`. I think `/etc/default/ovn-central` would probably also deserve to be there.

slices/ovn-central.yaml
```diff
@@ -0,0 +1,10 @@
+package: ovn-central
+
+slices:
+  base:
```
rebornplusplus, line 4: I would suggest renaming it to `core` to be more in line with other existing slice definitions.

slices/ovn-central.yaml
```diff
@@ -0,0 +1,10 @@
+package: ovn-central
+
+slices:
```
rebornplusplus, line 3: Should there be a `libs` (or `services`) slice as well? Aren't the following needed?
```
/lib/systemd/system/ovn-central.service
/lib/systemd/system/ovn-northd.service
/lib/systemd/system/ovn-ovsdb-server-nb.service
/lib/systemd/system/ovn-ovsdb-server-sb.service
```

slices/openvswitch-common.yaml
rebornplusplus: Same for `ovn-central`.

rebornplusplus (changes requested):
Hiya @javacruft, nice chiselling! I only have a few suggestions about conventions and naming. And a bunch of questions asking for more info. Overall, it's looking good. Let me know what you think about the comments. Cheers.

slices/dash.yaml
```diff
@@ -0,0 +1,11 @@
...
+
+slices:
+  bins:
+    essential:
+      - libc6_libs
+      - base-passwd_data
+      - base-files_base
+    contents:
+      /bin/dash:
+      /bin/sh:
```
rebornplusplus, line 11: Similar to my comment in https://github.com/canonical/chisel-releases/pull/61#discussion_r1326873419, we should probably drop this symlink from the slice definition file and add it manually afterwards, until chisel supports conflicting paths.

javacruft:
@rebornplusplus picking this up again in the next week or so - thanks for the feedback

slices/openvswitch-common.yaml
javacruft: i386 was dropped as a complete architecture in Ubuntu a few releases back - as such only a core set of libraries and components are built for this architecture.  OVN is not.

slices/openvswitch-common.yaml
javacruft: yep - ovs-lib is a script and very much depends on a) a functioning shell and b) misc other utilities for parsing arguments and figuring out what todo.

slices/dash.yaml
cjdcordeiro: well, IMO that makes sense in https://github.com/canonical/chisel-releases/pull/61#discussion_r1326873419 cause in that case the file is added via a maintainer script. In this case, the file is an integral part of the deb, and thus I believe it is safe to have it here

slices/ovn-central.yaml
javacruft: These are all systemd related service definitions which I don't think are relevant for ROCK's

slices/ovn-central.yaml
javacruft: Done

slices/ovn-central.yaml
javacruft: Done

slices/ovn-common.yaml
javacruft: Done

slices/ovn-common.yaml
javacruft: That's a good question - I'm not entirely clear on when a split should be made between bins and core - it might make more sense for this all to be in once slice as I'm not sure core is useful without bins.

slices/libcap-ng0.yaml
```diff
@@ -0,0 +1,8 @@
+package: libcap-ng0
+
+slices:
+  libs:
+    essential:
+      - libc6_libs
+    contents:
+      /lib/*-linux-*/libcap-ng.so.*:
```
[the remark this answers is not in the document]
javacruft: Done

slices/dash.yaml
```diff
@@ -0,0 +1,11 @@
+package: dash
+
+slices:
+  bins:
+    essential:
+      - libc6_libs
+      - base-passwd_data
+      - base-files_base
```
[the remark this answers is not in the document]
javacruft: When I tested the dash binary it needed files from both of these slices in order order to actually start - I'll recheck that as it was a month or two back I did the initial slicing.

slices/openvswitch-common.yaml
```diff
@@ -0,0 +1,23 @@
...
+      - libc6_libs
+      - libcap-ng0_libs
+      - libssl3_libs
+      - libunbound8_libs
+      - openssl_bins
+    contents:
+      /usr/bin/ovsdb-tool:
+      /usr/sbin/ovsdb-server:
+      /usr/share/openvswitch/scripts/ovs-lib:
```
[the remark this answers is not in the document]
javacruft: These files have re-use in the ovn slices  which use the OVS db tools and wrapper libs

cjdcordeiro:
@javacruft since it is still in draft mode, should we wait a bit longer before continuing with the reviews?

javacruft:
@cjdcordeiro yes please - I've not had a huge amount of time to re-focus on this topic and I appreciate chiseling and ROCKs have moved on around the work I originally did.

linostar:
Hey @javacruft

Are you still interested in picking up the PR and resuming from where you left off?

PR #47, opened 2023-08-03
[22.04] Add slice for libxml2

osener:
This change adds a new slice for [libxml2](https://packages.ubuntu.com/jammy/libxml2).

### DIFF
--- a/slices/libxml2.yaml
+++ b/slices/libxml2.yaml
@@ -8,4 +8,4 @@ slices:
       - liblzma5_libs
       - zlib1g_libs
     contents:
-      /usr/lib/*-linux-*/libxml2.so.*:
+      /usr/lib/*-linux-*/libxml2.so.2*:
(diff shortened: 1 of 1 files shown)


slices/libxml2.yaml
```diff
@@ -0,0 +1,11 @@
...
+
+slices:
+  libs:
+    essential:
+      - libc6_libs
+      - libicu70_libs
+      - liblzma5_libs
+      - zlib1g_libs
+    contents:
+      /usr/lib/*-linux-*/libxml2.so.*:
```
cjdcordeiro, line 11:
```suggestion
      /usr/lib/*-linux-*/libxml2.so.2*:
```

Just to be a bit more restrictive and thus safer from possible future conflicts :) 

Apart from this, it LGTM

cjdcordeiro (changes requested):

slices/libxml2.yaml
osener: Thank you!

cjdcordeiro (approved):
TYVM!

PR #49, opened 2023-08-08
[22.04] Add slices for libgnutls, libsoup and glib-networking

ozanvos:
This PR adds various networking libraries and their dependencies.


slices/gsettings-desktop-schemas.yaml
```diff
@@ -0,0 +1,7 @@
+package: gsettings-desktop-schemas
+
+slices:
+  data:
+    contents:
+      /usr/share/GConf/gsettings/*:
+      /usr/share/glib-2.0/schemas/*:
```
cjdcordeiro, line 7: same here. I believe this whole slice should be called `config`. `data` is more appropriate for something like `ca-certificates` which delivers data files like `/etc/ssl/certs/ca-certificates.crt`.

cjdcordeiro, line 7:
```suggestion
      /usr/share/glib-2.0/schemas/10_gsettings-desktop-schemas.gschema.override:
      /usr/share/glib-2.0/schemas/org.gnome.desktop.*.gschema.xml:
      /usr/share/glib-2.0/schemas/org.gnome.desktop.enums.xml:
      /usr/share/glib-2.0/schemas/org.gnome.system.*.gschema.xml:      
```

(I think this should work, but please double check :) )

So, like in previous comments, I prefer to encourage a conservative use of globs. Although they can make slices much smaller, at scale, they can also become a burden and hard to trace if they are too generic. Whenever possible, I'd try to find a sweet spot between simplicity, maintainability and explicitness. In this example, I've sacrificed 3 lines of the slice, but not only it makes the slice easier to read, but it also saves us from potential collisions with other packages in the future.

slices/libgnutls30.yaml
```diff
@@ -0,0 +1,16 @@
...
+    essential:
+      - libc6_libs
+      - libgmp10_libs
+      - libhogweed6_libs
+      - libidn2-0_libs
+      - libnettle8_libs
+      - libp11-kit0_libs
+      - libtasn1-6_libs
+      - libunistring2_libs
+
```
cjdcordeiro, line 14:
```suggestion
```

slices/libsoup2.4-1.yaml
```diff
@@ -0,0 +1,16 @@
+package: libsoup2.4-1
+
+slices:
+  libs:
+    essential:
```
cjdcordeiro, line 5: I think you're missing a dependency to `glib-networking_libs`

cjdcordeiro: also for `libsoup2.4-common` and `libxml2` (you can skip the former though as it doesn't bring any libraries)

slices/libsoup2.4-1.yaml
```diff
@@ -0,0 +1,16 @@
...
+      - libbrotli1_libs
+      - libc6_libs
+      - libglib2.0-0_core
+      - libgssapi-krb5-2_libs
+      - libicu70_libs
```
cjdcordeiro, line 10: I don't think this is a dependency
```suggestion
```

slices/libsoup2.4-1.yaml
```diff
@@ -0,0 +1,16 @@
...
+      - liblzma5_libs
```
cjdcordeiro, line 11: same here
```suggestion
```

slices/libsoup2.4-1.yaml
```diff
@@ -0,0 +1,16 @@
...
+      - libpsl5_libs
+      - libsqlite3-0_libs
+      - zlib1g_libs
+    contents:
+      /usr/lib/*-linux-*/libsoup-2.4.so.*:
```
cjdcordeiro, line 16:
```suggestion
      /usr/lib/*-linux-*/libsoup-2.4.so.1*:
```

slices/publicsuffix.yaml
```diff
@@ -0,0 +1,6 @@
+package: publicsuffix
+
+slices:
+  data:
+    contents:
+      /usr/share/publicsuffix/*:
```
cjdcordeiro, line 6: this path seems to be specific to this pkg, so the glob is a bit safer to use here. Nonetheless, given that there are only 3 files within, it is safer to just list them explicitly :)

cjdcordeiro (changes requested):
Thanks for the contribution @osener . This is a very welcome addition :)

Although functional, I've left a few requests that relate to the design of slices. The overall message is that we should be as conservative as possible when it comes to the use of globs. With these changes, it should be ready to be merged.

rebornplusplus:
Hi @ozanmakes, it has been some time since the last activity here. The PR could use a rebase/merge since there are some conflicts already. There are some changes requested as well. Will you please let me know if you have some time to go over the comments and update the PR? Thank you!

slices/gsettings-desktop-schemas.yaml
cjdcordeiro: these are all ascii files, so let's keep data as a catch all

clay-lake:
Hi Everyone, this PR has been quite for a while. I'd like to keep some of these changes, so I made a reworked PR here #379. If the rework is merged I will close this PR.

cjdcordeiro:
> Hi everyone, this PR has been quite for a while. I'd like to keep some of these changes, so I made a reworked PR here #379. If the rework is merged I will close this PR.

Thanks and ACK!

rebornplusplus:
Closed by #389.

PR #54, opened 2023-08-22
feat(focal): add slice definitions file for ncurses-base

rebornplusplus:
This PR adds the slice definitions file for the package `ncurses-base`. I will be needing this for Python3.8 interactive interpreter.

### DIFF
--- a/slices/ncurses-base.yaml
+++ b/slices/ncurses-base.yaml
@@ -1,0 +1,21 @@
+package: ncurses-base
+
+slices:
+  libs:
+    contents:
+      /lib/terminfo/E/Eterm*:
+      /lib/terminfo/a/ansi:
+      /lib/terminfo/c/cons25*:
+      /lib/terminfo/c/cygwin:
+      /lib/terminfo/d/dumb:
+      /lib/terminfo/h/hurd:
+      /lib/terminfo/l/linux:
+      /lib/terminfo/m/mach*:
+      /lib/terminfo/p/pcansi:
+      /lib/terminfo/r/rxvt*:
+      /lib/terminfo/s/screen*:
+      /lib/terminfo/s/sun:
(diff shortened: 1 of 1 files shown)


cjdcordeiro (commented):
thanks for this :)

is it possible to split it further? E.g. I don't see a use for `/lib/terminfo/c/cygwin` in most of our use cases. Same for `/lib/terminfo/s/screen*`.

rebornplusplus:
Hiya, I have added a few more slices. I was wondering if I should merge a few of those together, but since most of these are intended to be used on a need-only basis, it made sense to create a slice for each of the similar terminals.

slices/ncurses-base.yaml
```diff
@@ -0,0 +1,88 @@
+package: ncurses-base
+
+slices:
+  # the "data" slice contains all the /lib/terminfo files
+  data:
```
cjdcordeiro, line 5: are they all binary data files? (making sure `data` is the right slice name for it)

cjdcordeiro (commented):
thanks for the split. Despite being laborious, I feel it is not controversial since each slice corresponds to a terminal type.

To validate that these slices are in fact self-contained, for our ongoing python image, which slice do you need?

slices/ncurses-base.yaml
rebornplusplus: Yeah, they are all compiled terminfo entry. Contains binary data inside them. For example:

```sh
$ file /lib/terminfo/x/xterm
/lib/terminfo/x/xterm: Compiled terminfo entry "xterm"
```

rebornplusplus:
> To validate that these slices are in fact self-contained, for our ongoing python image, which slice do you need?

`ncurses-base_xterm` seems to do the trick for the issue we had.

cjdcordeiro (approved):
great! thanks

PR #56, opened 2023-08-29
[22.04] Add libasan6

timo-reymann:
Add slice for [libasan6](https://packages.ubuntu.com/focal/libasan6)

### DIFF
--- a/slices/libasan6.yaml
+++ b/slices/libasan6.yaml
@@ -6,5 +6,6 @@ slices:
       - libc6_libs
       - libstdc++6_libs
       - libgcc-s1_libs
+    # - gcc-11-base_libs Although gcc-11-base is a package dependency, it is not needed for this libs slice and is omitted.
     contents:
       /usr/lib/*-linux-*/libasan.so.6*:
--- a/slices/libasan6.yaml
+++ b/slices/libasan6.yaml
@@ -6,6 +6,6 @@ slices:
       - libc6_libs
       - libstdc++6_libs
       - libgcc-s1_libs
-    # - gcc-11-base_libs Although gcc-11-base is a package dependency, it is not needed for this libs slice and is omitted.
+    # Although gcc-11-base is a package dependency, it is not needed for this libs slice and is thus omitted.
     contents:
       /usr/lib/*-linux-*/libasan.so.6*:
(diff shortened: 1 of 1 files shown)


slices/libasan6.yaml
```diff
@@ -0,0 +1,9 @@
+package: libasan6
+
+slices:
+  libs:
+    essential:
```
cjdcordeiro, line 5: the dependencies should be `libc6`, `libgcc-s1` and `gcc-11-base` instead.

slices/libasan6.yaml
```diff
@@ -0,0 +1,9 @@
...
+      - libc6_libs
+      - libstdc++6_libs
+    contents:
+      /usr/lib/x86_64-linux-gnu/libasan.so*:
```
cjdcordeiro, line 9:
```suggestion
      /usr/lib/*-linux-*/libasan.so.6*:
```

You need the first two `*` to be able to cope with other architectures.

cjdcordeiro (changes requested):
Thank you very much for the contribution @timo-reymann. A couple of changes are needed, but it should be ok after that ;)

slices/libasan6.yaml
timo-reymann: Added the `libgcc-s1` dependency. I am unable to find `gcc-11-base`, in the slices folder? Do I miss something? 😅

cjdcordeiro: It doesn't exist yet. 

**However**, since it is only a collection of `/usr/share/doc/` files, it isn't really needed for the slice you're proposing. So I recommend simply adding a comment saying precisely that - "Although `gcc-11-base` is a package dependency, it is not needed for this `libs` slice"

timo-reymann: Alrighty, done :)

slices/libasan6.yaml
```diff
@@ -0,0 +1,11 @@
+package: libasan6
+
+slices:
+  libs:
+    essential:
+      - libc6_libs
+      - libstdc++6_libs
+      - libgcc-s1_libs
+    # - gcc-11-base_libs Although gcc-11-base is a package dependency, it is not needed for this libs slice and is omitted.
```
cjdcordeiro, line 9:
```suggestion
    # Although gcc-11-base is a package dependency, it is not needed for this libs slice and is thus omitted.
```

PR #57, opened 2023-08-29
[22.04] Add libgomp1

timo-reymann:
Add [libgomp1](https://packages.ubuntu.com/jammy/libgomp1) slice

### DIFF
--- a/slices/libgomp1.yaml
+++ b/slices/libgomp1.yaml
@@ -6,4 +6,4 @@ slices:
       - libc6_libs
     # Although gcc-12-base is a package dependency, it is not needed for this libs slice and is thus omitted.
     contents:
-      /usr/lib/*-linux-gnu/libgomp.so.*:
+      /usr/lib/*-linux-*/libgomp.so.1*:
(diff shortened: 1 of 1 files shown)


slices/libgomp1.yaml
```diff
@@ -0,0 +1,9 @@
+package: libgomp1
+
+slices:
+  libs:
+    essential:
+      - libc6_libs
+    # Although gcc-12-base is a package dependency, it is not needed for this libs slice and is thus omitted.
+    contents:
+      /usr/lib/*-linux-gnu/libgomp.so.*:
```
cjdcordeiro, line 9:
```suggestion
      /usr/lib/*-linux-*/libgomp.so.1*:
```

`gnu` need to be replaced by `*` otherwise it won't work for `arm`

cjdcordeiro (changes requested):
Thanks for yet another slice @timo-reymann.

There's a fix which is required, but it should be good after that ;)

cjdcordeiro (approved):
Thank you ;)

PR #58, opened 2023-09-01
feat: migrate slices from 23.04 to 23.10

cjdcordeiro:
This PR brings onto 23.10 the core slices from 23.04.

The major differences are:
   - `libicu71` is renamed to `libicu72` in Mantic

PR #59, opened 2023-09-05
feat: Ubuntu 23.10 slices for NGINX

cjdcordeiro:
Needs https://github.com/canonical/chisel-releases/pull/58

This PR adds the necessary slice definitions for NGINX and its dependencies. 

A couple of noteworthy mentions:
 - `iproute2` is also being sliced in this PR, but is ultimately not included in the list of dependencies for Nginx. The rationale is that, although listed as a pkg dependency, `iproute2` seems to only be used by the Nginx maintainer scripts and thus not essential for a correct nginx behaviour. May `iproute2` be actually needed for nginx at runtime, we can simply uncomment that dependency later on.
 - a similar rationally is being applied for `debconf` which seems to be used by some of these packages, but only for the maintainer scripts.

## Testing

The following will give you a functional Nginx filesystem. `nginx-common_index` is only needed for demonstrating the default index.html example, but can be dropped in production installations.

```bash
chisel cut --release $PWD \
                --root nginx-rootfs \
                base-files_var \
                base-passwd_data \
                nginx_bins \
                nginx-common_index
```

### DIFF
--- a/slices/iproute2.yaml
+++ b/slices/iproute2.yaml
@@ -1,0 +1,75 @@
+package: iproute2
+
+slices:
+  # iproute2 is a utility package which typically could be split into
+  # utility-specific slices. The `bins` slice however brings in all the
+  # utilities for cases where iproute2 is used as a dependency for another
+  # package (eg. nginx), where we can't easily know which utilities are
+  # actually going to be needed.
+  bins:
+    # Leaving the debconf dependency out as we don't currently need pkg mgmt.
+    # Also, debconf is used by the pkg's maintainer scripts, which are handled
+    # differently in Chisel. 
+    essential:
+      - iproute2_config
+      - iproute2_libs
+      - libcap2-bin_bins
+    contents:
+      /bin/ip:
+      /sbin/ip:
+      /bin/ss:
+      /sbin/bridge:
+      /sbin/dcb:
+      /sbin/devlink:
+      /sbin/rtacct:
+      /sbin/rtmon:
+      # "tc" is an obvious candidate for getting its own "traffic-control"
+      # slice in the future, if needed.
+      /sbin/tc:
+      /sbin/tipc:
+      /sbin/vdpa:
+      /usr/bin/lnstat:
+      /usr/bin/ctstat:
+      /usr/bin/rtstat:
+      /usr/bin/nstat:
+      /usr/bin/rdma:
+      /usr/bin/routel:
+      /usr/sbin/arpd:
+      /usr/sbin/genl:
+
+  config:
+    # Corresponds to the pkg's conffiles
+    contents:
+      /etc/iproute2/bpf_pinning:
+      /etc/iproute2/ematch_map:
+      /etc/iproute2/group:
+      /etc/iproute2/nl_protos:
+      /etc/iproute2/rt_dsfield:
+      /etc/iproute2/rt_protos:
+      /etc/iproute2/rt_protos.d/:
+      /etc/iproute2/rt_realms:
+      /etc/iproute2/rt_scopes:
+      /etc/iproute2/rt_tables:
+      /etc/iproute2/rt_tables.d/:
+
+  libs:
+    essential:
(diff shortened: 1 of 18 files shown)


slices/nginx-common.yaml
```diff
@@ -0,0 +1,60 @@
+package: nginx-common
+
+slices:
+  # Leaving the debconf dependency out as we don't currently need pkg mgmt.
+  # Also, debcong is especially used for the maintainer scripts, which are
```
rebornplusplus, line 5: nit: typo: ~~debcong~~ debconf.

slices/iproute2.yaml
rebornplusplus, line 1: Out of curiosity, which maintainer script in nginx package utilizes iproute2?

slices/iproute2.yaml
```diff
@@ -0,0 +1,75 @@
...
+      /sbin/tipc:
+      /sbin/vdpa:
+      /usr/bin/lnstat:
+      /usr/bin/ctstat:
+      /usr/bin/rtstat:
+      /usr/bin/nstat:
+      /usr/bin/rdma:
+      /usr/bin/routel:
+      /usr/sbin/arpd:
+      /usr/sbin/genl:
```
rebornplusplus, line 38: nit: I understand the reason behind keeping file and it's symlinks together. But I think it would be good for the future to keep the file list sorted.

slices/iproute2.yaml
```diff
@@ -0,0 +1,75 @@
...
+    essential:
+      - libbpf1_libs
+      - libbsd0_libs
+      - libc6_libs
+      - libcap2_libs
+      - libdb5.3_libs
+      - libelf1_libs
+      - libmnl0_libs
+      - libselinux1_libs
+      - libgssapi-krb5-2_libs
```
rebornplusplus, line 65: Should this have been `libtirpc3_libs` instead?

rebornplusplus (changes requested):
Nice!

The PR looks good. Maybe we should change the title though, and include "iproute2" in the name too. Additionally, I think it would be nice to split the last commit into two -- 1) for nginx and it's dependencies, and 2) for iproute2 and dependencies.

slices/iproute2.yaml
cjdcordeiro: the `postinst` script calls the `ss` binary:

```bash
(...) 
  if [ -z "$(ss -nlt 'sport = 80' | grep -v ^State)" ]; then
    invoke-rc.d nginx start || exit $?
  else
    echo "Not attempting to start NGINX, port 80 is already in use."
    exit 0
  fi
(...)
```

slices/iproute2.yaml
cjdcordeiro: Y I thought about this. I haven't yet convinced myself of what's the best approach. Nonetheless, and given that many other SDFs have a similar structure, I recommend we take this into consideration when writing the contribution guidelines (which is something that might go alongside the new CI we're planning to have)

slices/iproute2.yaml
cjdcordeiro: y just noticed my typo. nice catch ;) thanks. fixed in https://github.com/canonical/chisel-releases/pull/59/commits/71e5c3d671999b5610be5defeff5c4dc7f891835

cjdcordeiro:
Thanks @rebornplusplus .

I've rebased and split the commit in 2, as requested. I've also addressed your comments ;)

rebornplusplus (approved):
LGTM. Thanks for the changes!

norrisjeremy:
Hi @cjdcordeiro @rebornplusplus,

Is there a reason this was never forward ported to later Ubuntu releases?

Thanks,
Jeremy

norrisjeremy:
> Hi @cjdcordeiro @rebornplusplus,
> 
> Is there a reason this was never forward ported to later Ubuntu releases?
> 
> Thanks, Jeremy

See #421.

norrisjeremy:
See also #422.

PR #61, opened 2023-09-13
Add slices for bash and coreutils

weiiwang01:
This pull request introduces slice definitions for bash and GNU coreutils, along with some associated libraries.


slices/bash.yaml
```diff
@@ -0,0 +1,10 @@
+package: bash
+
+slices:
+  bins:
+    essential:
+      - libc6_libs
+      - libtinfo6_libs
+    contents:
+      /bin/bash:
+      /bin/sh: { symlink: /bin/bash }
```
rebornplusplus, line 10: Hmm.. this is a tricky one. We should have a symlink to `/bin/bash`, but perhaps including this in `contents` might be a bad idea. Because, chisel does not support conflicting paths across slices/packages, yet. See https://github.com/canonical/chisel/issues/50.
So I am worried including `/bin/sh` might raise this issue. For example, if we were to install `dash` and then create a symlink in a same way, we will have conflicts.
So, perhaps, it's better to drop this from the `contents`, at least for now. We could leave a comment/note to remind ourselves that we need to do this manually after installation.

slices/coreutils.yaml
```diff
@@ -0,0 +1,120 @@
...
+      /usr/bin/fold:
+      /usr/bin/groups:
+      /usr/bin/head:
+      /usr/bin/hostid:
+      /usr/bin/id:
+      /usr/bin/install:
+      /usr/bin/join:
+      /usr/bin/link:
+      /usr/bin/logname:
+      /usr/bin/md5sum:
```
rebornplusplus, line 72: `/usr/bin/md5sum.textutils` seems to be missing.

slices/coreutils.yaml
```diff
@@ -0,0 +1,120 @@
...
+      /usr/bin/tty:
+      /usr/bin/unexpand:
+      /usr/bin/uniq:
+      /usr/bin/unlink:
+      /usr/bin/users:
+      /usr/bin/wc:
+      /usr/bin/who:
+      /usr/bin/whoami:
+      /usr/bin/yes:
+      /usr/libexec/coreutils/libstdbuf.so:
```
rebornplusplus, line 119: I think this should be in a `libs` slice.You can then add that slices as an `essential` in the `bins` slice.

slices/libtinfo6.yaml
```diff
@@ -0,0 +1,8 @@
+package: libtinfo6
+
+slices:
+  libs:
+    essential:
+      - libc6_libs
+    contents:
+      /lib/*-linux-*/libtinfo.so.6*:
```
rebornplusplus, line 8: Perhaps we should add `/usr/lib/*-linux-*/libtic.so.6*` as done in https://github.com/canonical/chisel-releases/blob/ubuntu-20.04/slices/libtinfo6.yaml.

rebornplusplus (changes requested):
Nice! It looks quite good.

I have a few comments below. Let me know what you think of them.

slices/bash.yaml
weiiwang01: Agreed, removed the `/bin/sh` from contents and added a comment.

slices/coreutils.yaml
weiiwang01: Added, thanks!

slices/coreutils.yaml
weiiwang01: I'm uncertain if we should place this library in a separate `libs` section. It appears to be an internal library for `coreutils` and isn't meant for external use.

slices/libtinfo6.yaml
weiiwang01: Yes, indeed, added, thanks!

slices/coreutils.yaml
```diff
@@ -0,0 +1,121 @@
...
+      /usr/bin/groups:
+      /usr/bin/head:
+      /usr/bin/hostid:
+      /usr/bin/id:
+      /usr/bin/install:
+      /usr/bin/join:
+      /usr/bin/link:
+      /usr/bin/logname:
+      /usr/bin/md5sum:
+      /usr/bin/md5sum.textutils: { symlink: /usr/bin/md5sum }
```
rebornplusplus, line 73: SInce this already comes as a symlink from the package, you don't need to manually create it.
```suggestion
      /usr/bin/md5sum.textutils:
```

slices/coreutils.yaml
rebornplusplus: > It appears to be an internal library for `coreutils` and isn't meant for external use.

True. But typically we group the libraries (including internal ones) into `libs` and binaries into `bins` for better categorization.

slices/coreutils.yaml
weiiwang01: Updated, thanks!

slices/coreutils.yaml
weiiwang01: Make sense, updated, thanks!

rebornplusplus (approved):
Thanks for the changes.

rebornplusplus:
@woky will you please take a look at this PR?

slices/bash.yaml
```diff
@@ -0,0 +1,10 @@
+package: bash
+
+slices:
+  bins:
```
cjdcordeiro, line 4: this is called `bins`, however it isn't bringing all the binaries from that package. If you don't want the other binaries, I'd recommend renaming the slice to something more specific (like the binary name maybe...`bash`?)

slices/bash.yaml
```diff
@@ -0,0 +1,10 @@
...
+    essential:
+      - libc6_libs
+      - libtinfo6_libs
+    contents:
+      # user should manually link /bin/sh to /bin/bash
```
cjdcordeiro, line 9: you should create this symlink. It deb creates via via a maintainer script, so it should be part of the slice. If not via a mutation script, then as an explicit symlink (using the `symlink` directive)

slices/bash.yaml
```diff
@@ -0,0 +1,10 @@
+package: bash
+
+slices:
+  bins:
+    essential:
```
cjdcordeiro, line 5: I know `debianutils` seems to only be required by the maintainer scripts, and you don't need those here, but are you sure `base-files` isn't needed?

slices/libattr1.yaml
```diff
@@ -0,0 +1,8 @@
+package: libattr1
+
+slices:
```
cjdcordeiro, line 3: isn't `/etc/xattr.conf` needed?

cjdcordeiro (changes requested):
Thanks for this! I have a few minor change requests and questions, but in general, it looks good ;)

slices/bash.yaml
cjdcordeiro: **edit:** apologies. Rafid raised a good point about a potential conflict with `/bin/sh` from the `dash` package. I see you were already doing what I asked in a previous commit and changed it according to Rafid's comment, which is right. It sucks, but I guess the safest option is to do the symlink manually for now. You can therefore ignore this comment.

slices/bash.yaml
weiiwang01: Alright, I've added additional binaries including `rbash` and `clear_console`. However, I think we need to include `bashbug`.

slices/libattr1.yaml
weiiwang01: Yes, updated, thanks!

slices/bash.yaml
```diff
@@ -0,0 +1,10 @@
+package: bash
+
+slices:
```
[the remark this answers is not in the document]
weiiwang01: Added, thanks!

slices/bash.yaml
weiiwang01: I think bash alone is okay to execute scripts without the `base-files`, but bash script might assume something inside `base-files`.

slices/bash.yaml
```diff
@@ -1,10 +1,19 @@
...
+
   bins:
     essential:
+      - bash_config
       - libc6_libs
       - libtinfo6_libs
     contents:
-      # user should manually link /bin/sh to /bin/bash
       /bin/bash:
+      /bin/rbash: { symlink: /bin/bash }
```
cjdcordeiro, line 18:
```suggestion
      /bin/rbash:
```

The deb already contains the symlink so there's no need to re-define it.

slices/bash.yaml
cjdcordeiro: Y I would say so

slices/bash.yaml
```diff
@@ -1,10 +1,19 @@
...
+      /etc/skel/.bashrc:
+      /etc/skel/.profile:
+
   bins:
     essential:
+      - bash_config
       - libc6_libs
       - libtinfo6_libs
     contents:
-      # user should manually link /bin/sh to /bin/bash
```
cjdcordeiro: please keep this comment. My previous remark was wrong and Rafid was on point.

slices/bash.yaml
cjdcordeiro: fair enough

cjdcordeiro (changes requested):

slices/bash.yaml
weiiwang01: Gotcha, added, thanks!

rebornplusplus (approved):

slices/bash.yaml
weiiwang01: Added, thanks!

cjdcordeiro (approved):
LGTM thanks

slices/bash.yaml
```diff
@@ -0,0 +1,21 @@
...
+
+  bins:
+    essential:
+      - bash_config
+      - libc6_libs
+      - libtinfo6_libs
+    contents:
+      # user should manually link /bin/sh to /bin/bash
+      /bin/bash:
+      /bin/bashbug:
```
cjdcordeiro, line 19:
```suggestion
      /usr/bin/bashbug:
```

Actually there's a typo here

cjdcordeiro (changes requested):

slices/bash.yaml
weiiwang01: I found the `bashbug` is in two places, I will just create a symlink for this, thanks!

```
$ ls -lah /bin/bashbug
-rwxr-xr-x 1 root root 6.7K Jan  6  2022 /bin/bashbug
$ ls -lah /usr/bin/bashbug
-rwxr-xr-x 1 root root 6.7K Jan  6  2022 /usr/bin/bashbug
```

slices/bash.yaml
```diff
@@ -16,6 +16,7 @@ slices:
     contents:
       # user should manually link /bin/sh to /bin/bash
       /bin/bash:
-      /bin/bashbug:
+      /bin/bashbug: { symlink: /usr/bin/bashbug }
```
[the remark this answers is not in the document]
weiiwang01: I suspect this came with the system image, I will remove `/bin/bashbug` to better align it with the deb package, thanks!

slices/bash.yaml
cjdcordeiro: The symlink was rightfully ditched so we can resolve this anyway ;)

cjdcordeiro (approved):
Thanks again ;)

PR #69, opened 2023-11-09
feat: migrate slices from 23.10 to 24.04

rebornplusplus:
This PR brings onto 24.04 the core slices from 23.10.

There are no changes in any of the files introduced in this PR.


cjdcordeiro (approved):
thanks for porting these

PR #81, opened 2023-12-13
feat(24.04): add ubuntu archive signing key 2018

rebornplusplus:
This PR adds the Ubuntu Archive Automatic Signing Key (2018) as a "public-key" in the chisel.yaml. The armored data was obtained by executing the following commands on a Ubuntu 22.04 machine:

    $ gpg --keyserver keyserver.ubuntu.com --receive-keys 871920D1991BC93C
    $ gpg --armor --export 871920D1991BC93C

BREAKING CHANGE: This PR introduces the new chisel yaml format "chisel-v2", in which ``public-keys`` (top-level) and ``archive.<name>.public-keys`` fields are introduced. This change is related to the addition of integrity checks in chisel. [1]

[1] https://github.com/canonical/chisel/pull/106

### DIFF
--- a/chisel.yaml
+++ b/chisel.yaml
@@ -5,9 +5,9 @@ archives:
         version: 24.04
         components: [main, universe]
         suites: [noble, noble-security, noble-updates]
-        public-keys: [ubuntu-archive-key-2018]
+        v1-public-keys: [ubuntu-archive-key-2018]
 
-public-keys:
+v1-public-keys:
     # Ubuntu Archive Automatic Signing Key (2018) <ftpmaster@ubuntu.com>
     # rsa4096/f6ecb3762474eda9d21b7022871920d1991bc93c 2018-09-17T15:01:46Z
     ubuntu-archive-key-2018:
--- a/chisel.yaml
+++ b/chisel.yaml
@@ -1,4 +1,4 @@
-format: chisel-v2
+format: chisel-v1
 
 archives:
     ubuntu:
--- a/chisel.yaml
+++ b/chisel.yaml
@@ -1,7 +1,44 @@
-format: chisel-v1
+format: chisel-v2
 
 archives:
     ubuntu:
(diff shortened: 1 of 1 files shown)


cjdcordeiro:
**Don't merge before the 9th** (requested by MSFT)

cjdcordeiro (changes requested):
See my comment in https://github.com/canonical/chisel/pull/106#pullrequestreview-1782027147

I don't think we should bump to `chisel-v2` just yet. Chisel v0.8.1 will be able to cope with these `public-keys` for a while longer

cjdcordeiro:
> See my comment in [canonical/chisel#106 (review)](https://github.com/canonical/chisel/pull/106#pullrequestreview-1782027147)
> 
> I don't think we should bump to `chisel-v2` just yet. Chisel v0.8.1 will be able to cope with these `public-keys` for a while longer

as a note, we've agreed to later bump the format to just `v1`, in order to avoid a big version gap between the format and the chisel tool release. Things change is equally disruptive (intentionally) and will helps remove the ambiguity between the format version and chisel tool versions

rebornplusplus:
Due to recent developments, the format has been reverted to the previous version `chisel-v1`.

cjdcordeiro (changes requested):
As per recent discussions, we'll need to change `public-keys` to `v1-public-keys` (this applies to the other significant ones.

Then we can merge it for testing with https://github.com/canonical/chisel/pull/106.

Before merging, we **need to ensure v0.8.1** works with it

letFunny (approved):
Because we do not have a CI please test it locally as well to make sure. I did:

| v0.8.1 | https://github.com/canonical/chisel/pull/106 |
|--------|--------|
| ✓ | ✓ |

cjdcordeiro (approved):
looking good now. thanks for the changes

PR #85, opened 2023-12-27
fix(ubuntu-23.10): permission configuration for log files

shyim:
otherwise I get

```
------
 > [internal] load build context:
------
ERROR: failed to solve: error from sender: failed to xattr rootfs/var/log/nginx/access.log: permission denied
```


rebornplusplus (approved):
Nice catch! Thanks.

cjdcordeiro (approved):
good catch indeed. thanks

PR #86, opened 2024-01-01
nginx: add /etc/passwd dependency

shyim:
In the nginx.conf is www-data referenced, otherwise it won't start

### DIFF
--- a/slices/nginx-common.yaml
+++ b/slices/nginx-common.yaml
@@ -6,6 +6,11 @@ slices:
   # handled differently in Chisel.
   # There's also no need (yet) for any init- and systemd-related files.
   config:
+    essential:
+      # adds required /run directory
+      - base-files_var
+      # adds /etc/passwd so nginx can find the www-data user
+      - base-passwd_data
     contents:
       /etc/logrotate.d/nginx:
       /etc/nginx/fastcgi.conf:


rebornplusplus:
Asking for more info: Will you please share how you are trying to install and run?

shyim:
```
chisel cut  --root nginx/ nginx_bins
```

```
FROM scratch as nginx
COPY nginx/ /
ENTRYPOINT ["/usr/sbin/nginx", "-g", "daemon off;"]
```

shyim:
We can move that also to the nginx-common slice? 🤔

rebornplusplus:
Right, so the error message is 

```
[emerg] 1#1: getpwnam("www-data") failed (2: No such file or directory) in /etc/nginx/nginx.conf:1
```

Since the user is defined in `/etc/nginx/nginx.conf` which is in `nginx-common`, I also think this should be added as a dependency of `nginx-common_config` slice. Typically we install the `base-passwd_data` slice whenever we create a chiselled container image. But since this does not work without it at all, it makes sense to be added as a dependency.

rebornplusplus (approved):
Looks good to me, thanks!

cjdcordeiro (changes requested):
Thanks for the proposal.

TBH, I wouldn't merge this as it sets a dangerous precedent where slice definitions aren't a true reflection of their deb. I say this for 2 reasons:
 1. `base-passwd` and `base-files` aren't explicitly marked as dependencies of any of the `nginx*` debs
 2. what `nginx-common` does is set those `/var` directories via the maintainer scripts, and not use the `base-files` package. That would be acceptable IMO.

`base-passwd` also installs much more than what nginx needs, which is yet another reason for pushing back on this.

Since said slices can be explicitly installed at `cut` time, or added post-mortem, the only change I'd welcome is adding the `/var`manipulations via mutation scripts in `nginx-common`

shyim:
You mean I create with the mutate of `nginx-common` the `/run` directory and the `/etc/passwd` will be just put as additional package into the `cut` command right?

cjdcordeiro:
> You mean I create with the mutate of `nginx-common` the `/run` directory and the `/etc/passwd` will be just put as additional package into the `cut` command right?

yep, I think that is the right path. The mutation scripts should follow the `postinst` maintainer script

rebornplusplus:
Hiya @shyim, hope you are well. Any updates on this?

rebornplusplus:
Hello @shyim @cjdcordeiro, we should act on this soon. 23.10 is going out of support soon, if not already. What we could do is close this PR and if needed, open another one with the same purpose for 24.04, any later releases or any past LTS releases. I will keep it open for a few days, for now.

PR #87, opened 2024-01-02
nginx-common: remove logrotate config

shyim:
You won't need log rotate in a docker container, when the best practise is to use `/dev/stdout`

### DIFF
--- a/slices/nginx-common.yaml
+++ b/slices/nginx-common.yaml
@@ -7,7 +7,6 @@ slices:
   # There's also no need (yet) for any init- and systemd-related files.
   config:
     contents:
-      /etc/logrotate.d/nginx:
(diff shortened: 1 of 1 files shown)


cjdcordeiro (commented):
I don't think we should do this. Definitely not for 23.10. For a couple of reasons:
 1. these slices are out, so by removing this path from the `config` slice we would be breaking the installation for anyone already using and relying on it
 2. nginx's default behavior is to actually write to the access and error logs, so this change would make it easier for a user to miss the logrotate configuration and end up with an installation that is prone to have an ever-increasing log file

We can re-assess this for other chisel-releases where nginx hasn't yet been sliced, but please remember that Chisel isn't a container-specific tool and thus we should not make assumptions about how it's going to be used.

PR #93, opened 2024-01-04
feat(24.04): add netbase slices

rebornplusplus:
This PR adds netbase slices for 24.04.

---

<details>
<summary>Reference: the `postinst` script for 24.04</summary>

```sh
#!/bin/sh -e

create_hosts_file() {
  if [ -e /etc/hosts ]; then return 0; fi

  cat > /etc/hosts <<-EOF
	127.0.0.1	localhost
	::1		localhost ip6-localhost ip6-loopback
	ff02::1		ip6-allnodes
	ff02::2		ip6-allrouters

EOF
}

create_networks_file() {
  if [ -e /etc/networks ]; then return 0; fi

  cat > /etc/networks <<-EOF
	default		0.0.0.0
	loopback	127.0.0.0
	link-local	169.254.0.0

EOF
}

# create bindv6only.conf on new installs and upgrades from << 4.38
create_bindv6only_conf() {
  if [ -e /etc/sysctl.d/bindv6only.conf ]; then
    return 0
  fi
  [ "$(uname -s)" = "Linux" ] || return 0
  [ -d /etc/sysctl.d/ ] || mkdir /etc/sysctl.d/
  if [ "$2" ] && dpkg --compare-versions "$2" ge "4.38"; then
    return 0
  fi

  cat >> /etc/sysctl.d/bindv6only.conf <<-EOF
# This sysctl sets the default value of the IPV6_V6ONLY socket option.
#
# When disabled, IPv6 sockets will also be able to send and receive IPv4
# traffic with addresses in the form ::ffff:192.0.2.1 and daemons listening
# on IPv6 sockets will also accept IPv4 connections.
#
# When IPV6_V6ONLY is enabled, daemons interested in both IPv4 and IPv6
# connections must open two listening sockets.
# This is the default behaviour of almost all modern operating systems.

net.ipv6.bindv6only = 1
EOF
}

case "$1" in
    configure)
    if [ -z "$2" ]; then
	create_hosts_file
	create_networks_file
    fi
    #create_bindv6only_conf "$@"
    ;;

    abort-upgrade|abort-remove|abort-deconfigure)
    ;;

    *)
    echo "postinst called with unknown argument '$1'" >&2
    exit 1
    ;;
esac
```
</details>

### DIFF
--- a/slices/netbase.yaml
+++ b/slices/netbase.yaml
@@ -1,0 +1,22 @@
+package: netbase
+
+slices:
+  config:
+    contents:
+      /etc/ethertypes:
+      /etc/protocols:
+      /etc/rpc:
+      /etc/services:
+      # The following files are created from maintainer script,
+      # if they do not exist.
+      /etc/hosts:
+        text: |
+          127.0.0.1	localhost
+          ::1		localhost ip6-localhost ip6-loopback
+          ff02::1		ip6-allnodes
+          ff02::2		ip6-allrouters
+      /etc/networks:
+        text: |
+          default		0.0.0.0
+          loopback	127.0.0.0
+          link-local	169.254.0.0


rebornplusplus:
> the prob is that you're assuming we always start from an empty fs, which may not be the case, and this slice would end up overwriting those files. can't we do it via `mutate`?

Yes, that was the assumption.

We cannot do it with `mutate` unfortunately. What the mutation script in the deb does is, it check if the files are present or not. If absent, only then it creates those files with some default content.

While we can do it with logic in the `mutate` block, the problem lies in listing the paths. Note that mutation script cannot operate on a file that is not listed with a `mutable` flag. Say we do this:

```yaml
slices:
  config:
    contents:
      /etc/hosts: { text: FOO, mutable: true }
      /etc/networks: { mutable: true }
    mutate: |
      # Logic to check if the files are not present.
      # If absent, more logic to create them with default content.
```

There lies two problems here.

1. The `/etc/hosts` file will be (re-)created with new data (`FOO`) and there will be no way for us to retrieve it's original data if the file was present in the rootfs before.
2. The listing of `/etc/networks` will result in an extraction failure because it's not present in deb. So we cannot list the files in this way either.

---

Given the inability at this moment, I suppose it would be best to remove those entries and leave a comment for the user to manually create those files with default content if not present. Let me know what you think.

linostar:
isn't there any equivalent for maintainer scripts in chisel? people can change whatever they want either manually or in their *craft.yaml, but I imagine they will hardly be aware that they need to change anything if they're informed just through a comment in a chisel slice definition file.

cjdcordeiro (changes requested):
@rebornplusplus indeed, we would need either a `content.copy/move` or `content.remove` sort of functions for `mutate`. Maybe you can add that to our backlog as an idea.

For now, unless those files are absolutely needed, I'd leave them out and keep your comment instead.

cjdcordeiro:
If these files are needed, then I'd also propose another option: adding additional slices (e.g. `default-hosts` and `default-networks`). This way we don't make assumption when installing the rest of the package, while giving people a way to opt-in for those config files.

rebornplusplus:
> If these files are needed, then I'd also propose another option: adding additional slices (e.g. `default-hosts` and `default-networks`). This way we don't make assumption when installing the rest of the package, while giving people a way to opt-in for those config files.

I like this option. I added another slice `defaults` and kept the files there with a comment. Let me know what you think.

linostar (approved):
Lgtm

cjdcordeiro (changes requested):
ty for the change. tbh, I'd keep them separate. it is more explicit that way and less prone to assumptions. For example, atm, I would expect `/etc/sysctl.d/bindv6only.conf` to also be in defaults and it isn't. 

Given that this is a very particular case, driven by a limitation on the mutation scripts, I explicitly have slices for each default file.

rebornplusplus:
Fair enough, updated. Please take a look.

cjdcordeiro (approved):
Noice! Thank you

PR #95, opened 2024-01-04
fix(24.04): update file path in zlib1g_libs

rebornplusplus:
The libz.so files have moved location in 24.04. This commit updates the zlib1g_libs slice with proper path.


cjdcordeiro (approved):
Thanks for the fix!

PR #102, opened 2024-01-25
fix(23.10): add missing path in openssl_config

rebornplusplus:
`/usr/lib/ssl/certs` is a symlink to `/etc/ssl/certs`. But the latter was not included in the slice, resulting in a broken symlink in the rootfs if the ca-certificates slices were not installed.

This PR adds the missing `/etc/ssl/certs` path.

### DIFF
--- a/slices/openssl.yaml
+++ b/slices/openssl.yaml
@@ -14,8 +14,9 @@ slices:
 
   config:
     contents:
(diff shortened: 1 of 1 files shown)


cjdcordeiro (changes requested):
there's a typo - missing the `/` in the path. These don't work.

PR #103, opened 2024-01-29
feat(20.04): add slices for bash and coreutils

rebornplusplus:
This PR adds the slices for `bash` and `coreutils` in focal.


cjdcordeiro (approved):
These are functional and seem to be a backport from 22.04, so it LGTM. TY

PR #106, opened 2024-01-31
feat(22.04): add slices for (g)awk and dependencies

lwnexgen:
Adds slices for:
1. `gawk`
2. `libmpfr6`
3. `libsigsegv2`


cjdcordeiro:
Hi @lwnexgen ! Thank you for your contribution. I've assigned a few reviewers who will be looking at your slices soon.

In the meantime, we're applying a policy to this project where we are requesting all contributions to have the corresponding slice definitions also submitted to newer and still supported ubuntu releases. In this case, I'd kindly request you to also open the corresponding PRs for 23.10 and 24.04 :)

slices/gawk.yaml
```diff
@@ -0,0 +1,19 @@
+package: gawk
+
+slices:
+  libs:
+    essential:
+      - libmpfr6_libs
+      - libreadline8_libs
+      - libsigsegv2_libs
+    contents:
+      /usr/lib/x86_64-linux-gnu/gawk/*.so:
```
rebornplusplus, line 10:
```suggestion
      /usr/lib/*-linux-*/gawk/*.so:
```

Otherwise, it won't work for other arch.

slices/gawk.yaml
```diff
@@ -0,0 +1,19 @@
...
+  bins:
+    essential:
+      - libc6_libs
+      - gawk_libs
+    contents:
+      # user should manually link /usr/bin/gawk to /usr/bin/awk
+      /usr/bin/gawk:
+      /usr/lib/x86_64-linux-gnu/awk/grcat:
+      /usr/lib/x86_64-linux-gnu/awk/pwcat:
```
rebornplusplus, line 19:
```suggestion
      /usr/lib/*-linux-*/awk/grcat:
      /usr/lib/*-linux-*/awk/pwcat:
```

same as before.

slices/gawk.yaml
```diff
@@ -0,0 +1,19 @@
+package: gawk
+
+slices:
+  libs:
+    essential:
+      - libmpfr6_libs
+      - libreadline8_libs
+      - libsigsegv2_libs
```
rebornplusplus, line 8: There should be `libc6_libs` and `libgmp10_libs` as well since some libs depend on them.

slices/gawk.yaml
```diff
@@ -0,0 +1,19 @@
...
+    contents:
+      /usr/lib/x86_64-linux-gnu/gawk/*.so:
+  bins:
```
rebornplusplus, line 11:
```suggestion

  bins:
```
(nit)

slices/gawk.yaml
```diff
@@ -0,0 +1,19 @@
...
+    essential:
+      - libc6_libs
```
rebornplusplus, line 13: `libc6_libs` should be inside `gawk_libs`, so we can discard it here.

slices/gawk.yaml
```diff
@@ -0,0 +1,19 @@
+package: gawk
+
+slices:
```
rebornplusplus, line 3: How about adding a `config` slice which contains the following conffiles:

```
/etc/profile.d/gawk.csh
/etc/profile.d/gawk.sh
```

slices/gawk.yaml
```diff
@@ -0,0 +1,19 @@
+package: gawk
+
```
rebornplusplus, line 2: What do you think about the `/usr/include/gawkapi.h` file? Should it be in any slices?

rebornplusplus (changes requested):
Nice slicing! I left a few comments including a few nitpicks as well.

slices/gawk.yaml
lwnexgen: Are header files in scope for chiseled containers? I didn't see anything else that was including them (at least in `22.04`) and couldn't think of a good use case for someone doing builds inside a chiseled container.

I could add a `headers` (or maybe `dev`?) slice if there's something I'm missing though.

slices/gawk.yaml
lwnexgen: Oh I thought only direct dependencies were required - does this mean that `libmpfr6_libs` won't pull in `libc6_libs` and `libgmp10_libs`, even though they are specified as essential dependencies?

```
package: libmpfr6

slices:
  libs:
    essential:
      - libc6_libs
      - libgmp10_libs
    contents:
      /usr/lib/*-linux-*/libmpfr.so.6*:
```

slices/gawk.yaml
lwnexgen: Added a `headers` slice, let me know what you think.

lwnexgen:
> In this case, I'd kindly request you to also open the corresponding PRs for 23.10 and 24.04 :)

Will do once I get the issues sorted out with these PR(s) so I know they're good.

slices/gawk.yaml
rebornplusplus: Yup, we only list the direct dependencies even when some slices may come with some of the dependencies inside them. In this case, as per https://packages.ubuntu.com/jammy/gawk, libc6 and libgmp10 are dependencies as well, so it's best to list them here too for better maintainability.

slices/gawk.yaml
```diff
@@ -0,0 +1,27 @@
...
+slices:
+  libs:
+    essential:
+      - libmpfr6_libs
+      - libreadline8_libs
+      - libsigsegv2_libs
+      - libc6_libs
+      - libgmp10_libs
```
rebornplusplus, line 10: nit: would you please sort these entries in lexicographical order?

slices/gawk.yaml
rebornplusplus: Good point. I was just curious if we should start including them, but for now, there seems not to be much application for that. 
I will let you decide if you want to remove the `headers` slice or not. It's good for me either way.

slices/gawk.yaml
```diff
@@ -0,0 +1,27 @@
+package: gawk
+
+slices:
```
rebornplusplus, line 3: nit: It would also be nice to have the slice definitions in a lexicographic order based on slice names. So `bins`, `headers`, then `libs`.

rebornplusplus (commented):
Left a bunch of nitpicks. But looks nice overall.

rebornplusplus (approved):
Looks good to me. Slices are being installed properly on multiple arches as well.

cjdcordeiro (approved):
It all looks good! Thanks for the changes. I agree with @rebornplusplus 's comments. 

Approving it but holding the merge until the remaining PRs are created ;)

thanks again

rebornplusplus (approved):

lwnexgen:
Filed #110 and #111 for 23.10 and 24.04 - thanks for the help getting this through, everyone.

PR #105, opened 2024-01-30
feat(22.04): Add findutils slice

lwnexgen:
Adds `findutils` slice def to 22.04, to support running `find` and `xargs` in chiseled containers.

PR #109, opened 2024-01-31
feat(24.04): Add findutils slice

lwnexgen:
Adds findutils slice def to 24.04, to support running `find` and `xargs` in chiseled containers.

(see #105)


linostar (approved):
LGTM.

cjdcordeiro (approved):
All good here as well. thanks

PR #113, opened 2024-02-01
feat(22.04): Add grep slice

lwnexgen:
Add `grep` slice for Ubuntu 22.04.

### DIFF
--- a/slices/grep.yaml
+++ b/slices/grep.yaml
@@ -7,6 +7,7 @@ slices:
       - libpcre3_libs
     contents:
       /bin/grep:
+
   deprecated:
     # These are shell scripts requiring a symlink from /usr/bin/bash to /bin/sh
     # See: https://manpages.ubuntu.com/manpages/jammy/en/man1/grep.1.html
--- a/slices/grep.yaml
+++ b/slices/grep.yaml
@@ -6,7 +6,13 @@ slices:
       - libc6_libs
       - libpcre3_libs
     contents:
+      /bin/grep:
+  deprecated:
+    # These are shell scripts requiring a symlink from /usr/bin/bash to /bin/sh
+    # See: https://manpages.ubuntu.com/manpages/jammy/en/man1/grep.1.html
+    essential:
+      - bash_bins
+    contents:
       /bin/egrep:
       /bin/fgrep:
-      /bin/grep:
       /usr/bin/rgrep:
--- a/slices/grep.yaml
+++ b/slices/grep.yaml
@@ -1,0 +1,12 @@
+package: grep
+
+slices:
+  bins:
+    essential:
+      - libc6_libs
+      - libpcre3_libs
+    contents:
+      /bin/egrep:
+      /bin/grep:
+      /bin/fgrep:
+      /usr/bin/rgrep:
--- a/slices/grep.yaml
+++ b/slices/grep.yaml
@@ -7,6 +7,6 @@ slices:
       - libpcre3_libs
     contents:
       /bin/egrep:
-      /bin/grep:
       /bin/fgrep:
+      /bin/grep:
       /usr/bin/rgrep:


slices/grep.yaml
```diff
@@ -0,0 +1,12 @@
+package: grep
+
+slices:
+  bins:
+    essential:
+      - libc6_libs
+      - libpcre3_libs
+    contents:
+      /bin/egrep:
```
lwnexgen, line 9: `egrep`, `fgrep`, and `rgrep` are all shell scripts that depend on `#!/bin/sh` being installed - this needs to be a manually created symlink as per this [comment](https://github.com/canonical/chisel-releases/blob/47a2d67283e10f1a7494d5c8c023e073e0d8147a/slices/bash.yaml#L17)

Should these be excluded since they won't work as-is?

rebornplusplus (approved):
Looks good to me, thank you.

rebornplusplus:
Per https://github.com/canonical/chisel-releases/pull/106#issuecomment-1918560764, will you please add the slice for 23.10 and 24.04 as well? Thanks.

SamirPS (approved):

slices/grep.yaml
lwnexgen, line 9: Before I cut the other release PRs:

`egrep`, `fgrep`, and `rgrep` are all shell scripts that depend on `#!/bin/sh` being installed - this needs to be a manually created symlink as per this [comment](https://github.com/canonical/chisel-releases/blob/47a2d67283e10f1a7494d5c8c023e073e0d8147a/slices/bash.yaml#L17)

Should these be excluded since they won't work as-is?

cjdcordeiro: What would the symlink point to? Do you have a working example (Dockerfile perhaps)

lwnexgen: This is what I did in my container to get it working:

```
FROM mybaseimagewithchisel:latest as chiselbase

# Build dependencies for NGINX
RUN apt-get update && \
    apt-get install -y \
    build-essential \
    libgd-dev \
    libpcre3-dev \
    libssl-dev \
    zlib1g-dev \

WORKDIR /app/nginx-build

# Temporary repo for chisel-releases (slice definitions) until they are merged into the main chisel repo
RUN git clone https://github.com/lwnexgen/chisel-releases /app/nginx-build/chisel-releases && cd /app/nginx-build/chisel-releases && git checkout "gawk+findutils"

# Runtime dependencies for NGINX - these correspond to slices in the chisel-releases repo
# specified by <package_name>-<slice_name> (e.g. gawk_bins means "install the binary files from the gawk package")
RUN mkdir -p /rootfs/var/lib/dpkg/ \
    && chisel-wrapper --generate-dpkg-status /rootfs/var/lib/dpkg/status -- \
    --release /app/nginx-build/chisel-releases --root /rootfs \
    base-files_base \
    base-files_release-info \
    base-passwd_data \
    bash_bins \
    ca-certificates_data \
    coreutils_bins \
    findutils_bins \
    gawk_bins \
    gawk_libs \
    grep_bins \
    libc-bin_ldd \
    libc-bin_nsswitch \
    libc6_libs \
    libgcc-s1_libs \
    libpcre3_libs \
    libssl3_libs \
    libstdc++6_libs \
    zlib1g_libs \
    && echo "done"

# Setup the nginx user in the chroot
RUN groupadd -R /rootfs --system --gid 101 nginx && \
    useradd -R /rootfs --system --gid nginx --no-create-home --home /nonexistent --comment "nginx user" --shell /bin/false --uid 101 nginx

COPY *nginx* ./

# Fetch and download the latest NGINX source and extract it to /tmp for the build
# Install to /rootfs by passing that to our build script which will use the nginx makefile's DESTDIR
RUN ./get-latest-nginx.sh
RUN ./build-nginx.sh /tmp/nginx-build /tmp/nginx-debian /rootfs

# Nginx needs an empty conf.d directory
RUN mkdir -p /rootfs/etc/nginx/conf.d

# Copy entrypoint stuff from the upstream NGINX git repo
RUN mkdir -p /rootfs/docker-entrypoint.d
RUN cp /tmp/nginx-docker-git/mainline/debian/docker-entrypoint.sh /rootfs/
RUN cp /tmp/nginx-docker-git/mainline/debian/[0-9]*.sh /rootfs/docker-entrypoint.d/

# add aliased symlinks for bash -> sh and gawk -> awk
# (this is necessary because chisel won't create these symlinks for us)
RUN ln -sf bash /rootfs/bin/sh
RUN ln -sf gawk /rootfs/usr/bin/awk

# Runtime image is based on scratch
FROM scratch

# Setup initial minimal chroot by copying the chiseled rootfs from chiselbase
COPY --from=chiselbase /rootfs /

ENTRYPOINT ["/docker-entrypoint.sh"]
CMD ["nginx", "-g", "daemon off;"]
```

lwnexgen: This is the specific bit where I set up the symlinks.
```
# add aliased symlinks for bash -> sh and gawk -> awk
# (this is necessary because chisel won't create these symlinks for us)
RUN ln -sf bash /rootfs/bin/sh
RUN ln -sf gawk /rootfs/usr/bin/awk
```

I borrowed the `chisel-wrapper` script from Microsoft's .NET images.

cjdcordeiro: Tyvm. Right, true.

I think the only thing we can do is document it. Like the `bash` slice. 

I'll propose one thing do, which might improve the user experience:
 - can you please create another slice called `deprecated` and put the `{r,f,e}grep` binaries inside, while keeping the `bins` slice for `grep` only.
 - keep the comment on the `deprecated` slice, pointing to https://manpages.ubuntu.com/manpages/focal/en/man1/grep.1.html, and mentioning the need for the symlink if people really need to use those binaries.

cjdcordeiro (changes requested):

slices/grep.yaml
lwnexgen: Cool, updated. By the way, I was able to build a 13mb(!) base image for nginx using Chisel ; pretty remarkable.

lwnexgen (commented):
Resolved request by adding comment pointing to grep man page.

lwnexgen:
Filed #117 and #118 for 23.10 and 24.04

slices/grep.yaml
```diff
@@ -0,0 +1,18 @@
+package: grep
+
+slices:
+  bins:
+    essential:
+      - libc6_libs
+      - libpcre3_libs
+    contents:
+      /bin/grep:
+  deprecated:
```
```suggestion

  deprecated:
```

cjdcordeiro (approved):
Nice thank you! LGTM (just added an empty line between slices, to make it more readable)

merging

PR #117, opened 2024-02-02
feat(23.10): add grep slice

lwnexgen:
Add `grep` slice for Ubuntu 23.10.

### DIFF
--- a/slices/grep.yaml
+++ b/slices/grep.yaml
@@ -4,16 +4,16 @@ slices:
   bins:
     essential:
       - libc6_libs
-      - libpcre3_libs
+      - libpcre2-8-0_libs
     contents:
       /bin/grep:
 
   deprecated:
-    # These are shell scripts requiring a symlink from /usr/bin/bash to /bin/sh
+    # These are shell scripts requiring a symlink from /usr/bin/dash to /bin/sh
     # See: https://manpages.ubuntu.com/manpages/jammy/en/man1/grep.1.html
     essential:
-      - bash_bins
-      - libpcre3_libs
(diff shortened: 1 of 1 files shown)


slices/grep.yaml
```diff
@@ -0,0 +1,18 @@
+package: grep
+
+slices:
+  bins:
+    essential:
+      - libc6_libs
+      - libpcre3_libs
+    contents:
+      /bin/grep:
+  deprecated:
```
```suggestion

  deprecated:
```

cjdcordeiro (changes requested):
thanks! also missing `libpcre3` here

cjdcordeiro (changes requested):
Sorry, I didn't mean `libpcre3` was missing from the `deprecated` slice, but rather it is missing from the `ubuntu-23.10` release. So if you try to install your slices, it won't work.

But as @rebornplusplus mentioned in https://github.com/canonical/chisel-releases/pull/118#discussion_r1480906803, in 23.10, `grep` depends on libpcre2-8-0 and not libpcre3 (and the former already exists in this chisel ubuntu release)

lwnexgen:
> But as @rebornplusplus mentioned in https://github.com/canonical/chisel-releases/pull/118#discussion_r1480906803, in 23.10, grep depends on libpcre2-8-0 and not libpcre3 (and the former already exists in this chisel ubuntu release)

Whoops, good catch - will resolve and actually test this time.

cjdcordeiro (approved):
thanks! lgtm

rebornplusplus (approved):
Looks good to me.

PR #118, opened 2024-02-02
feat(24.04): add grep slice

lwnexgen:
Add `grep` slice for Ubuntu 24.04.

### DIFF
--- a/slices/grep.yaml
+++ b/slices/grep.yaml
@@ -13,7 +13,7 @@ slices:
     # See: https://manpages.ubuntu.com/manpages/noble/en/man1/grep.1.html
     essential:
       - dash_bins
-      - libpcre2-8-0_libs
+      - grep_bins
     contents:
       /usr/bin/egrep:
       /usr/bin/fgrep:
--- a/slices/grep.yaml
+++ b/slices/grep.yaml
@@ -7,11 +7,13 @@ slices:
       - libpcre3_libs
     contents:
       /bin/grep:
+
   deprecated:
     # These are shell scripts requiring a symlink from /usr/bin/bash to /bin/sh
-    # See: https://manpages.ubuntu.com/manpages/noble/en/man1/grep.1.html
+    # See: https://manpages.ubuntu.com/manpages/jammy/en/man1/grep.1.html
     essential:
       - bash_bins
+      - libpcre3_libs
     contents:
       /bin/egrep:
       /bin/fgrep:
--- a/slices/grep.yaml
+++ b/slices/grep.yaml
@@ -4,7 +4,7 @@ slices:
   bins:
     essential:
       - libc6_libs
-      - libpcre3_libs
+      - libpcre2-8-0_libs
     contents:
       /usr/bin/grep:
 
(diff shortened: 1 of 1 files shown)


slices/grep.yaml
```diff
@@ -0,0 +1,18 @@
+package: grep
+
+slices:
+  bins:
+    essential:
+      - libc6_libs
+      - libpcre3_libs
+    contents:
+      /bin/grep:
+  deprecated:
```
```suggestion

  deprecated:
```

slices/grep.yaml
cjdcordeiro, line 1: won't work cause libpcre3 is missing - an SDF for it must be included

cjdcordeiro (changes requested):
thanks! we'll need libpcre3 too though

slices/grep.yaml
```diff
@@ -0,0 +1,20 @@
+package: grep
+
+slices:
+  bins:
+    essential:
+      - libc6_libs
+      - libpcre3_libs
```
rebornplusplus, line 7: You will actually need [libpcre2-8-0](https://packages.ubuntu.com/noble/libpcre2-8-0) for noble: https://packages.ubuntu.com/noble/grep.

slices/grep.yaml
```diff
@@ -0,0 +1,20 @@
...
+    contents:
+      /bin/grep:
+
+  deprecated:
+    # These are shell scripts requiring a symlink from /usr/bin/bash to /bin/sh
+    # See: https://manpages.ubuntu.com/manpages/jammy/en/man1/grep.1.html
+    essential:
+      - bash_bins
```
rebornplusplus, line 15: I am wondering if we need `dash` or `bash`. If you only need classic `/bin/sh`, `dash` seems to be the better choice.

slices/grep.yaml
```diff
@@ -0,0 +1,20 @@
...
+      - libpcre3_libs
+    contents:
+      /bin/egrep:
+      /bin/fgrep:
+      /usr/bin/rgrep:
```
rebornplusplus, line 20: Some of the paths are wrong. See here: https://packages.ubuntu.com/noble/amd64/grep/filelist.

```
/usr/bin/egrep
/usr/bin/fgrep
/usr/bin/rgrep
```

slices/grep.yaml
```diff
@@ -0,0 +1,20 @@
...
+      - libpcre3_libs
+    contents:
+      /bin/grep:
+
+  deprecated:
+    # These are shell scripts requiring a symlink from /usr/bin/bash to /bin/sh
+    # See: https://manpages.ubuntu.com/manpages/jammy/en/man1/grep.1.html
+    essential:
+      - bash_bins
+      - libpcre3_libs
```
rebornplusplus, line 16: I think we will need `grep_bins` since the `egrep` (and others) call `grep` from inside. 

```sh
~# cat grep/usr/bin/egrep 
#!/bin/sh
cmd=${0##*/}
exec grep -E "$@"
```

We can remove the `libpcre2-8-0` dependency from this slice since `grep_bins` will have it.

rebornplusplus (changes requested):

slices/grep.yaml
cjdcordeiro: true!

slices/grep.yaml
lwnexgen: Oh ho, good catch - I only have a Jammy machine handy.

rebornplusplus (approved):
Looks good to me, thank you!

cjdcordeiro (approved):

PR #119, opened 2024-02-05
chore(main): add reusable workflow to test installing slices

rebornplusplus:
This PR adds a reusable workflow file `.github/workflows/install-slices.yaml` which can be used to test installing slices in changed or all slice definition files. It also adds a helper script `github/scripts/install-slices` used by the mentioned workflow file.

---

_Example: https://github.com/rebornplusplus/chisel-releases/pull/2_


cjdcordeiro (changes requested):
see https://github.com/canonical/chisel-releases/pull/120#pullrequestreview-1865177778

cjdcordeiro (changes requested):
hum there's quite a lot of repetition here. You don't need the two separate workflows. 

One is enough, where:

```yaml
on:
  workflow_call:
    inputs:
      install-all:
        description: "If true, install all slices in slices/"
        type: boolean
        default: false
        required: false
  push:
    branches:
      - "main"
    paths:
      - ".github/**"
  pull-request:
    branches:
      - "main"
    paths:
      - ".github/**"
  schedule:
    # Run at 00:00 every day.
    # Ref: https://man7.org/linux/man-pages/man5/crontab.5.html
    - cron: "0 0 * * *"
```

and then you can have a dedicated job just for creating the "installation" matrix

```yaml
    strategy:
      fail-fast: false
      matrix: ${{ fromJSON(needs.prepare-install.outputs.install-matrix) }}
```

rebornplusplus:
That is an interesting idea. I tried to do that. But the problem is that matrix will need to be different based on whether the workflow has been called via `workflow_call` or not. Maybe it's just me, but I didn't find any way to distinguish that, from within the reusable workflow. Let me know if you have some suggestions.

I added another input to remove duplication. Please let me know what you think!

> hum there's quite a lot of repetition here. You don't need the two separate workflows.
> 
> One is enough, where:
> 
> ```yaml
> on:
>   workflow_call:
>     inputs:
>       install-all:
>         description: "If true, install all slices in slices/"
>         type: boolean
>         default: false
>         required: false
>   push:
>     branches:
>       - "main"
>     paths:
>       - ".github/**"
>   pull-request:
>     branches:
>       - "main"
>     paths:
>       - ".github/**"
>   schedule:
>     # Run at 00:00 every day.
>     # Ref: https://man7.org/linux/man-pages/man5/crontab.5.html
>     - cron: "0 0 * * *"
> ```
> 
> and then you can have a dedicated job just for creating the "installation" matrix
> 
> ```yaml
>     strategy:
>       fail-fast: false
>       matrix: ${{ fromJSON(needs.prepare-install.outputs.install-matrix) }}
> ```

cjdcordeiro (commented):
Why does the matrix need to be different?

in both cases the matrix is:

```yaml
    strategy:
      fail-fast: false
      matrix:
        branch: []
        arch: []
```

So you'll just need a job, before the matrix one, to create that `map` and pass it, as a JSON output, to the matrix job

rebornplusplus:
> Why does the matrix need to be different?
> 
> in both cases the matrix is:
> 
> ```yaml
>     strategy:
>       fail-fast: false
>       matrix:
>         branch: []
>         arch: []
> ```

If I have a PR (or push) targeted at `ubuntu-22.04` branch, there needs not be any `branch` in the matrix there. It should simply run against that PR ref (or branch).

cjdcordeiro:
But that's precisely where you can apply DRY. When there's a PR or a push against a specific branch, your matrix should look like:

```yaml
    strategy:
      fail-fast: false
      matrix:
        branch: ["ubuntu-22.04"]
        arch: [amd64, ...]
```

And then your `checkout` action simply needs to use that `ref`

rebornplusplus:
On `push`, that matrix will work, yeah. But I don't think it will for PRs though. Because, if you "checkout" the `ubuntu-22.04` branch for a PR targeted at that branch, you won't be able to determine the changed files in that PR. Because the files will not be checked out. The new files will be part of the `refs/pull/<pr_number>/merge` [github.ref](https://docs.github.com/en/actions/learn-github-actions/contexts).

cjdcordeiro:
> On `push`, that matrix will work, yeah. But I don't think it will for PRs though. Because, if you "checkout" the `ubuntu-22.04` branch for a PR targeted at that branch, you won't be able to determine the changed files in that PR. Because the files will not be checked out. The new files will be part of the `refs/pull/<pr_number>/merge` [github.ref](https://docs.github.com/en/actions/learn-github-actions/contexts).

So you're saying that what you have now won't work?
https://github.com/canonical/chisel-releases/pull/119/files#diff-78bf4a33beef5f0e82b2661553d1e9fa2c2a17a5508f3be8eb5f1bf2e2fb1766R38-R41

rebornplusplus:
Added the changes as discussed in our sync: https://github.com/canonical/chisel-releases/pull/119/commits/868297402c25592dfcd07f352a0860734181b85f. Please take a look.

.github/workflows/install-slices.yaml
```diff
@@ -0,0 +1,121 @@
...
+        run: |
+          go install github.com/canonical/chisel/cmd/chisel@latest
+          go install github.com/mikefarah/yq/v4@latest
+          wget -q https://raw.githubusercontent.com/canonical/chisel-releases/main/.github/scripts/install-slices
+          chmod +x install-slices
+
+      - name: Install all slices
+        if: |
+          env.install-all == 'true' ||
+          steps.changed-paths.outputs.install-all == 'true'
```
cjdcordeiro, line 103: does this work if `changed-paths` was not executed?

.github/workflows/install-slices.yaml
```diff
@@ -0,0 +1,121 @@
...
+        run: |
+          set -e
+          # we need to enable globstar to use the ** patterns below
+          shopt -s globstar
+          ./install-slices --arch "${{ matrix.arch }}" --release ./ \
+            slices/**/*.yaml
+
+      - name: Install slices from changed files
+        if: |
+          steps.changed-paths.outputs.slices == 'true' &&
```
cjdcordeiro, line 113: same question as b4, i.e. if there is a `changed-paths` output, then it must mean `env.install-all` is `true`.

.github/workflows/install-slices.yaml
```diff
@@ -0,0 +1,121 @@
...
+          !(
+            env.install-all == 'true' ||
+            steps.changed-paths.outputs.install-all == 'true'
+          )
+        run: |
+          set -e
+          ./install-slices --arch "${{ matrix.arch }}" --release ./ \
```
cjdcordeiro, line 120: since the `run` is pretty much the same in these two last steps, you might want to try to merge them, and handle the conditions in shell, to avoid duplication

.github/workflows/install-slices.yaml
```diff
@@ -0,0 +1,121 @@
...
+          ]]; then
+            echo "matrix={\"ref\":${{ env.RELEASES }},\"arch\":${{ env.ARCHES }}}" >> $GITHUB_OUTPUT
+            echo "install_all=true" >> $GITHUB_OUTPUT
+          else
+            echo "matrix={\"arch\":${{ env.ARCHES }}}" >> $GITHUB_OUTPUT
+          fi
+
+  # The "install" job tests the slices by installing them.
+  # It installs **all** slices if:
+  #   - inputs.install-all is true
```
cjdcordeiro, line 49: not used anymore

cjdcordeiro (changes requested):
Looking better, thanks.

Do you have any test runs?

.github/workflows/install-slices.yaml
rebornplusplus: Hmm... if there is a `change-paths` output, then it means `env.install-all` **is not** `true`.

```yaml
      - name: Check changed paths
        id: changed-paths
        if: env.install-all != 'true'
        ...
```

.github/workflows/install-slices.yaml
rebornplusplus: If `env.install-all` is `true`, then the `changed-paths` step would not run resulting in an empty value for `steps.changed-paths.outputs.install-all` i.e. `steps.changed-paths.outputs.install-all=""`. Otherwise, if `env-install` is not `true`, the `changed-paths` step would be executed resulting in `true` or `false` values for the outputs of the job.

For this step, yeah, it does work because either `env.install-all` would be `true` or the `steps.changed-paths.outputs.install-all` would have some value `true` or `false`.

rebornplusplus: Note that this has been changed in https://github.com/canonical/chisel-releases/pull/119/commits/71a897eea58ac17332275feb6b252f1fb5d0931d. The comparison is now being done inside the `run` field.

.github/workflows/install-slices.yaml
rebornplusplus: Thanks, merged in https://github.com/canonical/chisel-releases/pull/119/commits/71a897eea58ac17332275feb6b252f1fb5d0931d.

.github/workflows/install-slices.yaml
rebornplusplus: Thanks! Removed in https://github.com/canonical/chisel-releases/pull/119/commits/71a897eea58ac17332275feb6b252f1fb5d0931d.

rebornplusplus:
> Do you have any test runs?

Yup.
1. Installing all slices: https://github.com/rebornplusplus/chisel-releases/actions/runs/7911557498/job/21595963180
   * This is triggered by `push` to the `main` branch.
2. Installing slices changed in a PR: https://github.com/rebornplusplus/chisel-releases/actions/runs/7868931941/job/21596112188?pr=2

.github/workflows/install-slices.yaml
cjdcordeiro: yes, that :smile:

cjdcordeiro:
> > Do you have any test runs?
> 
> Yup.
> 
> 1. Installing all slices: https://github.com/rebornplusplus/chisel-releases/actions/runs/7911557498/job/21595963180
>    
>    * This is triggered by `push` to the `main` branch.
> 2. Installing slices changed in a PR: https://github.com/rebornplusplus/chisel-releases/actions/runs/7868931941/job/21596112188?pr=2

great, thanks.

That made me realize something though -> the test should be fixed such that it ignores the arch if there's no deb for it upstream. Otherwise we'll get this everyday: https://github.com/rebornplusplus/chisel-releases/actions/runs/7911557498/job/21595963413, which is a false positive

rebornplusplus:
> That made me realize something though -> the test should be fixed such that it ignores the arch if there's no deb for it upstream. Otherwise we'll get this everyday: https://github.com/rebornplusplus/chisel-releases/actions/runs/7911557498/job/21595963413, which is a false positive

Are you practically saying to ignore these errors on a `chisel cut` run?

```sh
error: slice package "dotnet-host" missing from archive
```

(I wonder if there is any other way to check if the deb exists or not for a particular Ubuntu release, components and suites other than checking the `chisel cut` error.)

If so, then I can add an option to the script (something like `--ignore-pkg-missing`) which we should only use when we are trying to **install all** slices. If we use the option for PRs, then we might have contributors adding packages like `foo` which does not exist anywhere.

cjdcordeiro:
> check if the deb exists or not for a particular Ubuntu release

y this is what I'm saying. If it doesn't exist, it shouldn't throw an error. Like, .net only exists for arm64 and amd64, so, we'll be seeing errors everyday because of that, when in reality it's not an error. The script should only test the `chisel cut` for slices whose packages exist upstream, for the given release and arch

rebornplusplus:
I have added the option to ignore missing packages.
* When installing all slices, if a package has not been found in the archive, there will be no errors.
* In other cases, where we are only installing changed slices/paths, missing packages will not be ignored and will result in an error like before. This is useful in PRs

I have also re-written the "install-slices" script from bash to python for better maintainability. 

---

Example run for a PR: https://github.com/rebornplusplus/chisel-releases/actions/runs/7868931941/job/21646126428?pr=2
Example run for installing all slices: https://github.com/rebornplusplus/chisel-releases/actions/runs/7928229576/job/21646097589

cjdcordeiro (changes requested):
> I have added the option to ignore missing packages.
> 
> * When installing all slices, if a package has not been found in the archive, there will be no errors.
> * In other cases, where we are only installing changed slices/paths, missing packages will not be ignored and will result in an error like before. This is useful in PRs
> 
> I have also re-written the "install-slices" script from bash to python for better maintainability.
> 
> Example run for a PR: https://github.com/rebornplusplus/chisel-releases/actions/runs/7868931941/job/21646126428?pr=2 Example run for installing all slices: https://github.com/rebornplusplus/chisel-releases/actions/runs/7928229576/job/21646097589

I'm wondering...

> * When installing all slices, if a package has not been found in the archive, there will be no errors.

I tend to agree, **but** what if the package is removed from upstream? Like, completely...which afaik can happen for `dev` releases (noble atm). Then we would end up with a SDF in the release that doesn't work, at all.

I think the check should be: _the package must exist for at least 1 arch_. If it doesn't exist, at all, for a given release, then we'll be shooting ourselves in the foot everytime we open a new release, cause we won't be able to easily port the previous release onto the new one...not without potentially adding SDFs for packages that don't exist.


> * In other cases, where we are only installing changed slices/paths, missing packages will not be ignored and will result in an error like before. This is useful in PRs

Why? It's the same rationale as above...someone could try to add an SDF for a package that doesn't belong on that release.

---

I think there should be a check, before the multi-arch matrix, to validate that the slices can be tested - i.e. the package exists in the release.

rebornplusplus:
I have changed the script. The current behaviour of the script is the following:

* if `--ignore-missing` is set, it will first check if all the packages specified in the files passed are available for any arch in the archive. If it is, then it will filter out the arch-specific missing packages and only install whichever packages are available for that arch.
* otherwise, if unset, there will be no such checks. It will dumbly try to install every slice in the files.

We are only passing the `--ignore-missing` flag whenever we are installing all slices. Because of the new change, we will be assured that no package has been removed from the archive. And for scenarios where we are not passing `--ignore-missing`, we can be assured that whatever slice definition files we pass to the script, the script will try to install.
In short, we will know if we have any estranged packages in our repo/PRs.

---

<details>
<summary>Usage info of the script</summary>

```
usage: install-slices [-h] --arch ARCH --release RELEASE [--ignore-missing] [file ...]

Verify slice definition files by installing the slices

positional arguments:
  file               Chisel slice definition file(s)

options:
  -h, --help         show this help message and exit
  --arch ARCH        Package architecture
  --release RELEASE  Chisel-releases branch name or directory
  --ignore-missing   Ignore arch-specific package not found in archive errors. If set, also ensure that packages exist
                     for any arch.
```

</details>

---

Example runs:
- install-all: https://github.com/rebornplusplus/chisel-releases/actions/runs/7958827170
- only changes(PR): https://github.com/rebornplusplus/chisel-releases/actions/runs/7958884396?pr=2

.github/scripts/install-slices
```diff
@@ -0,0 +1,324 @@
...
+            continue
+        pkg = line.split("|")[0].strip()
+        available[pkg] = True
+    found = []
+    missing = []
+    for p in packages:
+        if available.get(p):
+            found.append(p)
+        else:
+            missing.append(p)
```
cjdcordeiro, line 205:
```suggestion
        found.append(line.split("|")[0].strip())
        
    missing = list(set(packages) - set(found))
```

.github/scripts/install-slices
```diff
@@ -0,0 +1,324 @@
...
+    """
+    _, missing = query_package_existence(
+        packages,
+        components=archive.components,
+        suites=archive.suites,
+    )
+    if len(missing) > 0:
+        logging.error(
+            "The following packages do not exist:\n"
+            + "\n".join(f"  - {p}" for p in missing)
```
cjdcordeiro, line 221: The error message by itself isn't too enlightening cause it's missing at least the `release`. This is probably a good place to use `archive.version`

.github/scripts/install-slices
```diff
@@ -0,0 +1,324 @@
...
+def ensure_package_existence(packages: list[str], archive: Archive) -> None:
+    """
+    Ensure that packages exist in the archive for any arch.
+    """
+    _, missing = query_package_existence(
+        packages,
+        components=archive.components,
+        suites=archive.suites,
+    )
+    if len(missing) > 0:
```
cjdcordeiro, line 218: this is a bit counter intuitive because this fn is called if `ignore_missing`, but then throws an error if pkgs are missing

cjdcordeiro: it's 2 different scopes of "missing", so maybe the fn & var & inputs names should be more explicit

.github/scripts/install-slices
```diff
@@ -0,0 +1,324 @@
...
+    Filter the Packages using the following criteria:
+        - if "ignore_missing" is true, ignore package if it does not exist
+          in the archive for [arch, release].
+    """
+    ignored = []
+    if ignore_missing:
+        package_names = [p.package for p in packages]
+        archive = parse_archive(release)
+        # ensure that packages exist in the archive.
+        ensure_package_existence(package_names, archive)
```
cjdcordeiro, line 242: this was the problem I was trying to sort out during our call -> this fn is arch-agnostic, but the script is being called for every matrix in the GH job. We're then re-doing the check multiple times, and possibly failing with the same error for every cell of the matrix

.github/scripts/install-slices
```diff
@@ -0,0 +1,324 @@
...
+        available = {}
+        for p in found:
+            available[p] = True
+        filtered = []
+        for p in packages:
+            if available.get(p.package):
+                filtered.append(p)
+            else:
+                ignored.append(p)
+        packages = filtered
```
cjdcordeiro, line 259:
```suggestion
        filtered = []
        for p in packages:
            if p.package in found:
                filtered.append(p)
            else:
                ignored.append(p)

        packages = filtered
```

.github/scripts/install-slices
```diff
@@ -0,0 +1,324 @@
...
+        logging.info("The following packages will be IGNORED:")
+        for pkg in ignored:
+            logging.info(f"  - {pkg.package}")
+    if len(packages) > 0:
+        logging.info("Slices of the following packages will be INSTALLED:")
+        for pkg in packages:
+            logging.info(f"  - {pkg.package}")
+    else:
+        logging.info("No slices will be installed.")
+        exit(0)
```
cjdcordeiro, line 316:
```suggestion
        return
```

cjdcordeiro (changes requested):
With a few additional changes you should also be able to significantly increase the pylint score.

.github/scripts/install-slices
```diff
@@ -0,0 +1,324 @@
...
+                version: 22.04
+                components: [main, universe]
+                suites: [jammy, jammy-security, jammy-updates]
+                ...
+        ...
+    """
+    logging.debug("Parsing ubuntu archive info...")
+    # (download and) parse chisel.yaml for ubuntu archive info
+    try:
+        if release.endswith("/"):
```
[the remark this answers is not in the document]
rebornplusplus: Changed to the following to mimic [Chisel's logic](https://github.com/canonical/chisel/blob/11a324ceca40c373ae5298ff6b93986340a945d7/cmd/chisel/cmd_cut.go#L59).
```py
        if "/" in release:
```

.github/scripts/install-slices
```diff
@@ -0,0 +1,324 @@
...
+            with open(f"{release}chisel.yaml", "r") as stream:
+                data = yaml.safe_load(stream)
+        else:
+            req_url = f"https://raw.githubusercontent.com/canonical/chisel-releases/{release}/chisel.yaml"
+            response = requests.get(req_url)
+            if response.status_code < 200 or response.status_code > 299:
+                logging.error(f"Cannot download chisel.yaml from remote")
+                exit(1)
```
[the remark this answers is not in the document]
rebornplusplus: Updated, thanks!

.github/scripts/install-slices
```diff
@@ -0,0 +1,324 @@
...
+    except Exception as e:
+        logging.error(e)
+        exit(1)
+    # load the yaml data into Archive
+    archive = Archive()
+    try:
+        archive_data = data["archives"]["ubuntu"]
+        archive.version = archive_data["version"]
+        archive.components = archive_data["components"]
+        archive.suites = archive_data["suites"]
```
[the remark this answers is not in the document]
rebornplusplus: Changed, but the above wouldn't work because `archive_data` contains also the `format` and `v1-public-keys` among other things.

.github/scripts/install-slices
```diff
@@ -0,0 +1,324 @@
...
+    except KeyError as e:
+        logging.error(f"{release}: key {e} not found")
```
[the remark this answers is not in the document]
rebornplusplus: Agreed.

.github/scripts/install-slices
```diff
@@ -0,0 +1,324 @@
...
+    cli_args.ignore_missing = args.ignore_missing
+    return cli_args
+
+
+class Archive:
+    """
+    Minimal data class replicating ubuntu archive in chisel.yaml
+    """
+
+    version: str
```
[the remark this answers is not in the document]
rebornplusplus: Used it in new commits.

.github/scripts/install-slices
rebornplusplus: Done, thanks!

.github/scripts/install-slices
rebornplusplus: Right, segregated them. Thank you.

.github/scripts/install-slices
rebornplusplus: We decided that although this means re-doing the check multiple times, and possibly failing with the same error for each arch in a particular release, we will go ahead with this for now and optimise/improve it in the future as needed.

rebornplusplus: Added a comment here: https://github.com/canonical/chisel-releases/pull/119/commits/bbb4c82a26d61da9f93b0eaa69577871eb7ba96c

rebornplusplus:
Example runs:
* install-all: https://github.com/rebornplusplus/chisel-releases/actions/runs/8000568097
* PR: https://github.com/rebornplusplus/chisel-releases/actions/runs/8000579476/job/21850229427?pr=2

.github/scripts/install-slices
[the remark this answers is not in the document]
cjdcordeiro: noted

.github/scripts/install_slices/pkglist
cjdcordeiro, line 1: nitpick: can you rename it to `requirements.sh` or something similar that conveys what the file is meant for?

cjdcordeiro (approved):
great! thanks! lgtm. let's see it in action

cjdcordeiro:
I'll merge it once we get a :+1: from @linostar

linostar (approved):
LGTM, great work!

PR #120, opened 2024-02-05
chore(22.04): add workflow to test slices

rebornplusplus:
This PR adds a workflow to test installing changed or all slices based on the files changed. It also runs a few jobs on schedule.

---

:red_circle: Depends on https://github.com/canonical/chisel-releases/pull/119.

---

Example: https://github.com/rebornplusplus/chisel-releases/pull/2

### DIFF
--- a/slices/base-files.yaml
+++ b/slices/base-files.yaml
@@ -36,7 +36,8 @@ slices:
       /var/cache/:
       /var/lib/:
       /var/log/:
-      /var/run/:
+      # The /var/run symlink is enforced by the package's maintainer scripts
+      /var/run/: {symlink: /run}
       /var/tmp/:
 
   home:
--- a/slices/curl.yaml
+++ b/slices/curl.yaml
@@ -1,0 +1,10 @@
+package: curl
+
+slices:
+  bins:
+    essential:
+      - libc6_libs
+      - libcurl4_libs    
+      - zlib1g_libs
+    contents:
+      /usr/bin/curl:
--- a/slices/libcurl4.yaml
+++ b/slices/libcurl4.yaml
@@ -1,0 +1,20 @@
+# Easy-to-use client-side URL transfer library (OpenSSL flavour)
+package: libcurl4
+
+slices:
+  libs:
+    essential:     
+      - libbrotli1_libs
+      - libc6_libs
+      - libgssapi-krb5-2_libs
+      - libidn2-0_libs
+      - libldap-2.5-0_libs
+      - libpsl5_libs
+      - libssh-4_libs
+      - libssl3_libs
+      - libnghttp2-14_libs
+      - libzstd1_libs
+      - librtmp1_libs
+      - zlib1g_libs
+    contents:
+      /usr/lib/*-linux-*/libcurl.so.4*:       
--- a/slices/libgcrypt20.yaml
+++ b/slices/libgcrypt20.yaml
@@ -1,0 +1,10 @@
+# LGPL Crypto library - runtime library
+package: libgcrypt20
+
+slices:
+  libs:
+    essential:
+      - libc6_libs
+      - libgpg-error0_libs         
+    contents:
+      /usr/lib/*-linux-*/libgcrypt.so.20*:
--- a/slices/libgpg-error0.yaml
+++ b/slices/libgpg-error0.yaml
@@ -1,0 +1,10 @@
+# Library that defines common error values, messages, and common
+# runtime functionality for all GnuPG components.
+package: libgpg-error0
+
+slices:
+  libs:
+    essential:
+      - libc6_libs
+    contents:
+      /lib/*-linux-*/libgpg-error.so.0*:
--- a/slices/libnghttp2-14.yaml
+++ b/slices/libnghttp2-14.yaml
@@ -1,0 +1,9 @@
+# Library implementing HTTP/2 protocol 
+package: libnghttp2-14
+
+slices:
+  libs:
+    essential:
+      - libc6_libs         
+    contents:
+      /usr/lib/*-linux-*/libnghttp2.so.14*:
--- a/slices/libpsl5.yaml
+++ b/slices/libpsl5.yaml
@@ -1,0 +1,13 @@
+# Libpsl5 allows checking domains against the Public Suffix List. It can be  
+# used to avoid privacy-leaking 'super-cookies', 'super domain' certificates, 
+# for domain highlighting purposes sorting domain lists by site and more.
+package: libpsl5
+
+slices:
+  libs:
+    essential:
+      - libc6_libs
+      - libidn2-0_libs
+      - libunistring2_libs
+    contents:
+      /usr/lib/*-linux-*/libpsl.so.5*:
--- a/slices/librtmp1.yaml
+++ b/slices/librtmp1.yaml
@@ -1,0 +1,14 @@
+# Toolkit for RTMP streams
+package: librtmp1
+
+slices:
+  libs:
+    essential:
+      - libc6_libs   
+      - libgmp10_libs
+      - libgnutls30_libs
+      - libhogweed6_libs
+      - libnettle8_libs
+      - zlib1g_libs 
+    contents:
+      /usr/lib/*-linux-*/librtmp.so.1*:
--- a/slices/libssh-4.yaml
+++ b/slices/libssh-4.yaml
@@ -1,0 +1,16 @@
+# Libssh2is a client-side C library implementing the SSH2 protocol.
+# It supports regular terminal, SCP and SFTP (v1-v5) sessions;
+# port forwarding, X11 forwarding; password, key-based and
+# keyboard-interactive authentication.
+package: libssh-4
+
+slices:
+  libs:
+    essential:
+      - libc6_libs
+      - libgssapi-krb5-2_libs
+      - libssl3_libs
+      - zlib1g_libs       
+    contents:
+      /usr/lib/*-linux-*/libssh.so.4*:
+       
--- a/slices/libzstd1.yaml
+++ b/slices/libzstd1.yaml
@@ -1,0 +1,9 @@
+# Fast lossless compression algorithm
+package: libzstd1
+
+slices:
+  libs:
+    essential:
+      - libc6_libs
+    contents:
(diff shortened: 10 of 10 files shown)


.github/workflows/tests.yaml
```diff
@@ -0,0 +1,75 @@
...
+
+  install-all:
+    name: Install all slices
+    needs: planner
+    if: ${{ needs.planner.outputs.install-all == 'true' }}
+    strategy:
+      fail-fast: false
+      matrix:
+        ARCH: [amd64, arm64, armhf, i386, ppc64el, riscv64, s390x]
+    uses: rebornplusplus/chisel-releases/.github/workflows/install-slices.yaml@main
```
linostar, line 57: this should change to canonica;/chisel-releases after PR #119 is merged.

.github/workflows/tests.yaml
```diff
@@ -0,0 +1,75 @@
...
+    name: Install slices in changed files
+    needs: planner
+    if: |
+      needs.planner.outputs.install-all != 'true' &&
+      needs.planner.outputs.install-changed == 'true'
+    strategy:
+      fail-fast: false
+      matrix:
+        ARCH: [amd64, arm64, armhf, i386, ppc64el, riscv64, s390x]
+    uses: rebornplusplus/chisel-releases/.github/workflows/install-slices.yaml@main
```
linostar, line 72: this should change to canonica;/chisel-releases after PR #119 is merged.

linostar (changes requested):

.github/workflows/tests.yaml
rebornplusplus: Indeed! I missed it, thanks. Updated in https://github.com/canonical/chisel-releases/pull/120/commits/5f0b6d38f8aed479521690043004ca561a4c8126.

.github/workflows/tests.yaml
rebornplusplus: Updated in https://github.com/canonical/chisel-releases/pull/120/commits/5f0b6d38f8aed479521690043004ca561a4c8126.

.github/workflows/tests.yaml
```diff
@@ -0,0 +1,75 @@
...
+      # changed-sdf is a space separated list of slice definition file
+      # paths who have been added or modified (not deleted).
+      changed-sdf: ${{ steps.changed-paths.outputs.slices_files }}
+    steps:
+      - uses: actions/checkout@v4
+        if: ${{ github.event_name != 'schedule' }}
+        with:
+          fetch-depth: 0
+
+      - name: Check changed paths
```
cjdcordeiro, line 33: the rationale is good, but it is slightly more complicated, but can use less lines of code. Eg.:

```yaml
        id: changed-paths
        if: ${{ github.event_name != 'schedule' }}
        uses: dorny/paths-filter@v3
        with:
          # ref: https://github.com/marketplace/actions/paths-changes-filter
          filters: |
            install-all:
              - 'chisel.yaml'
              - '.github/**'
              - deleted: 'slices/**'
            slices:
              - added|modified: 'slices/**/*.yaml'
          # Space delimited list usable as command-line argument list in
          # Linux shell. If needed, it uses single or double quotes to
          # wrap filename with unsafe characters.
          list-files: shell
```

and you can drop the `install-changed` output

cjdcordeiro (changes requested):

.github/workflows/tests.yaml
```diff
@@ -0,0 +1,75 @@
+name: Tests
+
+on:
+  push:
+    branches:
+      - "ubuntu-*"
+  pull_request:
+    branches:
+      - "ubuntu-*"
```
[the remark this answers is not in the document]
rebornplusplus: Yup, it works. See comments at https://github.com/rebornplusplus/chisel-releases/pull/5.

.github/workflows/tests.yaml
```diff
@@ -0,0 +1,75 @@
...
+  schedule:
+    # run at 00:00 on the first of every month
+    - cron: "0 0 1 * *"
```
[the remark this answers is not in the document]
rebornplusplus: fair enough, will do.

.github/workflows/tests.yaml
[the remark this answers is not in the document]
cjdcordeiro: Did you mean https://github.com/rebornplusplus/chisel-releases/pull/3?

.github/workflows/tests.yaml
```diff
@@ -0,0 +1,75 @@
...
+
+jobs:
```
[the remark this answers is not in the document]
rebornplusplus: Simplified in https://github.com/canonical/chisel-releases/pull/120/commits/366d73c78134305402816457ab623f5704c9fe5f. Please take another look.

.github/workflows/tests.yaml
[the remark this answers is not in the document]
rebornplusplus: I kept it as "tests" still because the new jobs are quite simplified for us to accommodate more future tests here if needed. Let me know if you still want to change the workflow name.

.github/workflows/tests.yaml
rebornplusplus: Updated, thanks!

.github/workflows/tests.yaml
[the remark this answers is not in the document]
rebornplusplus: Yeah, that too. rebornplusplus#3 is an example where it runs, rebornplusplus#5 where it doesn't (as it should).

.github/workflows/tests.yaml
```diff
@@ -8,68 +8,18 @@ on:
...
+    name: Installability
     strategy:
       fail-fast: false
       matrix:
         ARCH: [amd64, arm64, armhf, i386, ppc64el, riscv64, s390x]
     uses: canonical/chisel-releases/.github/workflows/install-slices.yaml@main
     with:
       arch: ${{ matrix.ARCH }}
-      files: ${{ needs.planner.outputs.changed-sdf }}
+      install-all: ${{ github.event_name == 'schedule' }}
```
cjdcordeiro, line 25: Looking much more maintainable now :) thanks

I also think you don't need this. You can have the schedule on `main`, with a matrix for every branch

.github/workflows/tests.yaml
```diff
@@ -8,68 +8,18 @@ on:
...
-    needs: planner
-    if: |
-      needs.planner.outputs.install-all != 'true' &&
-      needs.planner.outputs.install-changed == 'true'
+  installability:
+    name: Installability
     strategy:
       fail-fast: false
       matrix:
         ARCH: [amd64, arm64, armhf, i386, ppc64el, riscv64, s390x]
```
cjdcordeiro, line 21: this matrix can also be on `main`

.github/workflows/tests.yaml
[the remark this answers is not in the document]
cjdcordeiro: I think the question here is:
 - we'll have many more jobs, of different natures (e.g. CLA check, linting, etc.). Since we want to keep branches super clean and maintainable, we only have 2 choices: 1) this one single workflow, with a **super** generic name (e.g. `CI for ${{ github.ref_name }}`); or 2) have one workflow **per type** of job(s), which means this workflow would be something like "Reliability Tests".

Despite my first comment, I'm actually learning towards 1) as I believe it would increase maintainability

cjdcordeiro (changes requested):
so much better. 1 more pass and it should be good to go!

.github/workflows/tests.yaml
rebornplusplus: True. I can. But having the schedule on each branch would allow us in the future to define different schedules for different releases. For example, currently the ubuntu-24.04 branch might need more frequent testing than the other branches due to ever-changing paths in packages.

.github/workflows/tests.yaml
rebornplusplus: Done! https://github.com/canonical/chisel-releases/pull/120/commits/5b22495a138e38c97c6783986a013e3a4d915bc2

.github/workflows/tests.yaml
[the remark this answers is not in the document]
rebornplusplus: Ya, I also believe this would increase maintainability. I kept "Tests" again because it's quite generic for the things we might need to do. `CI for ${{ github.ref_name }}` might look a bit weird (and long) for PRs since the `ref_name` would be something like `<pr_number>/merge`.

.github/workflows/tests.yaml
cjdcordeiro: **If** and whenever that's needed, we can tackle it. For now let's focus on simplicity and maintainability. The only reason we have workflows in the `ubuntu-*` branches is due to a technical limitation in GitHub. The goal has always been to use `main`. That's the main (pun intended) reason it was created. So, whatever we put in these branches **must** be simple, portable and have no logic

.github/workflows/tests.yaml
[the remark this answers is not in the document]
cjdcordeiro: what about the CLA check? linting? policy enforcements (like cross-release portability)? commit compliance with conventional commits?

As for the dynamic name, you actually want it, because when you are in the job logs GUI, there's no indication of the branch it ran for. If you don't like `ref_name` there's always `github.ref`.

Remember that as branches grow, all of these runs will look the same in GUI. We don't want to be greping the logs just to find out which ref it's running against.

cjdcordeiro (changes requested):

.github/workflows/tests.yaml
rebornplusplus: Done. https://github.com/canonical/chisel-releases/pull/120/commits/137ab152f77fe0918d82d69a2ee84116ea22e780

.github/workflows/tests.yaml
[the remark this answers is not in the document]
rebornplusplus: Changed in https://github.com/canonical/chisel-releases/pull/120/commits/33dfb87f8a9fd7d14cdd64885dfd65a324dced80. Thanks.

cjdcordeiro (approved):
neat! gonna look at #119  one b4 merging this
thanks

linostar (approved):
LGTM, thanks!

cjdcordeiro:
ready to go. let's give it a try before merging the ones for newer releases (#140 #141 )

PR #122, opened 2024-02-06
feat(22.04): Use dash instead of bash for grep deprecated slice

lwnexgen:
As noted in #117 amd #118 by @cjdcordeiro  the `deprecated` slice is missing an essential def for `libpcre3`. Backports the fix from those PRs to 22.04.

### DIFF
--- a/slices/grep.yaml
+++ b/slices/grep.yaml
@@ -4,7 +4,7 @@ slices:
   bins:
     essential:
       - libc6_libs
-      - libpcre2-8-0_libs
+      - libpcre3_libs
     contents:
       /bin/grep:
 
@@ -13,7 +13,8 @@ slices:
     # See: https://manpages.ubuntu.com/manpages/jammy/en/man1/grep.1.html
     essential:
       - dash_bins
-      - libpcre2-8-0_libs
+      - grep_bins
+      - libpcre3_libs
     contents:
       /bin/egrep:
       /bin/fgrep:
--- a/slices/grep.yaml
+++ b/slices/grep.yaml
@@ -14,7 +14,6 @@ slices:
     essential:
       - dash_bins
       - grep_bins
-      - libpcre3_libs
     contents:
       /bin/egrep:
       /bin/fgrep:
--- a/slices/grep.yaml
+++ b/slices/grep.yaml
@@ -13,6 +13,7 @@ slices:
     # See: https://manpages.ubuntu.com/manpages/jammy/en/man1/grep.1.html
     essential:
       - bash_bins
+      - libpcre3_libs
     contents:
       /bin/egrep:
       /bin/fgrep:
--- a/slices/grep.yaml
+++ b/slices/grep.yaml
@@ -4,16 +4,16 @@ slices:
   bins:
     essential:
       - libc6_libs
-      - libpcre3_libs
+      - libpcre2-8-0_libs
     contents:
       /bin/grep:
 
   deprecated:
-    # These are shell scripts requiring a symlink from /usr/bin/bash to /bin/sh
+    # These are shell scripts requiring a symlink from /usr/bin/dash to /bin/sh
     # See: https://manpages.ubuntu.com/manpages/jammy/en/man1/grep.1.html
     essential:
-      - bash_bins
-      - libpcre3_libs
+      - dash_bins
+      - libpcre2-8-0_libs
     contents:
       /bin/egrep:
       /bin/fgrep:


cjdcordeiro (approved):
lgtm, thanks

slices/grep.yaml
```diff
@@ -13,6 +13,7 @@ slices:
     # See: https://manpages.ubuntu.com/manpages/jammy/en/man1/grep.1.html
     essential:
       - bash_bins
```
rebornplusplus, line 15: same as https://github.com/canonical/chisel-releases/pull/118#discussion_r1480907936

slices/grep.yaml
```diff
@@ -13,6 +13,7 @@ slices:
...
+      - libpcre3_libs
```
rebornplusplus, line 16: same as https://github.com/canonical/chisel-releases/pull/118#discussion_r1480910303

rebornplusplus (changes requested):

slices/grep.yaml
lwnexgen: Cool, I can do that. Didn't realize Debian / Ubuntu used Dash as the default for that ; when I used to do RHEL-like packaging it was always `bash`.

slices/grep.yaml
lwnexgen: Cool, I can do that. Didn't realize Debian / Ubuntu used Dash as the default for that ; when I used to do RHEL-like packaging it was always bash.

slices/grep.yaml
```diff
@@ -4,16 +4,16 @@ slices:
...
       /bin/grep:
 
   deprecated:
-    # These are shell scripts requiring a symlink from /usr/bin/bash to /bin/sh
+    # These are shell scripts requiring a symlink from /usr/bin/dash to /bin/sh
     # See: https://manpages.ubuntu.com/manpages/jammy/en/man1/grep.1.html
     essential:
-      - bash_bins
-      - libpcre3_libs
+      - dash_bins
```
cjdcordeiro, line 15: `dash` isn't yet sliced for 22.04 unfortunately

cjdcordeiro (changes requested):

slices/grep.yaml
lwnexgen: https://github.com/canonical/chisel-releases/blob/ubuntu-23.10/slices/dash.yaml includes the symlink that the package installs, it looks like. That make all of the link comments we've been throwing in obsolete, right, if I add a dash slice to `22.04`?

lwnexgen: Added #126 for Jammy `dash`

cjdcordeiro: merged

slices/grep.yaml
```diff
@@ -4,15 +4,16 @@ slices:
...
       /bin/grep:
 
   deprecated:
-    # These are shell scripts requiring a symlink from /usr/bin/bash to /bin/sh
+    # These are shell scripts requiring a symlink from /usr/bin/dash to /bin/sh
     # See: https://manpages.ubuntu.com/manpages/jammy/en/man1/grep.1.html
     essential:
-      - bash_bins
+      - dash_bins
+      - libpcre2-8-0_libs
```
cjdcordeiro, line 16: I think you misunderstood a previous review comment and changed this to libpcre2-8-0, but in 22.04, grep depends on libpcre3 -> https://packages.ubuntu.com/jammy/grep

cjdcordeiro (changes requested):

slices/grep.yaml
lwnexgen: Yep I sure did! I think it should be good now.

slices/grep.yaml
```diff
@@ -9,10 +9,11 @@ slices:
       /bin/grep:
 
   deprecated:
-    # These are shell scripts requiring a symlink from /usr/bin/bash to /bin/sh
+    # These are shell scripts requiring a symlink from /usr/bin/dash to /bin/sh
     # See: https://manpages.ubuntu.com/manpages/jammy/en/man1/grep.1.html
     essential:
-      - bash_bins
+      - dash_bins
+      - grep_bins
```
cjdcordeiro, line 16: out of curiosity, Why do you need grep_bins? aren't these purely relying on `dash` and `dash` only?

lwnexgen: They just shell out to `grep` so they'll fail if `grep_bins` isn't installed.

lwnexgen: This is what `egrep` is on my jammy instance:

```
#!/bin/sh
exec grep -E "$@"
```

rebornplusplus (approved):
Looks good to me now, thanks!

slices/grep.yaml
cjdcordeiro: Noted!

cjdcordeiro (approved):

PR #126, opened 2024-02-07
feat(22.04): add slice for dash

lwnexgen:
Adds slice for `dash` (already exists in `23.10` and `24.04`).


cjdcordeiro (approved):
LGTM thanks

PR #128, opened 2024-02-12
bug(22.04): make /var/run a symlink to /run

lwnexgen:
On full Ubuntu VMs/hardware installs, as well as the upstream Ubuntu containers, `/var/run` is a symlink to `/run`. Chiseled containers do not make this symlink - this causes issues with Docker Swarm installs (and probably other projects), which mount secrets into `/run`.

### DIFF
--- a/slices/base-files.yaml
+++ b/slices/base-files.yaml
@@ -36,8 +36,7 @@ slices:
       /var/cache/:
       /var/lib/:
       /var/log/:
-      # Recreate symlink that is in place
-      # on full Ubuntu installs and containers
+      # The /var/run symlink is enforced by the package's maintainer scripts
       /var/run/: {symlink: /run}
       /var/tmp/:
 
--- a/slices/base-files.yaml
+++ b/slices/base-files.yaml
@@ -36,7 +36,7 @@ slices:
       /var/cache/:
       /var/lib/:
       /var/log/:
-      /var/run/:
+      /var/run/: {symlink: /run}
       /var/tmp/:
 
   home:
--- a/slices/base-files.yaml
+++ b/slices/base-files.yaml
@@ -36,6 +36,8 @@ slices:
       /var/cache/:
       /var/lib/:
       /var/log/:
+      # Recreate symlink that is in place
+      # on full Ubuntu installs and containers
       /var/run/: {symlink: /run}
       /var/tmp/:
 


linostar (approved):

cjdcordeiro (commented):
True. Thanks for the suggestion @lwnexgen 

@rebornplusplus , the `base-files` postinst is a bit weird, cause even though this is a regular directory in the deb, the maintainer script will proceed with its recreation as a symlink.

```bash
migrate_directory() {
  if [ ! -L $1 ]; then
    rmdir $1
    ln -s $2 $1
  fi
}
```

There's no way we can do this in a mutation script right?

rebornplusplus:
So we can compare, but we cannot really create symlinks in mutation scripts though.

> @rebornplusplus , the `base-files` postinst is a bit weird, cause even though this is a regular directory in the deb, the maintainer script will proceed with its recreation as a symlink.
> 
> ```shell
> migrate_directory() {
>   if [ ! -L $1 ]; then
>     rmdir $1
>     ln -s $2 $1
>   fi
> }
> ```
> 
> There's no way we can do this in a mutation script right?

slices/base-files.yaml
```diff
@@ -36,7 +36,7 @@ slices:
       /var/cache/:
       /var/lib/:
       /var/log/:
-      /var/run/:
+      /var/run/: {symlink: /run}
```
cjdcordeiro, line 39: Given the mutation script's limitation and the fact that the corresponding maintainer script is quite deterministic about this, I think it is reasonable do have this symlink here.

Could you please add a comment for it though? Just so that users understand "why" this is a symlink instead of the regular folder as in the deb's contents.

cjdcordeiro (changes requested):

slices/base-files.yaml
lwnexgen: Yep, done.

slices/base-files.yaml
```diff
@@ -36,7 +36,9 @@ slices:
       /var/cache/:
       /var/lib/:
       /var/log/:
-      /var/run/:
+      # Recreate symlink that is in place
```
cjdcordeiro, line 39: can you pls apply the same as in #130

cjdcordeiro (changes requested):

cjdcordeiro (approved):
great, thanks!

PR #129, opened 2024-02-12
bug(23.10): make /var/run a symlink to /run

lwnexgen:
On full Ubuntu VMs/hardware installs, as well as the upstream Ubuntu containers, /var/run is a symlink to /run. Chiseled containers do not make this symlink - this causes issues with Docker Swarm installs (and probably other projects), which mount secrets into /run.

### DIFF
--- a/slices/base-files.yaml
+++ b/slices/base-files.yaml
@@ -36,8 +36,7 @@ slices:
       /var/cache/:
       /var/lib/:
       /var/log/:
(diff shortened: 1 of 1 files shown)


linostar (approved):

rebornplusplus (approved):
Thanks!

cjdcordeiro (approved):
LGTM as well ;) ty

PR #131, opened 2024-02-12
ci(cla-check): add CLA check for the repo

cjdcordeiro:
The proposed GH workflow will check if the author of the PR has signed our CLA.

This is a reusable workflow that can be called from the release branches.

Tested in https://github.com/cjdcordeiro/chisel-releases/actions/runs/7876013322.

### DIFF
--- a/.github/workflows/cla-check.yaml
+++ b/.github/workflows/cla-check.yaml
@@ -1,0 +1,11 @@
+name: CLA Check
+on:
+  pull_request:
+  workflow_call:
+
+jobs:
+  cla-check:
+    runs-on: ubuntu-20.04
+    steps:
+      - name: Check if CLA signed
+        uses: canonical/has-signed-canonical-cla@v1


linostar (approved):
LGTM

.github/workflows/cla-check.yaml
```diff
@@ -0,0 +1,11 @@
+name: CLA Check
+on:
+  pull_request:
+  workflow_call:
+
+jobs:
+  cla-check:
+    runs-on: ubuntu-20.04
```
rebornplusplus, line 8: nit: why not `ubuntu-latest` though?

rebornplusplus (approved):

.github/workflows/cla-check.yaml
cjdcordeiro: not special reason. I just prefer to avoid the risk of it failing in the future due to some incompatibility with a release that is yet to come

cjdcordeiro:
the 24.04 errors are related to the SDFs and upstream changes (and not this PR). merging

PR #132, opened 2024-02-12
ci(cla-check): add CLA check for ubuntu-20.04

cjdcordeiro:
Needs #120 

This PR adds a new job that will call the CLA check reusable workflow (from https://github.com/canonical/chisel-releases/pull/131) on every PR.


.github/workflows/ci.yaml
```diff
@@ -0,0 +1,16 @@
+name: CI
+run-name: CI for ${{ github.ref }}
+
+on:
+  push:
```
linostar, line 5: What's the use of `on push` here if the jobs only works for pull_request events?

linostar (changes requested):
What

.github/workflows/ci.yaml
cjdcordeiro: Cause this will be merged with the work being done in #120 . It should all be part of the same workflow

linostar (approved):
LGTM

rebornplusplus (approved):
Looks good to me, thanks.

But let's block this until #120 and #131 are in.

PR #134, opened 2024-02-13
chore: add PR template

cjdcordeiro:
This PR adds a new template for all future new PRs.

Here's an example: https://github.com/cjdcordeiro/chisel-releases/pull/7

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,0 +1,40 @@
+# Proposed changes
+<!-- Describe the changes proposed in this PR.
+
+Provide good PR descriptions as the project maintainers aren't necessarily
+familiar with the packages you are slicing.
+
+We use conventional commits
+(https://www.conventionalcommits.org/en/v1.0.0/#specification), so if not yet
+specified in your commit messages, make sure you describe the type of change
+being proposed in this PR (i.e. feat, test, fix, ci, chore, docs).
+-->
+
+## Related issues/PRs
+<!-- If any -->
+
+### Forward porting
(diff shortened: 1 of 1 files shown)


.github/pull_request_template.md
```diff
@@ -0,0 +1,40 @@
...
+
+## Testing
+<!-- Provide proof of testing and/or testing instructions, when applicable,
+to help speed up the review process. -->
+
+```bash
+# Example:
+#  chisel cut ... --root <path> <pkg>_<proposed-slice>
+#  sudo chroot <path> <app>
+```
```
rebornplusplus, line 28: After #120 is merged, will this be needed? In my opinion, this can be part of the "Additional Context" section.

linostar (changes requested):
LGTM, with exception for the Testing section which can be removed after PR #120 is merged.

.github/pull_request_template.md
cjdcordeiro: I was thinking about functional tests. It's something our CI won't cover for now. I could change the wording, but I think it would be useful if contributors could provide some reproducible way for us (reviewers) to ensure the resulting slices are functional in a real life scenario (e.g. if someone is slicing nodejs, I could use a sample app and execution instructions to test with). Wdyt?

rebornplusplus: Ah that. Yeah, sure, I am okay with that. Let's keep it then.

rebornplusplus (approved):
LGTM.

linostar (approved):
LGTM.

PR #139, opened 2024-02-21
Add slice for `libicu74`

lbussell:
Needed for https://github.com/dotnet/dotnet-docker/issues/5181.

I see that there are two versions of `libicu` in Noble - 72 and 74. Is this temporary and will Noble only ship with version 74?


rebornplusplus (approved):
Looks good to me, thank you!

> I see that there are two versions of `libicu` in Noble - 72 and 74. Is this temporary and will Noble only ship with version 74?

I cannot confirm personally, but it could be the case given that the source package [icu](https://packages.ubuntu.com/source/noble/icu) is of version 74.x.

lbussell:
@cjdcordeiro, can you please take a look and merge when appropriate? Thanks!

cjdcordeiro (approved):
All green! LGTM, thanks ;)

PR #144, opened 2024-02-23
chore: fix archive version while logging

rebornplusplus:
Archive version was being changed to 23.1 instead 23.10 for values with trailing zeroes. This PR fixes the bug.

https://github.com/canonical/chisel-releases/actions/runs/8014505888/job/21893215155?pr=14

### DIFF
--- a/.github/scripts/install_slices/install_slices.py
+++ b/.github/scripts/install_slices/install_slices.py
@@ -128,9 +128,10 @@ def parse_archive(release: str) -> Archive:
         sys.exit(1)
     # load the yaml data into Archive
     archive_data = data["archives"]["ubuntu"]
-    archive = Archive(
-        str(archive_data["version"]), archive_data["components"], archive_data["suites"]
-    )
+    version = archive_data["version"]
+    if isinstance(version, float):
+        version = f"{version:.2f}"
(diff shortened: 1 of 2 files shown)


.github/scripts/install_slices/install_slices.py
```diff
@@ -128,9 +128,10 @@ def parse_archive(release: str) -> Archive:
         sys.exit(1)
     # load the yaml data into Archive
     archive_data = data["archives"]["ubuntu"]
-    archive = Archive(
-        str(archive_data["version"]), archive_data["components"], archive_data["suites"]
-    )
+    version = archive_data["version"]
+    if isinstance(version, float):
```
cjdcordeiro, line 132: is there ever a time where this won't be a float?

cjdcordeiro (approved):
good catch. lgtm

.github/scripts/install_slices/install_slices.py
linostar: > is there ever a time where this won't be a float?

I think that may happen if something encloses the version with quotes.

linostar:
The CI tests are showing us now what we have missed in several slices because we couldn't manually test them on every architecture :)
image

PR #148, opened 2024-02-23
fix(24.04): fix release errors

cjdcordeiro:
This release is in development and thus packages keep changing upstream.

This PR fixes the issues found in the [current tests](https://github.com/canonical/chisel-releases/actions/runs/8019314799). Those include various path re-definitions and the removal of `libunwind-13`


rebornplusplus (approved):
Wow, that is a lot of changes. Thanks for adding the PR!

slices/base-files.yaml
```diff
@@ -16,14 +16,14 @@ slices:
 
   bin:
     contents:
-      /bin/:
-      /sbin/:
```
letFunny, line 20: Out of curiosity, why are we removing the trailing slash here? It seems like we are using it below, e.g. `/usr/bin/:`

cjdcordeiro: these are symlinks now

linostar (approved):
LGTM, thanks!

slices/base-files.yaml
letFunny: That means that it is extracting symlinks included in the tarball?

letFunny (approved):
Thanks!

slices/base-files.yaml
cjdcordeiro: Yes, exactly

cjdcordeiro: Please re-open if you think this should be done differently :)

PR #149, opened 2024-02-23
fix(22.04): fix arch specific paths for openjdk-8

cjdcordeiro:
Certain paths were not valid for all architectures. This PR adds the arch-specific constraints.
@vpa1977 can you pls double check if it makes sense?

PR #150, opened 2024-02-23
fix(20.04): fix arch specific paths for openjdk-8

cjdcordeiro:
Same as #149 

@vpa1977 can you also pls double check this one?

### DIFF
--- a/slices/openjdk-8-jre-headless.yaml
+++ b/slices/openjdk-8-jre-headless.yaml
@@ -29,8 +29,16 @@ slices:
       /usr/lib/jvm/java-8-openjdk-*/jre/lib/*/libjsig.so:
       /usr/lib/jvm/java-8-openjdk-*/jre/lib/*/client/libjsig.so: {arch: armhf}
       /usr/lib/jvm/java-8-openjdk-*/jre/lib/*/client/libjvm.so: {arch: armhf}
-      /usr/lib/jvm/java-8-openjdk-*/jre/lib/*/server/libjsig.so: {arch: [amd64,arm64,ppc64el]}
-      /usr/lib/jvm/java-8-openjdk-*/jre/lib/*/server/libjvm.so: {arch: [amd64,arm64,ppc64el]}
+      /usr/lib/jvm/java-8-openjdk-*/jre/lib/*/server/libjsig.so:
+        arch:
+          - amd64
+          - arm64
+          - ppc64el
+      /usr/lib/jvm/java-8-openjdk-*/jre/lib/*/server/libjvm.so:
+        arch:
+          - amd64
+          - arm64
+          - ppc64el
       /usr/lib/jvm/java-8-openjdk-*/jre/lib/hijrah-config-umalqura.properties:
       /usr/lib/jvm/java-8-openjdk-*/jre/lib/calendars.properties:
       /usr/lib/jvm/java-8-openjdk-*/jre/lib/content-types.properties:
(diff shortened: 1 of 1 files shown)


vpa1977:
/client and /server hotspot machines could be built together and present in the same installation, though it is not something that openjdk-8 packaging does. 
Maybe would it be possible to split those into separate slices (if chisel allows) ? 
Otherwise the change look fine.

cjdcordeiro:
> /client and /server hotspot machines could be built together and present in the same installation, though it is not something that openjdk-8 packaging does. Maybe would it be possible to split those into separate slices (if chisel allows) ? Otherwise the change look fine.

when you say 

> not something that openjdk-8 packaging does

can you please clarify? for `armhf` it seems that there's only `client`, while `server` is only available for other archs. Are you saying that, for `armhf`, the `client` should not be installed as part of the current slice?

vpa1977:
> can you please clarify? for `armhf` it seems that there's only `client`, while `server` is only available for other archs. Are you saying that, for `armhf`, the `client` should not be installed as part of the current slice?

I apologise for confusion. Yes, we build client for the armhf. 
I was wondering if we can do something like:

```
slices:
  # The "core" slice provides a minimal, yet functional JRE8.
  core:
    essential:
      - libc6_libs
      - libgcc-s1_libs
      - libstdc++6_libs
      - zlib1g_libs
      - openjdk-8-jre-headless_client_vm
      - openjdk-8-jre-headless_server_vm

... 
```
and move the client and server vm contents into separate slices.

slices/openjdk-8-jre-headless.yaml
```diff
@@ -27,8 +27,10 @@ slices:
       /usr/lib/jvm/java-8-openjdk-*/jre/lib/*/libverify.so:
       /usr/lib/jvm/java-8-openjdk-*/jre/lib/*/libzip.so:
       /usr/lib/jvm/java-8-openjdk-*/jre/lib/*/libjsig.so:
-      /usr/lib/jvm/java-8-openjdk-*/jre/lib/*/server/libjsig.so:
-      /usr/lib/jvm/java-8-openjdk-*/jre/lib/*/server/libjvm.so:
+      /usr/lib/jvm/java-8-openjdk-*/jre/lib/*/client/libjsig.so: {arch: armhf}
+      /usr/lib/jvm/java-8-openjdk-*/jre/lib/*/client/libjvm.so: {arch: armhf}
+      /usr/lib/jvm/java-8-openjdk-*/jre/lib/*/server/libjsig.so: {arch: [amd64,arm64,ppc64el]}
+      /usr/lib/jvm/java-8-openjdk-*/jre/lib/*/server/libjvm.so: {arch: [amd64,arm64,ppc64el]}
```
rebornplusplus, line 33: nit: https://github.com/canonical/chisel-releases/pull/149#discussion_r1503688416.

rebornplusplus (approved):
Looks good to me. I will let you guys sort out the organizing of files.

slices/openjdk-8-jre-headless.yaml
cjdcordeiro: done

cjdcordeiro:
> > can you please clarify? for `armhf` it seems that there's only `client`, while `server` is only available for other archs. Are you saying that, for `armhf`, the `client` should not be installed as part of the current slice?
> 
> I apologise for confusion. Yes, we build client for the armhf. I was wondering if we can do something like:
> 
> ```
> slices:
>   # The "core" slice provides a minimal, yet functional JRE8.
>   core:
>     essential:
>       - libc6_libs
>       - libgcc-s1_libs
>       - libstdc++6_libs
>       - zlib1g_libs
>       - openjdk-8-jre-headless_client_vm
>       - openjdk-8-jre-headless_server_vm
> 
> ... 
> ```
> 
> and move the client and server vm contents into separate slices.

Ah...I see. We could do that indeed. It would be a refactoring task that I would address separately from this PR though, cause it would mean doing it for other releases too. See https://github.com/canonical/chisel-releases/issues/152

linostar (approved):
LGTM.

Issue #153, opened 2024-02-27
ubuntu-24.04 release doesn't seem to work


norrisjeremy:
I am unable to build a simple chiseled container using `ubuntu-24.04`:

Example Dockerfile:
```
FROM ubuntu:22.04 AS chisel
ADD https://github.com/canonical/chisel/releases/download/v0.9.0/chisel_v0.9.0_linux_amd64.tar.gz chisel.tar.gz
RUN apt-get update && \
    DEBIAN_FRONTEND=noninteractive apt-get install -y ca-certificates && \
    tar -xvf chisel.tar.gz -C /usr/bin/ && \
    mkdir -p /rootfs && \
    chisel cut --release ubuntu-24.04 --root /rootfs \
      base-files_base \
      base-files_release-info \
      base-passwd_data \
      libc6_config \
      libc6_libs \
      libgcc-s1_libs \
      dash_bins

FROM scratch
COPY --from=chisel /rootfs /
ENTRYPOINT ["/bin/sh", "-c"]
CMD ["echo foo"]
```

If I then try to build & execute this container, I get an error:
```
$ docker build . -t foo:1 && docker run -it --rm foo:1
...
exec /bin/sh: no such file or directory
```

Simply changing to `ubuntu-22.04` as follows:
```
FROM ubuntu:22.04 AS chisel
ADD https://github.com/canonical/chisel/releases/download/v0.9.0/chisel_v0.9.0_linux_amd64.tar.gz chisel.tar.gz
RUN apt-get update && \
    DEBIAN_FRONTEND=noninteractive apt-get install -y ca-certificates && \
    tar -xvf chisel.tar.gz -C /usr/bin/ && \
    mkdir -p /rootfs && \
    chisel cut --release ubuntu-22.04 --root /rootfs \
      base-files_base \
      base-files_release-info \
      base-passwd_data \
      libc6_config \
      libc6_libs \
      libgcc-s1_libs \
      dash_bins

FROM scratch
COPY --from=chisel /rootfs /
ENTRYPOINT ["/bin/sh", "-c"]
CMD ["echo foo"]
```

And then building & executing this container works:
```
$ docker build . -t bar:1 && docker run -it --rm bar:1
...
foo
```

Is there some package/slice that was introduced in 24.04 that I am missing that needs to be added in order to build a simple chiseled container?

cjdcordeiro:
Hi.

[...]

PR #154, opened 2024-02-27
fix(24.04): missing lib64 in base-files

cjdcordeiro:
fixes https://github.com/canonical/chisel-releases/issues/153

`/lib64` is now in `base-files`, so this PR includes said paths in its `libs` slice.


norrisjeremy:
Thanks!

norrisjeremy:
FYI, I believe `/lib64` is owned by the `libc6` package, not `base-files`.
That is what shows up when I run `dpkg -S /lib64` on a regular 24.04 container.
I think that's why your CI builds fail.

norrisjeremy:
```diff --git a/slices/libc6.yaml b/slices/libc6.yaml
index 5f67118..7310473 100644
--- a/slices/libc6.yaml
+++ b/slices/libc6.yaml
@@ -28,3 +28,4 @@ slices:
       /usr/lib/*-linux-*/libthread_db.so.*:
       /usr/lib/*-linux-*/libutil.so.*:
       /usr/lib*/ld*.so.*:
+      /lib64:
```

norrisjeremy:
```
$ docker run -it --rm ubuntu:24.04 dpkg -S /lib64  
libc6:amd64: /lib64
```

norrisjeremy:
```
diff --git a/slices/libc6.yaml b/slices/libc6.yaml
index 5f67118..fcd25b4 100644
--- a/slices/libc6.yaml
+++ b/slices/libc6.yaml
@@ -28,3 +28,4 @@ slices:
       /usr/lib/*-linux-*/libthread_db.so.*:
       /usr/lib/*-linux-*/libutil.so.*:
       /usr/lib*/ld*.so.*:
+      /lib64: {arch: [amd64,ppc64el]}
```

cjdcordeiro:
The errors are coming from the fact that `/lib64` isn't present for all architectures in `base-files`.

Noble is a development release and thus changing all the time. We'll take this internally to double check where `/lib64` is eventually landing and whether it should belong to `base-files` or `libc6`. The symlink definition, according to `dpkg -c`, is currently under `base-files`. 

```bash
# dpkg -c base-files_13ubuntu7_amd64.deb | grep lib64
drwxr-xr-x root/root         0 2024-02-22 14:39 ./usr/lib64/
lrwxrwxrwx root/root         0 2024-02-22 14:39 ./lib64 -> usr/lib64
```

But this could be changing as we speak so let's wait for the upstream to finish updating.

norrisjeremy:
Yes, you are correct about the CI errors.
However, I really think the `/lib64` symlink needs to be created by the `libc6_libs` slice.
Otherwise, it is impossible to execute any dynamically linked executable unless a user depends upon the `base-files_lib` slice instead of simply replying upon `libc6_libs` as they used to.

norrisjeremy:
Apologies for earlier reply, I misread your response.

rebornplusplus (approved):
LGTM, thank you.

slices/base-files.yaml
```diff
@@ -25,6 +25,8 @@ slices:
     contents:
       /lib:
       /usr/lib/:
+      /lib64:
+      /usr/lib64/:
```
rebornplusplus, line 29: These seem to be available only in a handful or arch. Maybe we should specify them.

rebornplusplus (changes requested):

slices/base-files.yaml
cjdcordeiro: Y for sure. Just wanted to make sure those archs weren't being updated upstream. Should be good now

norrisjeremy (approved):

rebornplusplus (approved):
LGTM.

slices/base-files.yaml
```diff
@@ -25,6 +25,8 @@ slices:
     contents:
       /lib:
       /usr/lib/:
+      /lib64: {arch: [amd64,ppc64el]}
```
letFunny, line 28: Open question, would it be better to specify the symlink in Chisel instead of relying in the one from the tarball? I find the difference between `lib64/` (extract folder) and `lib64` (extract symlink) not that clear.

letFunny (approved):
LGTM

slices/base-files.yaml
cjdcordeiro: It is definitely something we can discuss. I'm not sure we should let chisel make that decision though, cause then you'd lose the freedom of choice on whether to include the symlinks or not. We also have the `symlink` path property, which probably could be used here to reduce the confusion, if needed

PR #187, opened 2024-03-12
chore: add workflow to check package dependencies

rebornplusplus:
# Proposed changes

This PR adds a new reusable workflow which can be run on Pull Requests to see if the packages listed in "essential" of a slice are the true dependencies of the slice package. This workflow will add a PR comment if it finds any errors.

## Related issues/PRs
https://github.com/canonical/chisel-releases/pull/188

## Testing
See demo at https://github.com/rebornplusplus/chisel-releases/pull/9.

## Checklist

### DIFF
--- a/.github/workflows/pkg-deps.yaml
+++ b/.github/workflows/pkg-deps.yaml
@@ -40,7 +40,7 @@ jobs:
       - name: Check dependencies
         id: check-deps
         env:
-          script-dir: "${{ env.main-branch-path }}/.github/scripts/install_slices"
+          script-dir: "${{ env.main-branch-path }}/.github/scripts/pkg-deps"
         run: |
           set -ex
           ./${{ env.script-dir }}/pkg-deps \
--- a/.github/workflows/pkg-deps.yaml
+++ b/.github/workflows/pkg-deps.yaml
@@ -1,0 +1,102 @@
+name: Package dependencies
+
+on:
+  workflow_call:
+
+jobs:
+  check-dependency:
+    name: Check dependency
+    runs-on: ubuntu-latest
+    if: |
+      github.event_name == 'pull_request' &&
+      startswith(github.base_ref, 'ubuntu-')
+    env:
+      branch: ${{ github.base_ref }}
+      msg_file: dependencies.message
+    permissions:
+      pull-requests: write
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Check changed paths
+        id: changed-paths
+        uses: dorny/paths-filter@v3
+        with:
+          # ref: https://github.com/marketplace/actions/paths-changes-filter
+          filters: |
+            slices:
+              - added|modified: 'slices/**/*.yaml'
+          # Space delimited list usable as command-line argument list in
+          # Linux shell. If needed, it uses single or double quotes to
+          # wrap filename with unsafe characters.
+          list-files: shell
+
+      - name: Check dependencies
+        id: check-deps
+        if: steps.changed-paths.outputs.slices == 'true'
+        run: |
+          set -e
+
+          export LC_COLLATE=C
+
+          version=$(echo "$branch" | grep -Eo '[0-9.]+')
+          docker run -i -d --rm --name ubuntu ubuntu:"$version"
+          docker exec ubuntu apt-get update
+
+          cleanup() {
+            docker stop ubuntu
+          }
+          trap cleanup EXIT
+
+          echo "Slice package dependency issues:" > "$msg_file"
+          for f in ${{ steps.changed-paths.outputs.slices_files }}; do
+            echo "Processing $f.." >&2
+            pkg=$(yq '.package' "$f")
+
+            mapfile -t deps < <(
+              docker exec ubuntu apt depends \
+                --no-recommends --no-suggests --no-conflicts \
+                --no-breaks --no-replaces --no-enhances \
+                "$pkg" 2>/dev/null | \
+              sed -nr 's/.*Depends:\s(\S*).*/\1/p' | \
+              sed 's/<//; s/>//; s/:any//' | \
+              sort | uniq
+            )
+
+            mapfile -t listed < <(
+              yq '.slices.[].essential[]' "$f" | \
+              sed "s/_.*//; /^$pkg$/d" | sort | uniq
+            )
+
+            errs=()
+            for p in "${listed[@]}"; do
+              if printf '%s\0' "${deps[@]}" | grep -Fxqz -- "$p"; then
+                continue
+              fi
+              errs+=("$p")
+            done
+
+            if (( "${#errs[@]}")); then
+              echo "<details>" >> "$msg_file"
+              echo -e "<summary>$f</summary>\n" >> "$msg_file"
+              for e in "${errs[@]}"; do
+                echo "  - \`$e\` is listed in \`essential\`, but is not a dependency." >> "$msg_file"
+              done
+              echo -e "\n</details>" >> "$msg_file"
+            fi
+          done
+          echo -e "\n---" >> "$msg_file"
+          cat "$msg_file"
+
+      - name: Craft default message if no errors
+        run: |
+          set -e
+          if [[ ! -f "$msg_file" ]]; then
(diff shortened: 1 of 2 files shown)


.github/workflows/pkg-deps.yaml
```diff
@@ -0,0 +1,102 @@
...
+      - name: Check dependencies
+        id: check-deps
+        if: steps.changed-paths.outputs.slices == 'true'
+        run: |
+          set -e
+
+          export LC_COLLATE=C
+
+          version=$(echo "$branch" | grep -Eo '[0-9.]+')
+          docker run -i -d --rm --name ubuntu ubuntu:"$version"
```
cjdcordeiro, line 43: fyi, for future reference, you can run GH jobs in a container: https://docs.github.com/en/actions/using-jobs/running-jobs-in-a-container

(probably wouldn't help much here though, but still :) )

.github/workflows/pkg-deps.yaml
```diff
@@ -0,0 +1,102 @@
...
+          docker exec ubuntu apt-get update
+
+          cleanup() {
+            docker stop ubuntu
+          }
+          trap cleanup EXIT
```
cjdcordeiro, line 49: you may want to do this right after the `docker run`

and btw, why `stop` and not `rm -f`?

.github/workflows/pkg-deps.yaml
```diff
@@ -0,0 +1,102 @@
...
+
+          echo "Slice package dependency issues:" > "$msg_file"
```
cjdcordeiro, line 51:
```suggestion
          echo "Diff of dependencies:" > "$msg_file"
```

let's not call it "issues". We don't want to convey the message that something is wrong. It might not be. this is purely informative.

cjdcordeiro, line 51: If you're turning this into a standalone bash script, you may want to consider having msg_file as a `mktemp` file, so we can safely run this locally without worrying about overwriting prior diffs nor polluting the local filesystem. Eg.:

```bash
msg_file=`mktemp`
echo "Writing dependencies diff to $msg_file"
if [ -n "$GITHUB_OUTPUT" ]
then
    echo "msg_file=$msg_file" >> $GITHUB_OUTPUT
fi
```

.github/workflows/pkg-deps.yaml
```diff
@@ -0,0 +1,102 @@
...
+              sed 's/<//; s/>//; s/:any//' | \
+              sort | uniq
+            )
+
+            mapfile -t listed < <(
+              yq '.slices.[].essential[]' "$f" | \
+              sed "s/_.*//; /^$pkg$/d" | sort | uniq
+            )
+
+            errs=()
```
cjdcordeiro, line 71: nitpick: not really an `err`

.github/workflows/pkg-deps.yaml
```diff
@@ -0,0 +1,102 @@
...
+                continue
+              fi
+              errs+=("$p")
+            done
+
+            if (( "${#errs[@]}")); then
+              echo "<details>" >> "$msg_file"
+              echo -e "<summary>$f</summary>\n" >> "$msg_file"
+              for e in "${errs[@]}"; do
+                echo "  - \`$e\` is listed in \`essential\`, but is not a dependency." >> "$msg_file"
```
cjdcordeiro, line 83: would it be possible to have this as a real `diff` instead? This way we don't write any text that might convey whether things are erratic or not

.github/workflows/pkg-deps.yaml
```diff
@@ -0,0 +1,102 @@
...
+          echo -e "\n---" >> "$msg_file"
+          cat "$msg_file"
+
+      - name: Craft default message if no errors
+        run: |
+          set -e
+          if [[ ! -f "$msg_file" ]]; then
+            echo "Slice package dependency issues:" > "$msg_file"
+            echo -e "\tNone found." >> "$msg_file"
+          fi
```
cjdcordeiro, line 97: Same comments as above about the "error" connotation.

And maybe this snippet can be merged with the script above? So there's no need to duplicate the echo of the first line or  assume the existence of the msg_file

cjdcordeiro (changes requested):
this is looking really nice. I think all comments are purely about refactoring and aesthetics.

.github/workflows/pkg-deps.yaml
```diff
@@ -0,0 +1,102 @@
...
+
+on:
+  workflow_call:
+
+jobs:
+  check-dependency:
+    name: Check dependency
+    runs-on: ubuntu-latest
+    if: |
+      github.event_name == 'pull_request' &&
```
linostar, line 11: I am not 100% sure, but I think that approach won't work with workflow_call, as the github.event_name received here will be "workflow_call" and not that of the calling parent "pull_request", so github.base_ref won't be accessible either. The alternative is to use `inputs` under `workflow_call` to pass the branch name.

linostar (changes requested):

.github/workflows/pkg-deps.yaml
rebornplusplus: No no. This will work.

> When a reusable workflow is triggered by a caller workflow, the github context is always associated with the caller workflow. The called workflow is automatically granted access to github.token and secrets.GITHUB_TOKEN. For more information about the github context, see "[Contexts](https://docs.github.com/en/actions/learn-github-actions/contexts#github-context)."

-- https://docs.github.com/en/actions/using-workflows/reusing-workflows

We are already doing some of this in the `install-slices` workflow.

.github/workflows/pkg-deps.yaml
rebornplusplus: Ya, it won't help much here tbh. But this is nice to know, thank you!

.github/workflows/pkg-deps.yaml
rebornplusplus: I wanted to be nice with the stopping, but oka, I will do `rm -f`. :laughing:

.github/workflows/pkg-deps.yaml
rebornplusplus: How do you mean? `diff` from what?

rebornplusplus:
Note that this PR is running the "Install Slices" tests. Raised #192 to remove them.

.github/workflows/pkg-deps.yaml
cjdcordeiro: something like

<details><summary>Diff of dependencies</summary>
<p>
```diff
diff --git a/pkg-deps b/slice-deps
index 86e041d..b7b05f8 100644
--- a/pkg-deps
+++ b/slice-deps
@@ -1,3 +1,2 @@
-foo
 bar
-baz
+something-else
```
</p>
</details>

rebornplusplus: I still do not understand to be honest. Do you want something like the following

> Diff of dependencies:
> <details>
> <summary>slices/coreutils.yaml</summary>
> 
> ```diff
> + libcrypt1
> + libgcc-s1
> + libpcre2-8-0
> ```
> </details>

instead of below?

> Diff of dependencies:
> <details>
> <summary>slices/coreutils.yaml</summary>
> 
>   - `libcrypt1` is listed in `essential`, but is not a dependency.
>   - `libgcc-s1` is listed in `essential`, but is not a dependency.
>   - `libpcre2-8-0` is listed in `essential`, but is not a dependency.
>
> </details>

rebornplusplus: Added https://github.com/canonical/chisel-releases/pull/187/commits/a011766c35308ef500f0fad78316e6a6af8ab217. The comment will now look like the following:

> Diff of dependencies:
> 
> <details>
> <summary>slices/coreutils.yaml</summary>
> 
> ```diff
> @@ -1,5 +1,8 @@
>  libacl1
> -libattr1
>  libc6
> +libcrypt1
> +libgcc-s1
>  libgmp10
> +libpcre2-8-0
>  libselinux1
> ```
> 
> </details>
>

linostar (approved):
Good job!

PR #196, opened 2024-03-14
chore: fix permissions of pkg-deps job

rebornplusplus:
This PR adds the permission required to write in PR to the `pkg-deps` job in CI.

There have been some failures like this: https://github.com/canonical/chisel-releases/actions/runs/8268278480/job/22620739872?pr=195.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -22,5 +22,7 @@ jobs:
 
   pkg-deps:
     if: github.event_name == 'pull_request'
+    permissions:
+      pull-requests: write
     name: Package dependencies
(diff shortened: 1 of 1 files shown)


cjdcordeiro:
I wonder is this is the cause, or if maybe the `ubuntu-*` branch job that calls the reusable wf also needs this permission :thinking: I hope not :grimacing:

rebornplusplus:
I think we will need to configure a few settings in the repository actually. I was hoping to sync on it in the stand-up.

rebornplusplus:
Opened #198 and #199 for the fixes. Closing this in favour of them.

PR #198, opened 2024-03-15
chore(main): pkg-deps workflow supports pull_request_target

rebornplusplus:
# Proposed changes

This PR amends the `pkg-deps` workflow to support `pull_request_target` events, which is what we will be needing from the caller workflow.

## Related issues/PRs
22.04: #199 

## Testing
Demo: https://github.com/rebornplusplus/chisel-releases/pull/10

## Checklist

* [x] I have read the [contributing guidelines](
https://github.com/canonical/chisel-releases/blob/main/CONTRIBUTING.md)
* [x] I have already submitted the [CLA form](
https://ubuntu.com/legal/contributors/agreement)

PR #199, opened 2024-03-15
chore(22.04): fix pkg-deps job

rebornplusplus:
# Proposed changes
This PR fixes the `pkg-deps` job with proper `event` and permissions. The `pkg-deps` job is moved from the `ci.yaml` workflow to a new `ci-pr-target.yaml` workflow which accommodates jobs which need the `pull_request_target` event.

## Related issues/PRs
#198

## Testing
Demo: https://github.com/rebornplusplus/chisel-releases/pull/10

## Checklist

### DIFF
--- a/.github/workflows/ci-pr-target.yaml
+++ b/.github/workflows/ci-pr-target.yaml
@@ -1,0 +1,14 @@
+# This workflow supports jobs which needs the on.pull_request_target event
+# instead of the usual on.pull_request event.
+name: CI
+run-name: CI for ${{ github.ref }}
+
+on:
+  pull_request_target:
+    branches:
+      - "ubuntu-*"
+
+jobs:
+  pkg-deps:
+    name: Package dependencies
+    uses: rebornplusplus/chisel-releases/.github/workflows/pkg-deps.yaml@main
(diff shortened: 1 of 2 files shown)


.github/workflows/ci-pr-target.yaml
```diff
@@ -0,0 +1,14 @@
+# This workflow supports jobs which needs the on.pull_request_target event
+# instead of the usual on.pull_request event.
+name: CI
```
cjdcordeiro, line 3: Good digging into the problem :+1: 

Since we are forced to have 2 wf files, we should probably rename them, otherwise is gets hard to identify them from the Actions tab.

Before we had one big wf for "CI" in general. Now we two, we can be a bit more specific, especially about this one. Suggestion: `name: Pkg coverage` (as the incoming last wf will also need the same PR commenting rights)

cjdcordeiro (changes requested):
ping @linostar so we can merge it today

.github/workflows/ci-pr-target.yaml
rebornplusplus: Changed in https://github.com/canonical/chisel-releases/pull/199/commits/a4e618255f8e5363d9e9cbe212a20717ef95a874.

cjdcordeiro (approved):
lgtm thanks

linostar (approved):
LGTM!

PR #207, opened 2024-03-31
feat(22.04): Add python bin symlinks

bryantbiggs:
# Proposed changes

- Add python bin symlinks. Most applications that reach out to python do so under the `python` or `python3` bin names; these symlinks help maintain that compatibility

Directory structure on `python:slim` image:
image

## Related issues/PRs
- #207
- #208
- #209
- #210

### Forward porting

🆘 Please let me know how this works - do I need to open PRs against each branch? 

## Testing

```Dockerfile
ARG UBUNTU_RELEASE=22.04

# Build the chiselled filesystem based on the desired slices.
FROM ubuntu:${UBUNTU_RELEASE} AS builder

ARG ARCH=amd64
ARG CHISEL_VERSION=0.9.1

ADD "https://github.com/canonical/chisel/releases/download/v${CHISEL_VERSION}/chisel_v${CHISEL_VERSION}_linux_${ARCH}.tar.gz" chisel.tar.gz
RUN tar -xvf chisel.tar.gz -C /usr/bin/

WORKDIR /rootfs
COPY . .

# RUN chisel cut --release ubuntu-$UBUNTU_RELEASE --root /rootfs \
RUN chisel cut --release ./ --root /rootfs \
  base-files_base \
  base-files_release-info \
  ca-certificates_data \
  libgcc-s1_libs \
  libc6_libs \
  libssl3_libs \
  # https://github.com/canonical/chiselled-python/blob/main/python3.11/rockcraft.yaml
  openssl_config \
  python3.11_standard

FROM scratch

COPY --from=builder /rootfs /

ENTRYPOINT [ "python", "--version"]
```

```sh
docker build -t test .
docker run test
> Python 3.11.0rc1
```

## Checklist

### DIFF
--- a/slices/python3.11-minimal.yaml
+++ b/slices/python3.11-minimal.yaml
@@ -9,6 +9,8 @@ slices:
       - zlib1g_libs
     contents:
       /usr/bin/python3.11:
+      /usr/bin/python3: {symlink: /usr/bin/python3.11}
+      /usr/bin/python: {symlink: /usr/bin/python3}
       # The next two directories are created to mimic the behaviour in
       # the "postinst" script.
       /usr/local/lib/python3.11/: {make: true, mode: 02775}


rebornplusplus:
This is indeed nice and required. But the problem is that chisel doesn't yet have the notion of conflicting paths. Say you would also like to point the `python` symlink to `python3.12` binary when installing the python3.12 slices. Chisel won't support it yet, since you are describing conflicting paths -- one which points to `python3.11` and another to 3.12. 

There are plans to support the conflicting paths. Meanwhile, of course, we can say that the default python version for a release is `python3.x` and we can then point the `python` symlink to `python3.x` binary. But that will require the discussions of which version should be default.

I will keep it blocked for now. @cjdcordeiro do you wanna add anything more?

bryantbiggs:
yes, its a fair point. I was thinking of how this might cause conflicts but dismissed them because, what use cases are there where you want to use only the bare minimum (i.e. - the point in which someone reached for chisel), and yet you want to also install multiple python versions? yes, it could happen, but it seems like an anti-pattern of what chisel is trying to offer and a niche edge case. but I will let you all deliberate and decide, thank you!

PR #142, opened 2024-02-22
feat(24.04): add openssh-{client,server,sftp-server} and its dependencies

Meulengracht:
First PR of a longer chain to add all dependencies needed to create a rootfs for core bases, as a part of switching to Chisel for building the rootfs for UC26.

This has been tested as of 22/02 and works.

PR #219, opened 2024-04-09
chore(lint): increase line-length max value

rebornplusplus:
~~`line-length` errors are now warnings.~~

This PR increases the maximum length of a line in the yaml files.

Relevant discussions at https://github.com/canonical/chisel-releases/pull/142.


cjdcordeiro (changes requested):
I think this is dangerous.

You are opening the doors for contributors to start writing huge lines, including in places where such shouldn't happen (like `contents`).

This seems to be driven by a problem in a mutation script in #142 . For that, I'd rather:
 - just fix the mutation script. I get that it is a copy&paste, but if that copy&paste is "broken" that's not an excuse for us to accept it like that.
     -  if the issue however is that the YAML format is forcing an indentation that is causing that copy&paste to be longer than acceptable, then let's just increase the `max` value
 - another option is fiddling with the yamllint configuration. I'm not sure if it is possible, but ideally, we could be permissive about **some** blocks of yaml. E.g. you could allow the "level: warning", but only for the "mutation" blocks. Again, not sure if that's possible
     - or, don't lint `#` comments

Meulengracht:
I think the issue in my PR is that the linter does not account for the indentation made before the contents itself. The length of the contents itself does not exceed the 80 characters.

cjdcordeiro:
> I think the issue in my PR is that the linter does not account for the indentation made before the contents itself. The length of the contents itself does not exceed the 80 characters.

That's good to hear, as it makes me more confident about simply increasing the `max` value to accommodate the expected indentations (if no cleaner yamllint-specific remediation is available)

rebornplusplus:
Updated the PR. Set the limit to 100 to account for any indentation spaces.

linostar (approved):
LGTM.

PR #226, opened 2024-04-22
feat(22.04): add slice to generate DB

rebornplusplus:
This PR adds a new slice ``base-files_chisel`` which does not extract anything from the package, rather comes with an artificial path where the Chisel DB should be generated.

This PR is the first one to introduce the "generate" keyword to paths. The "generate" keyword is designed to _generate_ various bookkeeping stuff for a chiselled file system. Right now, it only supports one value: "manifest". Mark a path with "generate: manifest" to generate a Chisel DB at that location. Please note that the marking path must be in the following format:

    /absolute/path/to/dir/**

The path must be an absolute directory path with no wildcard in it's name followed by trailing ** which makes sure nothing else is generated in that directory.

## Additional Context
Won't work until chisel supports the `generate` keyword.


cjdcordeiro (approved):
looking good now ! Thanks!

I've updated a comment but that's it.

niemeyer (approved):
Looks good!

PR #228, opened 2024-04-30
feat(24.04): add 'copyright' slice to all SDFs

cjdcordeiro:
# Proposed changes

A new `copyright` slice is being added to all Slice Definitions Files, for packages that actually install a copyright file.
In addition, this copyright file now becomes a global dependency (`essential`) for all slices within the SDF, via the top-level `essential` field.

## Related issues/PRs

https://github.com/canonical/chisel/pull/127

## Checklist

* [x] I have read the [contributing guidelines](
https://github.com/canonical/chisel-releases/blob/main/CONTRIBUTING.md)
* [x] I have already submitted the [CLA form](
https://ubuntu.com/legal/contributors/agreement)

PR #232, opened 2024-05-01
ci: check if copyright files are installed during tests

cjdcordeiro:
# Proposed changes

 - add a .gitignore file to ignore pycache
 - let the CI also test that the copyright files are installed properly on every slice installation, if the copyright files exist in the deb

## Related issues/PRs

#228 

## Tests

Tested in https://github.com/cjdcordeiro/chisel-releases/actions/runs/8912454734/job/24475940490?pr=10:
 - passes when the copyright exists and is installed
 - passes when the copyright is not installed, but also doesn't exist
 - fails when the copyright isn't installed but should be (`readline-common` in this example)

## Checklist


.github/scripts/install_slices/install_slices.py
```diff
@@ -280,6 +293,49 @@ def install_slice(slice: str, arch: str, release: str) -> None:
                 res.stderr.rstrip(),
             )
             sys.exit(res.returncode)
+        # Check if the copyright file has been installed with this slice
+        if list(
```
letFunny, line 297: [Nit]: Can we use `any(...)` instead?

Another nit is that it might be easier to read if:
```python
copyright_present = any(filter(
                lambda path: f"/usr/share/doc/{pkg}/copyright" in str(path)
                and path.is_file(),
                pathlib.Path(tmpfs).rglob("*"),
            ))
missing_copyright[slice_name] = !copyright_present
```
but this comes down to personal opinion. If we end up doing this, seeing the logic here and in the main function, it might be easier to use `has_copyright` instead an invert the logic.

.github/scripts/install_slices/install_slices.py
```diff
@@ -280,6 +293,49 @@ def install_slice(slice: str, arch: str, release: str) -> None:
...
+    the installation (at least not without duplicating a some of the Chisel
+    codebase).
+
+    TODO: update this function once the Chisel DB is available, as the pkg
+    SHAs will be available from the DB itself.
+    """
+    pkg_shas = os.popen(
+        f"grep -A 25 -R 'Package: {pkg}$' {str(CHISEL_PKG_CACHE)} | grep SHA256 | cut -d ' ' -f 2"
+    ).read()
+    for deb in pkg_shas.splitlines():
```
letFunny, line 324: If I understand this correctly, this might be problematic if we have different architectures and different versions of the same package. Maybe as a good-enough solution for now, we could remove the cache or start a new container for each architecture / version test (if we are not doing that already).

.github/scripts/install_slices/install_slices.py
```diff
@@ -328,11 +389,21 @@ def main() -> None:
...
                 install_slice(
-                    full_slice_name(pkg.package, slice),
+                    pkg.package,
+                    slice,
                     cli_args.arch,
                     cli_args.release,
+                    missing_copyright,
                 )
 
+            if any(missing_copyright.values()):
```
letFunny, line 399: Shouldn't the check here be `if !missing_copyright[full_slice_name(pkg.package, slice)]:`.

.github/scripts/install_slices/test_install_slices.py
```diff
@@ -214,14 +216,46 @@ def test_install_slice(self):
...
+                "foo123", "bar", "amd64", "ubuntu-22.04", mock_missing_copyright
+            )
             assert False
         except SystemExit as e:
             self.assertEqual(e.code, 1)
 
+    @unittest.mock.patch("os.popen")
+    @unittest.mock.patch("pathlib.Path.is_file")
+    @unittest.mock.patch("apt.debfile.DebPackage.__new__")
+    def test_deb_has_copyright_file(self, mock_debpackage, mock_is_file, mock_popen):
```
letFunny, line 234: Thanks for these! Once we land chisel.db we can hopefully simplify these even further.

letFunny (commented):
Thanks for these! I just have a couple of questions.

.github/scripts/install_slices/install_slices.py
cjdcordeiro: That's not a problem because this is running in a matrix, with the `arch` and `release` as inputs. So you'll get an executor for every combination of those, and thus a clean worker/cache

.github/scripts/install_slices/install_slices.py
letFunny: In fact, now that I re-read this, couldn't we use [os.path.isfile](https://docs.python.org/3/library/os.path.html#os.path.isfile) directly to avoid enumerating all of the files? Or is the copyright sometimes named differently?

cjdcordeiro: done in 906a61e

cjdcordeiro: > In fact, now that I re-read this, couldn't we use [os.path.isfile](https://docs.python.org/3/library/os.path.html#os.path.isfile) directly to avoid enumerating all of the files? Or is the copyright sometimes named differently?

hum...sounds about right. I'll update it

cjdcordeiro: done in https://github.com/canonical/chisel-releases/pull/232/commits/620487cdff3c94f8e38ceb5b0037db4d56fa6c14

.github/scripts/install_slices/install_slices.py
cjdcordeiro: this was actually too indented. We just need to check per pkg. 
This has been updated since then.

.github/scripts/install_slices/test_install_slices.py
cjdcordeiro: :crossed_fingers:

.github/scripts/install-slices/install_slices.py
```diff
@@ -328,10 +382,25 @@ def main() -> None:
...
+                    has_copyright,
+                )
+
+        if not all(has_copyright.values()):
+            # Does the copyright file exist in the deb?
+            if deb_has_copyright_file(pkg.package):
+                err = "{} has a copyright file but it wasn't installed with: {}".format(
+                    pkg.package,
+                    ",".join(
+                        list(
```
letFunny, line 398: [Nit]: I think you don't need to create an in-memory list, `join` can consume the iterator directly.

.github/scripts/install-slices/install_slices.py
```diff
@@ -280,6 +293,42 @@ def install_slice(slice: str, arch: str, release: str) -> None:
                 res.stderr.rstrip(),
             )
             sys.exit(res.returncode)
+        # Check if the copyright file has been installed with this slice
+        has_copyright[slice_name] = pathlib.Path(
```
letFunny, line 297: I think we could change the name to `installed_copyright` because when reading this I thought it was talking about the upstream package having copyright.

letFunny (commented):
A couple of more nits :P

.github/scripts/install-slices/install_slices.py
cjdcordeiro: sure -> 673670e

.github/scripts/install-slices/install_slices.py
cjdcordeiro: done 673670e

.github/scripts/install-slices/install_slices.py
```diff
@@ -280,6 +293,43 @@ def install_slice(slice: str, arch: str, release: str) -> None:
...
+    the installation (at least not without duplicating a some of the Chisel
+    codebase).
+
+    TODO: update this function once the Chisel DB is available, as the pkg
+    SHAs will be available from the DB itself.
+    """
+    pkg_shas = os.popen(
+        f"grep -A 25 -R 'Package: {pkg}$' {str(CHISEL_PKG_CACHE)} | grep SHA256 | cut -d ' ' -f 2"
+    ).read()
+    for deb in pkg_shas.splitlines():
```
rebornplusplus, line 318: What do you think about iterating over all the files and checking (a) if the file is a Debian package and (b) if so, if it is the package we are looking for?

I think this `grep` here might slow the script down a bit. For example, this command took 33 seconds in my local machine:

```
time grep -A 25 -R "Package: base-files" ~/.cache/chisel/sha256  | grep SHA256 | cut -d ' ' -f 2
...
real	0m33.735s
user	0m7.389s
sys	0m7.526s
```

cjdcordeiro:
**note** : the 24.04 CI errors are being fixed in https://github.com/canonical/chisel-releases/pull/228/commits/a15e6d81e874e0deef28076f9d47a900d18fdb2c

rebornplusplus (commented):
Looks good to me. Left only a comment wondering what would be the better way.

.github/scripts/install-slices/install_slices.py
letFunny: I think this might be less of a problem if the cache is almost empty which might be the case for the CI (not sure though).

.github/scripts/install-slices/install_slices.py
```diff
@@ -319,6 +369,11 @@ def main() -> None:
         return
     # Install the slices in each package.
     for pkg in packages:
+        # Keep track of whether the copyright file is installed on every "cut"
+        # This should always be the case, whether enforced by a global "essential"
+        # or Chisel itself. Exception: the copyright file will not be installed
+        # if it doesn't exist in the deb itself.
+        installed_copyright = {}
```
letFunny, line 376: Sorry for not finding this in earlier reviews but my Python was a bit rusty, shouldn't `installed_copyright` be a set? This way we can check membership using `in` and check for emptyness below instead of `not all(installed_copyright.values())`.

letFunny (commented):
Looking good apart from the long times (tracking in [comment](https://github.com/canonical/chisel-releases/pull/232/files#r1587451137)).

.github/scripts/install-slices/install_slices.py
cjdcordeiro: should be improved by a factor of ~2x in 3f1fbac

.github/scripts/install-slices/install_slices.py
cjdcordeiro: could, but then you wouldn't been able to identify which slices failed to install it. The CI installs slice by slice (not all in one go). So having it as a dict allows us to have the log message below, saying which slices are the culprit

.github/scripts/install-slices/install_slices.py
```diff
@@ -280,6 +294,44 @@ def install_slice(slice: str, arch: str, release: str) -> None:
...
+        if sha_type and "debian.binary-package" in sha_type:
+            deb_path = str(pathlib.Path(CHISEL_PKG_CACHE / sha_file))
+            sha_pkg = os.popen(
+                f"dpkg-deb -f {deb_path} Package"
+            ).read().strip()
+
+            if sha_pkg == pkg:
+                deb = DebPackage(deb_path)
+                if f"usr/share/doc/{pkg}/copyright" in deb.filelist:
+                    return True
```
letFunny, line 332: [Nit]: `return f"usr/share/doc/{pkg}/copyright" in deb.filelist`.

.github/scripts/install-slices/install_slices.py
letFunny: You are right! I was thinking a set of slices that did not install copyright (sorry to go back there again), then we can check for len() > 0 and all of the keys are faulty and can be reported.

letFunny (approved):
Thanks! Left a couple of comments, it is almost ready

.github/scripts/install-slices/install_slices.py
cjdcordeiro: see 0bf534c

.github/scripts/install-slices/install_slices.py
cjdcordeiro: also done in 0bf534c

rebornplusplus (approved):
LGTM.

letFunny (approved):
LGTM

PR #255, opened 2024-05-30
feat: add python3.12-venv to 24.04

dulmandakh:
# Proposed changes

This PR adds python3.12-venv to Ubuntu. 24.04. This package helps with creating Python virtual environments.

## Testing

```bash
# Example:
#  chisel cut ... --root <path> <pkg>_<proposed-slice>
#  sudo chroot <path> <app>
```

## Checklist


slices/python3.12-venv.yaml
```diff
@@ -0,0 +1,10 @@
+package: python3.12-venv
+
+slices:
+  libs:
+    essential:
+      - python3.12_standard
+      - python3-pip-whl_libs
+      - python3-setuptools-whl_libs
+    contents:
+      /usr/lib/python3.12/ensurepip/*.py:
```
rebornplusplus, line 10: Please include the copyright slice below as well. Just list the symlink as it comes. That should do it.

slices/python3.12-venv.yaml
```diff
@@ -0,0 +1,10 @@
+package: python3.12-venv
+
+slices:
+  libs:
+    essential:
+      - python3.12_standard
```
rebornplusplus, line 6: Will it suffice with `python3.12_core`? If not, what's missing?

rebornplusplus (changes requested):
Maybe you should include `python3-venv` package as well.

Note that `libpython3.12-stdlib` contains `venv` modules already. And this is a snippet from the `ensurepip` module:

```py
def _ensurepip_is_disabled_in_debian_for_system():
    # Detect if ensurepip is being executed inside of a python-virtualenv
    # environment and return early if so.
    if hasattr(sys, 'real_prefix'):
        return

    # Detect if ensurepip is being executed inside of a stdlib venv
    # environment and return early if so.
    if sys.prefix != getattr(sys, "base_prefix", sys.prefix):
        return

    # If we've gotten here, then we are running inside of the system Python
    # and we don't want to use ensurepip to install into the system Python
    # so instead we'll redirect the user to using dpkg and apt-get.
    print('''\
ensurepip is disabled in Debian/Ubuntu for the system python.

Python modules for the system python are usually handled by dpkg and apt-get.

    apt install python3-<module name>

Install the python3-pip package to use pip itself.  Using pip together
with the system python might have unexpected results for any system installed
module, so use it on your own risk, or make sure to only use it in virtual
environments.
''')
    sys.exit(1)
```

slices/python3-pip-whl.yaml
```diff
@@ -0,0 +1,15 @@
+package: python3-pip-whl
+
+essential:
+  - python3-pip-whl_copyright
+
+slices:
+  libs:
```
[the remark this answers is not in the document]
dulmandakh: me too, but there was no such case. I'll rename it.

slices/python3.12-venv.yaml
dulmandakh: there was no copyright file in this package. I use dpkg-query -L to see file list of installed files.

dulmandakh: please see https://packages.ubuntu.com/noble/amd64/python3.12-venv/filelist

slices/python3.12-venv.yaml
dulmandakh: I tried with **python3.12_core**, and got below error

```sh
# python3.12 -m venv .venv
/usr/bin/python3.12: No module named venv
```

Then found that venv module is in distribution slice, and added it.
```sh
# python3.12 -m venv .venv
Error: Command '['/.venv/bin/python3.12', '-m', 'ensurepip', '--upgrade', '--default-pip']' returned non-zero exit status 1.
```

Also added internet, net-data slices but still got errors. Then got thinking that most packages on pypi expect to have standard python, but not having it might cause a lot of issues and hours of debugging.

dulmandakh:
@rebornplusplus I found it surprising that there is no python3 slices, but python3.12. So my expectation was the same, only provide python3.12-venv package.

dulmandakh:
I'll create the same slices for Ubuntu 22.04 if this PR merged.

slices/python3.12-venv.yaml
```diff
@@ -0,0 +1,10 @@
+package: python3.12-venv
+
+slices:
+  libs:
```
cjdcordeiro, line 4:
```suggestion
  ensurepip:
```

cjdcordeiro (changes requested):

slices/python3.12-venv.yaml
cjdcordeiro: Well, it has `./usr/share/doc/python3.12-venv -> python3.12` which points to python3.12's copyright. So you can do something like 

```yaml
/usr/share/doc/python3.12-venv: {symlink: /usr/share/doc/python3.12/copyright}
# or even
# /usr/share/doc/python3.12-venv:
# but that could bring additional file that we don't need
```

cjdcordeiro (changes requested):

slices/python3.12-venv.yaml
cjdcordeiro: P.S. you need to have it on the dedicated `copyright` slice

cjdcordeiro (changes requested):
just need to fix a linting issue and should be good to go!

P.S. the CLA check is failing. Have you signed the CLA?

dulmandakh:
Yep, I contributed to Canonical Kubernetes.

cjdcordeiro:
> Yep, I contributed to Canonical Kubernetes.

:+1: ty

Just the linting then

cjdcordeiro (approved):
thanks for the change. lgtm. I'll wait for a 2nd approval

linostar (approved):
LGTM.

PR #267, opened 2024-06-24
Rocks 994 ca cert mutation scripts 22.04

zhijie-yang:
# Proposed changes

The script `update-ca-certificates` and its corresponding config file `/etc/ca-certificates.conf` is added to the SDF of `ca-certificates`.

## Related issues/PRs
ROCKS-994

### Forward porting

#266 

## Checklist

* [x] I have read the [contributing guidelines](
https://github.com/canonical/chisel-releases/blob/main/CONTRIBUTING.md)
* [x] I have already submitted the [CLA form](
https://ubuntu.com/legal/contributors/agreement)

PR #266, opened 2024-06-24
Rocks 994 ca cert mutation scripts 24.04

zhijie-yang:
# Proposed changes

The script `update-ca-certificates` and its corresponding config file `/etc/ca-certificates.conf` is added to the SDF of `ca-certificates`.

## Related issues/PRs
ROCKS-994

### Forward porting

Not applicable.

## Checklist

* [x] I have read the [contributing guidelines](
https://github.com/canonical/chisel-releases/blob/main/CONTRIBUTING.md)
* [x] I have already submitted the [CLA form](
https://ubuntu.com/legal/contributors/agreement)

PR #268, opened 2024-06-24
Rocks 994 ca cert mutation scripts 20.04

zhijie-yang:
# Proposed changes

The script `update-ca-certificates` and its corresponding config file `/etc/ca-certificates.conf` is added to the SDF of `ca-certificates`.

## Related issues/PRs
ROCKS-994

### Forward porting

#266 
#267 

## Checklist


slices/ca-certificates.yaml
```diff
@@ -15,6 +16,17 @@ slices:
         content.read(certs_dir + path) for path in content.list(certs_dir)
       ]
       content.write("/etc/ssl/certs/ca-certificates.crt", "".join(certs))
+      names = [
+        "mozilla/{}".format(name) for name in content.list(certs_dir)
```
linostar, line 20: doesn't that mean you're using the paths /usr/share/ca-cerificates/mozilla/*.crt?
However by having `until: mutate` for:
```      /usr/share/ca-certificates/mozilla/*: {until: mutate}```
it means you're deleting those paths

File: slices/ca-certificates.yaml [hunk 2]
```diff
@@ -15,6 +16,17 @@ slices:
...
+      ]
+      content.write("/etc/ca-certificates.conf", "\n".join(names))
+
+  tools:
```
linostar, line 24: If there is a liberty for the slice naming, I'd suggest something more correspondant to the contents, like `full` or `fulldata`.

linostar (changes requested):

slices/ca-certificates.yaml
zhijie-yang: Yeah, since we are having the bundle at `/etc/ssl/certs/ca-certificates.crt`, and the symlink `/usr/lib/ssl/certs.pem -> /etc/ssl/certs/ca-certificates.crt`, it is possible for SSL clients to use the bundle rather than these single certificates for authentication. Since we are not able to run `openssl rehash` or `c_hash` during slicing, symlinking `/etc/ssl/certs/* -> /usr/share/ca-certificates/mozilla/*` is generally useless, as SSL clients don't use the `*.pem` certificates in `capath = /etc/ssl/certs` without them being rehashed.

slices/ca-certificates.yaml
zhijie-yang: > Yeah, since we are having the bundle at `/etc/ssl/certs/ca-certificates.crt`, and the symlink `/usr/lib/ssl/certs.pem -> /etc/ssl/certs/ca-certificates.crt`, it is possible for SSL clients to use the bundle rather than these single certificates for authentication. Since we are not able to run `openssl rehash` or `c_hash` during slicing, symlinking `/etc/ssl/certs/* -> /usr/share/ca-certificates/mozilla/*` is generally useless, as SSL clients don't use the `*.pem` certificates in `capath = /etc/ssl/certs` without them being rehashed.

With this, I don't think it would be necessary to call it `full` or `fulldata`.

slices/ca-certificates.yaml
linostar: I understand. But do we need to list certificates in ` /usr/share/ca-cerificates/mozilla/` inside `/etc/ca-certificates.conf` if the dir ` /usr/share/ca-cerificates/mozilla/` doesn't exist on system?

cjdcordeiro (changes requested):
same as https://github.com/canonical/chisel-releases/pull/267#pullrequestreview-2180449349

slices/ca-certificates.yaml
```diff
@@ -4,18 +4,56 @@ essential:
...
       content.write("/etc/ssl/certs/ca-certificates.crt", "".join(certs))
 
+  bins:
+    essential:
+      - openssl_bins
+      # The `/usr/sbin/update-ca-certificates` script requires sed to run.
+      - sed_bins
+    contents:
+      # To run update-ca-certificates without breaking the default certificates
+      # at /etc/ssl/certs/ca-certificates.crt, the `_data-with-certs` slice should
```
cjdcordeiro, line 53: this slice doesn't exist

cjdcordeiro (changes requested):

slices/ca-certificates.yaml
cjdcordeiro: ref: https://github.com/canonical/chisel-releases/issues/285

zhijie-yang: Fixed.

cjdcordeiro (approved):

linostar (approved):
LGTM, Thanks!

PR #274, opened 2024-07-05
feat(24.04): systemd and dependencies

Meulengracht:
# Proposed changes

Add systemd slice and it's dependencies

## Testing

Smoke testing mount

```bash
mkdir mount-root
chisel cut --root mount-root bash_bins coreutils_bins passwd_config base-files_base mount_bins
sudo chroot mount-root/
mkdir /proc
mount -t proc proc /proc
umount /proc
```

Smoke testing systemd

```bash
mkdir systemd-root
chisel cut --root systemd-root bash_bins coreutils_bins passwd_config base-files_base systemd_bins
sudo chroot systemd-root
mkdir /proc
mount -t proc proc /proc
systemctl enable getty@tty1.service
systemctl disable getty@tty1.service
systemctl preset-all
systemd --help
journalctl --update-catalog
umount /proc
```

## Checklist

* [x] I have read the [contributing guidelines](
https://github.com/canonical/chisel-releases/blob/main/CONTRIBUTING.md)
* [x] I have already submitted the [CLA form](
https://ubuntu.com/legal/contributors/agreement)

PR #276, opened 2024-07-05
feat(24.04): netplan.io and dependencies

Meulengracht:
# Proposed changes

Add netplan and it's dependencies. Netplan is a bit useless without other networking daemons up and running (i.e systemd-networkd), and it's also quite useless without udev running. The app itself is working and the only errors it throws are due to either a chroot'ed environment or because some other system service is missing.

## Testing

```bash
chisel cut bash_bins coreutils_bins passwd_config netplan.io_bins
mkdir -p <root>/proc
mkdir -p <root>/dev
mount --bind /dev <root>/dev
sudo chroot <root>
mount -t proc proc /proc
netplan --help
netplan apply # useless without udev
netplan generate
netplan status # useless without networking daemons
```

## Checklist

## Additional Context
This is based on top of https://github.com/canonical/chisel-releases/pull/274 since netplan requires systemd and mount libraries.


slices/python3.12-minimal.yaml
```diff
@@ -11,6 +11,7 @@ slices:
       - libpython3.12-minimal_libs
       - zlib1g_libs
     contents:
+      /usr/bin/python3: {symlink: /usr/bin/python3.12}
```
Meulengracht, line 14: This probably needs a bit of discussion, should I perhaps not add this and create it manually in the tests? Until we support the slice precedence? @cjdcordeiro

rebornplusplus: Yup, that would be better. We are not installing `python3` paths currently.

tests/spread/integration/iproute2/test.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+
+# smoketest ip commands
+ip --help | grep "Usage:"
+
+# test some basic commands and against loopback that
+# we can kinda expect is there
+ip link | grep "LOOPBACK,UP"
```
cjdcordeiro, line 8: can you please also test the other binaries being installed? I know it might be tedious, but that's the only way we can ensure the slice has all it needs for them to behave properly

tests/spread/integration/netplan.io/test.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
```
cjdcordeiro, line 1: are any of the `scripts` testable?

slices/python3-yaml.yaml
cjdcordeiro, line 1: could you please add some tests for these python-* SDFs?

cjdcordeiro (changes requested):

PR #280, opened 2024-07-12
ci(spread): add reusable workflow for Spread tests

cjdcordeiro:
# Proposed changes

This PR introduces a new reusable GitHub workflow for running Spread tests. 

Each branch will have its own `spread.yaml` configuration, and this workflow will be called from the branches, on every PR. 

The tests shall only run for:
 - the SDFs which are added or modified
 - the tests which are added or modified 

## Testing

Tested the execution of this workflow here:
 - https://github.com/cjdcordeiro/chisel-releases/pull/11 (with additional mock commits to ensure the logic works when filtering the added and modified files)
 - https://github.com/cjdcordeiro/chisel-releases/pull/13

## Checklist

### DIFF
--- a/.github/workflows/spread.yaml
+++ b/.github/workflows/spread.yaml
@@ -1,0 +1,69 @@
+name: Spread
+run-name: Spread for ${{ github.ref }}
+
+on:
+  workflow_call:
+
+jobs:
+  spread-tests:
+    name: Run Spread tests
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          ref: ${{ github.head_ref }}
+          repository: ${{ github.event.pull_request.head.repo.full_name }}
+
+      - name: Check changed paths
+        id: changed-slices
+        uses: dorny/paths-filter@v3
+        with:
+          filters: |
+            slices:
+              - added|modified: 'slices/**/*.yaml'
+          list-files: shell
+      
+      - name: Check changed test directories
+        uses: tj-actions/changed-files@v35
+        id: changed-tests
+        with:
+          separator: " "
+          dir_names: "true"
+          files: |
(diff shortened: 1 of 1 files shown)


letFunny (approved):
Thanks Cris, it is looking very good! From the linked PRs I am missing the use case where the spread tests already exists and changing the SDF triggers the integration test (i.e. `changed-slices` triggering instead of `changed-tests`).

cjdcordeiro:
> Thanks Cris, it is looking very good! From the linked PRs I am missing the use case where the spread tests already exists and changing the SDF triggers the integration test (i.e. `changed-slices` triggering instead of `changed-tests`).

Thanks.

As for that use case, it's this one here: https://github.com/cjdcordeiro/chisel-releases/actions/runs/9905003363/job/27363593631?pr=11#step:8:27

From the logs you can see the SDF being considered as a trigger for that test. It just so happens that the test itself is also a trigger, cause it is also being created for the first time.

.github/workflows/spread.yaml
```diff
@@ -0,0 +1,69 @@
...
+        id: changed-slices
+        uses: dorny/paths-filter@v3
+        with:
+          filters: |
+            slices:
+              - added|modified: 'slices/**/*.yaml'
+          list-files: shell
+      
+      - name: Check changed test directories
+        uses: tj-actions/changed-files@v35
```
rebornplusplus, line 27: Can't we use the `dorny/paths-filter` action from above here too?

rebornplusplus (approved):
Looks good to me, thank you! 

Just one quick question about using less distinct actions.

.github/workflows/spread.yaml
rebornplusplus: Ah okay, thought so.
No worries, approved it already. :+1:

PR #282, opened 2024-07-15
feat: add new slice definitions for 'git' and dependancies

clay-lake:
# Proposed changes
Add slices for git and dependency libcurl3t64-gnutls.

## Testing
Rudimentary smoke testing was provided by the script below. The bins slice successfully executes basic git operations, except for cloning a repository over http, which requires the core slice. Cloning over ssh has not yet been tested.

```bash
set -e

function echo_status(){
        $@ > /dev/null
        echo  $@ " " $? 
}

cd figlet-local
echo_status git status
echo_status git log
echo_status git config --global user.email "NA"
echo_status git config --global user.name "NA"
echo_status git checkout -b test
echo //new comment > getopt.c
echo_status git commit -a -m test_commit
echo_status git checkout master
echo_status git branch
echo_status git merge test

echo_status cd ~

echo_status mkdir figlet-http
echo_status git clone https://github.com/cmatsuoka/figlet.git figlet-http

echo TESTS PASS
```

All testing was performed on a chiseled (ubuntu-24.04) container containing the following slices.
```
base-files_base 
base-files_release-info 
busybox_bins 
git_bins # or git_core
ca-certificates_data
```

## Checklist


slices/git.yaml
```diff
@@ -0,0 +1,41 @@
...
+      - zlib1g_libs
+      - git_bins
+    contents:
+      /usr/lib/git-core/**:
+
+  contrib:
+    essential:
+      - git_perl5
+    contents:
+      /usr/share/doc/git/contrib/**:
```
rebornplusplus, line 32: These two paths would probably break since the symlink target might not exist:
```
lrwxrwxrwx root/root         0 2024-05-20 11:15 ./usr/share/doc/git/contrib/persistent-https/LICENSE -> ../../../../common-licenses/Apache-2.0
lrwxrwxrwx root/root         0 2024-05-20 11:15 ./usr/share/doc/git/contrib/subtree/COPYING -> ../../../../common-licenses/GPL-2
```

slices/git.yaml
```diff
@@ -0,0 +1,41 @@
...
+      - libc6_libs
+      - libcurl3t64-gnutls_libs
+      - libexpat1_libs
+      - libpcre2-8-0_libs
+      - zlib1g_libs
+      - git_bins
+    contents:
+      /usr/lib/git-core/**:
+
+  contrib:
```
rebornplusplus, line 28: What's the usage of the `contrib` slice?

slices/git.yaml
```diff
@@ -0,0 +1,41 @@
...
+      /usr/bin/scalar:
+
+  core:
+    essential:
+      - libc6_libs
+      - libcurl3t64-gnutls_libs
+      - libexpat1_libs
+      - libpcre2-8-0_libs
+      - zlib1g_libs
+      - git_bins
```
rebornplusplus, line 24: You could omit `libc6`, `libpcre2..` and `zlib1g` items from the list since `git_bins` already has them.

rebornplusplus (changes requested):
Nice! Note that the Lint is failing.

slices/git.yaml
clay-lake: Fixed! Good catch

slices/git.yaml
clay-lake: That is a good point. I thought somehow it was relevant because of the scripts in that directory, but it looks like the source code here needs to be built first. I will exclude it, which also fixes some broken symlinks you pointed out.

slices/git.yaml
```diff
@@ -0,0 +1,34 @@
...
+      - libexpat1_libs
+    contents:
+      /usr/lib/git-core/**:
+
+  perl5:
+    essential:
+      - git_core
+    contents:
+      /usr/share/perl5/Git.pm:
+      /usr/share/perl5/Git/*:
```
rebornplusplus, line 30: I think this is also extra, that we currently don't need. I have no strong objections against having it. But the problem would be removing it in future if we add it now.

slices/git.yaml
```diff
@@ -0,0 +1,34 @@
...
+      /usr/bin/git-shell:
+      /usr/bin/scalar:
+
+  core:
+    essential:
+      - git_bins
+      - libcurl3t64-gnutls_libs
+      - libexpat1_libs
+    contents:
+      /usr/lib/git-core/**:
```
rebornplusplus, line 23: I wonder if we should merge the `bins` and `core` slices into one: `bins`. Since together they are offering binaries and executables. Let me know what you think!

rebornplusplus (commented):
Looking good! I just left two comments. let me know what you think!

slices/git.yaml
clay-lake: I would keep both, there are a few extra bins in core which are not needed for simple tasks. Infact, it might be nice to include a slice that only include the single git binary (/usr/bin/git). We could call it minimal for instance. The use case here would be build envs that need to extract details about the repo.

slices/git.yaml
clay-lake: I agree. done!

slices/git.yaml
rebornplusplus: Alright, sounds good.

rebornplusplus (approved):
Looks good to me, thank you!

slices/git.yaml
```diff
@@ -0,0 +1,27 @@
...
```
cjdcordeiro, line 23: I actually agree with both of you here :)

The thing is that in other SDFs, we use the term `core` to refer to the actual core minimum that is required for the pkg's functionality to work (e.g. `python3.12_core`).

Here, I'm afraid the slice name conveys the opposite, as `bins` is more fundamental than `core`. 

I get that the name is given after the folder name, but what about we compromise and just call it `git-core`?

cjdcordeiro (changes requested):
Great! Thanks.

Are you sure you don't need `liberror-perl` and `perl`? If so, then there's just one small comment and it should be good to go

slices/git.yaml
clay-lake: That is a good call! I'll change that.

cjdcordeiro (changes requested):
let's add some tests too

cjdcordeiro (changes requested):
Thanks for this. The SDFs LGTM.
The test is also nice. I do think it should be simplified a bit though.

I.e.
 - we should always aim to have test that target only the slices we're defining. Otherwise we might unintentionally bring in contents from other packages that end up making the proposed slices work. In this case, I don't think you need `busybox`, at least not for the basic `git` testing. All the directory handling, printing, etc., can be done from outside the chroot.

clay-lake:
@cjdcordeiro That is a good point. Let me see what I can do to refactor the tests.

lengau:
@clay-lake I took a swing at the tests in my PR: https://github.com/canonical/chisel-releases/blob/b3a23c67505afa0bd352927aebd669c40f0ed2bb/tests/spread/integration/git/task.yaml

slices/python3-pyftpdlib.yaml
```diff
@@ -0,0 +1,15 @@
+package: python3-pyftpdlib
```
lengau, line 1: I'm guessing we'll get a request to test this standalone (not just as part of the git tests). Would you like me to write a test on my branch?

lengau (approved):
LGTM, thanks! One question about the pyftpd, but otherwise I'm ready to do a final sync to my branch as soon as we hear a go.

slices/python3-pyftpdlib.yaml
clay-lake: Sure! Thank you.

lengau: [Here ya go!](https://github.com/canonical/chisel-releases/blob/5c73271a1027f3f01bcc55f4bc27fd706638565c/tests/spread/integration/python3-pyftpdlib/task.yaml)

slices/git.yaml
```diff
@@ -6,33 +6,36 @@ essential:
...
+      /usr/bin/git: { symlink: /usr/lib/git-core/git }
       /usr/lib/git-core/git:
 
   daemon:
-    # support for git daemon
     essential:
       - git_bins
     contents:
       /usr/lib/git-core/git-daemon:
+      /usr/lib/git-core/git-upload-pack:
```
cjdcordeiro, line 35: Should this be in the `daemon` slice?

cjdcordeiro (changes requested):
The python test also needs fixing it seems

slices/git.yaml
clay-lake: This seems to be a leftover from a prior state; the current test runs fine without it. Thanks!

clay-lake:
The python tests are working now. It was a mismatch between python versions.

tests/spread/integration/git/task.yaml
```diff
@@ -0,0 +1,157 @@
...
+    chroot_git clone --depth 1 https://git.launchpad.net/ubuntu/+source/hello hello
+    cd $_
+
+    chroot_git --no-pager log
+    chroot_git fetch origin
+    chroot_git --no-pager  branch --remotes --contains $(chroot_git rev-list -n 1 HEAD)
+  }
+
+  function test_ftp_clone(){
+    ftp_root=$(install-slices python3-pyftpdlib_libs)
```
rebornplusplus, line 82: Do you need to install the `python3-pyftpdlib` slices in the tests or should it work by installing with apt too?

tests/spread/integration/git/task.yaml
```diff
@@ -0,0 +1,157 @@
...
+    chroot_git clone git://localhost/daemon-src daemon-dst
+
+    # Cleanup. TERM or KILL cause git to return an en error.
+    kill -INT $(fuser 9418/tcp  2>/dev/null)
+  }
+
+  rootfs=$(install-slices $SLICE $EXTRA_SLICES)
+  setup $rootfs
+  $RUN $rootfs
+  cleanup
```
rebornplusplus, line 157: Nice!

rebornplusplus (changes requested):
Looks nice and small, thanks! I left a few comments below.

slices/git.yaml
```diff
@@ -0,0 +1,56 @@
+package: git
+
+essential:
+  - git_copyright
+
+slices:
+  bins:
```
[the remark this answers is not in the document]
clay-lake: I would not be surprised if `pull` works already, but agreed :)

tests/spread/integration/git/task.yaml
clay-lake: Thank @lengau :smiley:

tests/spread/integration/git/task.yaml
clay-lake: It should work with the version installed via apt as well. I ended up chiseling this package because it was relatively simple to add for testing.

slices/git.yaml
```diff
@@ -0,0 +1,56 @@
...
+      /usr/bin/git: { symlink: /usr/lib/git-core/git }
+      /usr/lib/git-core/git:
+
+  daemon:
+    essential:
+      - git_bins
+    contents:
+      /usr/lib/git-core/git-daemon:
+
+  http-clone:
```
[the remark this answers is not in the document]
clay-lake: Fixed, as discussed I swapped "clone" in these slices for "support"

slices/python3-pyftpdlib.yaml
```diff
@@ -0,0 +1,15 @@
...
+
+essential:
+  - python3-pyftpdlib_copyright
+
+slices:
+  libs:
+    essential:
+      - python3.12_standard
```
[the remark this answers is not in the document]
clay-lake: updated! :partying_face:

cjdcordeiro (approved):
lgtm thanks

cjdcordeiro:
@clay-lake can you pls raise these changes to 24.10 too?

slices/git.yaml
```diff
@@ -0,0 +1,56 @@
...
+    # - log
+    # - merge
+    # - rev-list
+    # - status
+    essential:
+      - libc6_libs
+      - libpcre2-8-0_libs
+      - zlib1g_libs
+    contents:
+      /usr/bin/git: { symlink: /usr/lib/git-core/git }
```
rebornplusplus, line 27:
```suggestion
      # /usr/bin/git is a regular file in the package, but shares
      # the same hash of /usr/lib/git-core/git. That's why this
      # is added as a symlink to reduce space.
      /usr/bin/git: { symlink: /usr/lib/git-core/git }
```

rebornplusplus (approved):
Looks good to me, thank you very much!

clay-lake:
Will do! @cjdcordeiro https://github.com/canonical/chisel-releases/pull/396

@lengau Merge is complete :)
There are few more changes for you to pull before you merge your PR.

PR #293, opened 2024-07-24
feat(24.04): add dbus slices

Meulengracht:
# Proposed changes

Add the dbus and related packages that we need for Ubuntu Core.

## Testing

I have a provided an integration test to verify that dbus is working. The test requires a bit of setup so I'd advise reading the test to get an idea.

## Checklist


spread.yaml
```diff
@@ -17,7 +17,8 @@ backends:
       echo "Allocating $SPREAD_SYSTEM..."
       docker_image=$(echo $SPREAD_SYSTEM | awk -F '-' '{print $1":"$2}')
       docker_arch=$(echo $SPREAD_SYSTEM | awk -F '-' '{print $NF}')
-      docker run --rm -e DEBIAN_FRONTEND=noninteractice \
+      docker run --cap-add SYS_ADMIN --security-opt apparmor=unconfined \
```
cjdcordeiro, line 20: as for the others and previous discussions, we'll need to disallow this. can you test these slices with something more basic?

Meulengracht: I dropped the added capabilities and now just use LXD backend, it passes there

slices/dbus-session-bus-common.yaml
cjdcordeiro, line 1: missing the copyright here

slices/dbus-system-bus-common.yaml
cjdcordeiro, line 1: also missing the copyright here

slices/dbus.yaml
```diff
@@ -0,0 +1,34 @@
...
+      - libsystemd0_libs
+    contents:
+      /etc/init.d/dbus:
+      /usr/lib/dbus-1.0/dbus-daemon-launch-helper:
+
+  config:
+    contents:
+      /etc/default/dbus:
+
+  extras:
```
cjdcordeiro, line 25: hum, maybe `services` instead of `extras` to be more representative of what's being installed

cjdcordeiro (changes requested):

slices/dbus.yaml
Meulengracht: It's also more consistent with previous packages actually, updating this!

cjdcordeiro (approved):
thanks. lgtm

_ps: please try not to force push otherwise the review gets more time-consuming as we'll need to go new commit by new commit instead of just checking the `diff` since the last review_

Meulengracht:
> thanks. lgtm
> 
> _ps: please try not to force push otherwise the review gets more time-consuming as we'll need to go new commit by new commit instead of just checking the `diff` since the last review_

Ah sorry - I think it "force pushes" when I rebase the branch from the web GUI, I am not changing any commits so all commits should still be there

linostar (approved):
Thanks. LGTM!

slices/dbus-daemon.yaml
```diff
@@ -0,0 +1,30 @@
...
+      - libc6_libs
+      - libcap-ng0_libs
+      - libdbus-1-3_libs
+      - libexpat1_libs
+      - libselinux1_libs
+      - libsystemd0_libs
+    contents:
+      /usr/bin/dbus-daemon:
+      /usr/bin/dbus-run-session:
+
```
rebornplusplus, line 23: Should we create the `/var/lib/dbus` directory as well? The following are from `postinst`:

```sh
if [ "$1" = configure ]; then
    # This is idempotent, so it's OK to do every time. The system bus' init
    # script does this anyway, but you also have to do this before a session
    # bus will work on non-systemd systems, so we do this here for the
    # benefit of people starting a temporary session bus in a chroot.
    mkdir -p /var/lib/dbus
    dbus-uuidgen --ensure                                                                                               
fi
```

slices/dbus-system-bus-common.yaml
```diff
@@ -0,0 +1,23 @@
+package: dbus-system-bus-common
+
+essential:
+  - dbus-system-bus-common_copyright
+
+slices:
+  # Depends on adduser, however it's used by the dbus postinst
+  # to create the following user:
+  # 'messagebus' with home '/var/run/dbus' and disabled password
```
rebornplusplus, line 9: I thought the user did not have any home dirs:

```sh
    in_sysroot adduser --system \
            --quiet \
            --home /nonexistent \                                                                                       
            --no-create-home \
            --disabled-password \
            --group "$MESSAGEUSER"
```

slices/dbus.yaml
```diff
@@ -0,0 +1,35 @@
...
+      - dbus-bin_bins
+      - dbus-daemon_bins
+      - dbus-system-bus-common_config
+      - dbus_config
+      - libc6_libs
+      - libdbus-1-3_libs
+      - libexpat1_libs
+      - libsystemd0_libs
+    contents:
+      /etc/init.d/dbus:
```
rebornplusplus, line 18: This should be in `config`.

slices/dbus.yaml
```diff
@@ -0,0 +1,35 @@
...
+      /usr/lib/dbus-1.0/dbus-daemon-launch-helper:
```
rebornplusplus, line 19: I wonder if we should omit the `1.0` part.
```suggestion
      /usr/lib/dbus-*/dbus-daemon-launch-helper:
```

rebornplusplus (commented):
The slices look nice! Only have a few comments below.

slices/dbus-system-bus-common.yaml
Meulengracht: I was fooled by the `MESSAGEHOME` variable that seems unused :-) corrected it, thanks!

slices/dbus.yaml
Meulengracht: Are we sure? It's a binary

rebornplusplus: Ah, okay. I did not realise this was an executable.

rebornplusplus (approved):
Looks good to me, thank you!

PR #304, opened 2024-08-02
feat(24.04): add secureboot-db and sbsigntool

Meulengracht:
# Proposed changes

These are needed by Ubuntu Core to handle UEFI db and dbx updates.

## Checklist


cjdcordeiro (changes requested):
ty - the `sbsigntool` seems like it could be tested

tests/spread/integration/sbsigntool/MOK.der
cjdcordeiro, line 1: I'm being a bit cautious of size since Chisel currently pulls these branches when cutting slices. Is there a way these assets could be generated during the tests? or at least made smaller?

cjdcordeiro (changes requested):

tests/spread/integration/sbsigntool/MOK.der
Meulengracht: It's less than 1KB, I don't think there is a need for it to be smaller. These are the default certificates/keys that come with the Ubuntu 24.04 dist.

Meulengracht: Mmm, maybe I was wrong, maybe they are generated on install - let me just see if I can replicate the commands

Meulengracht: Turned out to be pretty easy - the way I initially did it was way incredibly complicated - I amended it in the PR

tests/spread/integration/sbsigntool/task.yaml
```diff
@@ -0,0 +1,18 @@
+summary: Integration tests for sbsigntool
+
+systems:
+  - -ubuntu-24.04-ppc64le
+  - -ubuntu-24.04-s390x
```
cjdcordeiro, line 5: nice

slices/secureboot-db.yaml
```diff
@@ -0,0 +1,19 @@
+package: secureboot-db
+
+essential:
+  - secureboot-db_copyright
+
+slices:
+  dist:
```
cjdcordeiro, line 7: wouldn't `data` be a more appropriate name? (since we already have other slices using a similar convention)

slices/secureboot-db.yaml
```diff
@@ -0,0 +1,19 @@
...
+    essential:
+      - sbsigntool_bins
```
cjdcordeiro, line 9: afaik this is a binary data file, so why would it need this dependency?

slices/secureboot-db.yaml
```diff
@@ -0,0 +1,19 @@
...
+    contents:
+      /usr/share/secureboot/updates/dbx/dbxupdate_*.bin:
+
+  extras:
```
cjdcordeiro, line 13:
```suggestion
  services:
```

cjdcordeiro (changes requested):

slices/secureboot-db.yaml
Meulengracht: In theory it doesn't have to - but the secureboot-db package pulls in sbsigntools, and it was to keep this dependency

cjdcordeiro: yeah but I think it's actually the `services` slice that needs that, cause the service will do:

```
ExecStart=/usr/bin/sbkeysync --no-default-keystores --keystore /usr/share/secureboot/updates --verbose
```

Meulengracht: Right good point - i'll move the dep to that slice

cjdcordeiro (approved):
nice thanks

PR #306, opened 2024-08-02
feat(24.04): add iptables, sudo and add mutation script for pam-auth-update

Meulengracht:
# Proposed changes

Iptables is something we include in Ubuntu Core. Unfortunately to test iptables I need both sudo (which I then figured wasn't working due to bad libpam support), which resulted in a large mutation script that emulates what pam-auth-update does.

There are dedicated integration tests for the libpam generation and sudo, and the test for iptables currently is pretty shallow since it needs kernel modules loaded that aren't available.

## Checklist


tests/spread/integration/iptables/task.yaml
```diff
@@ -0,0 +1,13 @@
...
+
+execute: |
+  rootfs="$(install-slices passwd_config libpam-runtime_config netbase_default-hosts sudo_bins iptables_bins)"
+  host=$(hostname)
+  
+  # setup /etc/hosts for the current hostname, otherwise
+  # sudo will be confused
+  sed -i "/127.0.0.1\tlocalhost/a 127.0.0.1\t$host" "$rootfs"/etc/hosts
+
+  ls /nono
```
Meulengracht, line 11: this test needs more work as mentioned in the PR description, will finish it after the weekend

cjdcordeiro:
@Meulengracht following up on this - it'd be nice to try these tests with LXD as well, once we get sort out why you're having those issues in #318

Meulengracht:
> @Meulengracht following up on this - it'd be nice to try these tests with LXD as well, once we get sort out why you're having those issues in #318

It passes on the LXD backend locally, I removed the caps and restored the tests

Meulengracht:
@cjdcordeiro @rebornplusplus this is ready for review

slices/iptables.yaml
```diff
@@ -0,0 +1,204 @@
...
+      /usr/sbin/iptables-restore-translate:
+      /usr/sbin/iptables-translate:
+      /usr/sbin/nfnl_osf:
+      /usr/sbin/xtables-legacy-multi:
+      /usr/sbin/xtables-monitor:
+      /usr/sbin/xtables-nft-multi:
+
+  # The xlst is used to convert XML configuration into something
+  # iptables can understand, and vice-versa.
+  extras:
```
cjdcordeiro, line 54:
```suggestion
  converters:
```

or something else a bit more representative

slices/iptables.yaml
```diff
@@ -0,0 +1,204 @@
...
+      /usr/sbin/ip6tables-legacy:
+      /usr/sbin/ip6tables-legacy-restore:
+      /usr/sbin/ip6tables-legacy-save:
+      /usr/sbin/ip6tables-nft:
+      /usr/sbin/ip6tables-nft-restore:
+      /usr/sbin/ip6tables-nft-save:
+      /usr/sbin/ip6tables-restore-translate:
+      /usr/sbin/ip6tables-translate:
+      /usr/sbin/iptables-apply:
+      /usr/sbin/iptables-legacy:
```
cjdcordeiro, line 39: it is acceptable to do this later. but if it makes sense to your use case, it could also be done now. And that is:
 - is there a way we could split the libs and bins slices further? I.e., right now, when installing `bins`, you pretty much get the whole pkg. However, it seems like some `bins` could be nicely grouped by the type of functionality/domain they operate on. For example, we can have slices for `ip6tables`, `iptables`, `legacy`, etc. Would that make sense? (again, if you need them all today, then we can still create the splits later)

slices/libpam-runtime.yaml
```diff
@@ -18,41 +18,123 @@ slices:
...
+      /etc/pam.d/common-session: {text: '', mutable: true}
+      /etc/pam.d/common-session-noninteractive: {text: '', mutable: true}
       /etc/pam.d/other:
+      /usr/share/pam-configs/unix: { until: mutate }
+      /usr/share/pam/common-account: { until: mutate }
+      /usr/share/pam/common-auth: { until: mutate }
+      /usr/share/pam/common-password: { until: mutate }
+      /usr/share/pam/common-session: { until: mutate }
+      /usr/share/pam/common-session-noninteractive: { until: mutate }
+    mutate: |
```
cjdcordeiro, line 42: nice

slices/libpam-runtime.yaml
```diff
@@ -18,41 +18,123 @@ slices:
...
-      /usr/share/pam/common-account.md5sums:
-      /usr/share/pam/common-auth:
-      /usr/share/pam/common-auth.md5sums:
-      /usr/share/pam/common-password:
-      /usr/share/pam/common-password.md5sums:
-      /usr/share/pam/common-session:
-      /usr/share/pam/common-session-noninteractive:
-      /usr/share/pam/common-session-noninteractive.md5sums:
-      /usr/share/pam/common-session.md5sums:
-
```
cjdcordeiro, line 55: we cannot delete slices, otherwise it will be a regression for whoever is already using them. I suggest merging this with the mutation work you did above

slices/sudo.yaml
```diff
@@ -0,0 +1,54 @@
...
+  config:
+    contents:
+      /etc/pam.d/sudo:
+      /etc/pam.d/sudo-i:
+      /etc/sudo.conf:
+      /etc/sudo_logsrvd.conf:
+      /etc/sudoers:
+      /usr/lib/tmpfiles.d/sudo.conf:
+
+  extras:
```
cjdcordeiro, line 38:
```suggestion
  services:
```

cjdcordeiro (changes requested):

slices/libpam-runtime.yaml
Meulengracht: But for this specific case it wouldn't make sense for someone to use these? These are used as templates or input strictly by the scripts only to generate the actual files needed - it does not make sense for these to be in the end-result.

Meulengracht: I can reduce them to a single `essential` that depends on config if that really is better, but I don't like that it needs to clutter for slices that were essentially "useless" for regular usage

slices/iptables.yaml
Meulengracht: For our use case today we need all of them, if we were to reduce this it would require us to understand what the consequences would be of not including this. Unfortunately we have customers that tend to use whatever they please so it's really hard for us to just drop these kinds of tools

slices/libpam-runtime.yaml
cjdcordeiro: the slice can be modified. that's ok. but you can imagine that existing users who may have pipelines running `chisel cut ..._pam-defaults` would now have the pipeline suddenly fail for what is supposed to be a stable release. That's a regression, regardless of whether this slice makes sense. So it either stays (broken if must be) or it is fixed, ensuring the ultimate chiselled filesystem has at least the same contents as what it currently delivers.

slices/iptables.yaml
cjdcordeiro: that's fair. if your use case requires all, then it's ok to start from a sub-optimal starting point. we can always slice it further whenever such a use case arises.

slices/libpam-runtime.yaml
Meulengracht: I just comitted db301b7d44e4361d7dd589d04a51c2ec6e5d6949 with readded slices, however the -defaults one is modified to deliver working content instead of broken. Would this be fine?

slices/libpam-runtime.yaml
```diff
@@ -18,40 +18,133 @@ slices:
...
   pam-config:
     contents:
       /usr/share/pam-configs/unix:
 
-  # default templates for the /etc/pam.d files that are used by
-  # pam-auth-update to generate the /etc/pam.d versions based on
-  # additional plugs in /usr/share/pam-configs/*.
   pam-defaults:
+    essential:
+      - libpam-runtime_config
```
cjdcordeiro, line 142: Why though? It feels like `pam-defaults` should deliver the defaults, which are the templates. Since you're not depending on `pam-defaults` anywhere, I suggest keeping it as is to avoid regressions.

cjdcordeiro (commented):
It's all good to me, but I just have a final comment and need a 2nd review.

slices/libpam-runtime.yaml
linostar: For clarification: files in `/usr/share/pam/` are just template files that are used by pam-auth-update to generate `/etc/pam.d/common-*` files. Since the mutate script is mimicking that, there is no need to keep them anymore.

I don't see any of our slices referencing `pam-defaults`, so it may be fine to remove those files from it, but there is some small risk that someone is using that slice in a container image to run `pam-auth-update` with it.

tests/spread/integration/iptables/task.yaml
```diff
@@ -0,0 +1,23 @@
+summary: Integration tests for iptables
+
+execute: |
+  rootfs="$(install-slices passwd_config libpam-runtime_config netbase_default-hosts sudo_bins iptables_bins)"
+  host=$(hostname)
+  
+  # setup /etc/hosts for the current hostname, otherwise
+  # sudo will be confused
+  sed -i "/127.0.0.1\tlocalhost/a 127.0.0.1\t$host" "$rootfs"/etc/hosts
```
linostar, line 9:
```suggestion
  sed -i "/127.0.0.1\slocalhost/a 127.0.0.1\t$host" "$rootfs"/etc/hosts
```
On my 24.04 system they are space-separated, so better use `\s` to make it work across all systems.

tests/spread/integration/sudo/task.yaml
```diff
@@ -0,0 +1,12 @@
+summary: Integration tests for sudo
+
+execute: |
+  rootfs="$(install-slices passwd_config libpam-runtime_config netbase_default-hosts coreutils_bins sudo_bins)"
+  host=$(hostname)
+  
+  # setup /etc/hosts for the current hostname, otherwise
+  # sudo will be confused
+  sed -i "/127.0.0.1\tlocalhost/a 127.0.0.1\t$host" "$rootfs"/etc/hosts
```
linostar, line 9:
```suggestion
  sed -i "/127.0.0.1\slocalhost/a 127.0.0.1\t$host" "$rootfs"/etc/hosts
```
see comment above.

linostar (changes requested):
Just small nitpicks.

slices/libpam-runtime.yaml
Meulengracht: @linostar i don't think that's an issue, there was no slice delivering pam-uath-update prior to this?

slices/libpam-runtime.yaml
```diff
@@ -18,40 +18,133 @@ slices:
   # pam-config and pam-defaults slices are only used for generation, and are useless
   # for anything else.
 
-  # Applications that need libpam for now should just rely on the config and var slices for now -
-  # until appropriate solution for emulating what pam-auth-update can do (or maybe something else).
+  # Emulate in part what pam-auth-update does. There is a short coming right now, we
+  # cannot include other files from other slices that install into
+  # /usr/share/pam-configs/*
+  # since this mutation script wont be able to access them as of writing.
```
rebornplusplus, line 24:
```suggestion
  # Emulate in part what pam-auth-update does. There is a short coming right now, we
  # cannot include other files from other slices that install into
  #   /usr/share/pam-configs/*
  # since this mutation script wont be able to access them as of writing.
```

slices/libpam-runtime.yaml
```diff
@@ -18,40 +18,133 @@ slices:
...
+
+        # session also includes settings from the session-noninteractive,
+        # but not the other way around
+        if fn == "session":
+          pb = build_block(confdata, "session-noninteractive", "Primary", pb)
+          ab = build_block(confdata, "session-noninteractive", "Additional", ab)
+
+        template = template.replace(idnames[i] + "_primary", pb)
+        template = template.replace(idnames[i] + "_additional", ab)
+        content.write("/etc/pam.d/common-" + fn, template)
```
rebornplusplus, line 131: Impressive! :100:

rebornplusplus: I do wish however we could store this big mutation script somewhere else instead of having it directly in the YAML file. But we can't do that yet sadly.

slices/libpam-runtime.yaml
rebornplusplus: Since we are keeping the `pam-defaults` slice in the fear that someone could still be using it, it seems reasonable to me to not change the contents it had. `config` provides the new contents anyway.

rebornplusplus (commented):
Looks quite nice, thank you! Impressive work on the mutation script. 

I think the only thing major in my comments below is the future of `pam-defaults` slice. Let's reach a conclusion on that. And then, I think we can merge it.

slices/libpam-runtime.yaml
cjdcordeiro: let's keep the pam-defaults as is to avoid regressions. And config can indeed deliver the mutated defaults and resulting configs

slices/libpam-runtime.yaml
```diff
@@ -18,28 +18,125 @@ slices:
...
+
+  # the following two slices are kept for backwards-compatibility
+  # however the pam-defaults have been modified to deliver correct
+  # files instead of templates
```
zhijie-yang, line 135:
```suggestion
  # The following two slices are kept for backwards compatibility.
  # However, the pam-defaults have been modified to deliver correct
  # files instead of templates.
```

slices/libpam-runtime.yaml
```diff
@@ -18,28 +18,125 @@ slices:
...
+        for i in range(len(lines)):
+          vals = lines[i].split(":")
+          if vals[0] == "Priority":
+            p = vals[1].strip()
+          elif vals[0] in ["Auth-Type", "Account-Type", "Session-Type", "Password-Type"]:
+            m, t = parse_type(vals)
+          elif vals[0] == "Session-Interactive-Only":
+            if vals[1].strip() == "yes":
+              m = "session"
+          elif vals[0] in ["Auth", "Account", "Session", "Password"]:
```
zhijie-yang, line 74: It seems that, in a clean container, we should use the values from the keys with `-Initial` suffices for these keys, e.g., `Auth-Initial`

zhijie-yang (changes requested):
Looks nice. Just checked the output of the `pam-auth-update` from a fresh container, and we have a set of diverged pam configs. The reason is that `pam-auth-update` reads the values from `*-Initial` for the first pam config of each kind.

``` bash
320 # return the lines for a given config name, type, and position in the stack
321 sub lines_for_module_and_type
322 {                           
323         my ($profiles, $mod, $type, $modpos) = @_;
324         if ($modpos == 0 && $profiles->{$mod}{$type . '-Initial'}) {
325                 return $profiles->{$mod}{$type . '-Initial'};
326         }   
327         return $profiles->{$mod}{$type};
328 }   
```

slices/libpam-runtime.yaml
cjdcordeiro: The remark seems reasonable to me. After inspecting existing usages of libpam (especially in containers like ubuntu/grafana), the -Initial configurations seem to be the default.

tests/spread/integration/libpam-runtime/task.yaml
```diff
@@ -0,0 +1,13 @@
+summary: Integration tests for libpam-runtime
+
+execute: |
+  rootfs="$(install-slices passwd_config libpam-runtime_config netbase_default-hosts coreutils_bins sudo_bins)"
+
+  # test that libpam correctly generated things
+  cat "$rootfs"/etc/pam.d/common-account | grep -E "account\s+\[success=1 new_authtok_reqd=done default=ignore\]\s+pam_unix\.so"
+  cat "$rootfs"/etc/pam.d/common-auth | grep -E "auth\s+\[success=1 default=ignore\]\s+pam_unix\.so\s+nullok"
+  cat "$rootfs"/etc/pam.d/common-password | grep -E "password\s+\[success=1 default=ignore\]\s+pam_unix\.so\s+obscure\s+use_authtok\s+try_first_pass\s+yescrypt"
```
zhijie-yang, line 9:
```suggestion
  cat "$rootfs"/etc/pam.d/common-password | grep -E "password\s+\[success=1 default=ignore\]\s+pam_unix\.so\s+obscure\s+yescrypt"
```

slices/libpam-runtime.yaml
```diff
@@ -18,28 +18,125 @@ slices:
...
+        for i in range(len(lines)):
+          vals = lines[i].split(":")
+          if vals[0] == "Priority":
+            p = vals[1].strip()
+          elif vals[0] in ["Auth-Type", "Account-Type", "Session-Type", "Password-Type"]:
+            m, t = parse_type(vals)
+          elif vals[0] == "Session-Interactive-Only":
+            if vals[1].strip() == "yes":
+              m = "session"
+          elif vals[0] in ["Auth-Initial", "Account-Initial", "Session-Initial", "Password-Initial"]:
```
zhijie-yang, line 74:
```suggestion
          elif vals[0] in ["Auth-Initial", "Account-Initial", 
                           "Session-Initial", "Password-Initial"]:
```

zhijie-yang (changes requested):
Some minor changes are required for the linting and tests. @cjdcordeiro

zhijie-yang (approved):
Looks good to me. Thanks for the changes!

cjdcordeiro (approved):
I think this is in a good state now. It would be nice to also have a final pass from @clay-lake

clay-lake (approved):
Approved! :) I'm a little concerned with the mutation script due its complexity. The current output seems consistent with what I expect in a fresh noble container, and the script should work in a fresh rootfs. Not now, but we might want to consider refactoring it if we need to adapt it in the future. Maybe something like [this](https://gist.github.com/clay-lake/0f978966ead504cf0b8df98dd66a0112)?   I made some changes while reviewing the script.

clay-lake (approved):

cjdcordeiro:
thanks @clay-lake 
that sounds reasonable. Maybe you can add that as a TODO to our list of actions when opening 25.04? (we have an issue for that)

PR #302, opened 2024-07-31
feat(24.04): Add slice definitions for libgif7, libpng16-16t64 and metapixel

sjauld:
# Proposed changes

Adds slices for libgif7, libpng16-16t64 and metapixel

GIFLIB is a package of portable tools and library routines for working with GIF images.

libpng is a library implementing an interface for reading and writing PNG (Portable Network Graphics) format files.

Metapixel is a program for generating photomosaics.

## Related issues/PRs
N/A

### Forward porting
N/A

## Checklist

* [x] I have read the [contributing guidelines](
https://github.com/canonical/chisel-releases/blob/main/CONTRIBUTING.md)
* [x] I have tested my changes ([see how](https://github.com/canonical/chisel-releases/blob/main/CONTRIBUTING.md#7-test-your-slices-before-opening-a-pr))
* [x] I have already submitted the [CLA form](
https://ubuntu.com/legal/contributors/agreement)

PR #308, opened 2024-08-07
feat(24.04): add openjdk slices

vpa1977:
# Proposed changes

Copy openjdk slice definitions from 22.04 as they are missing from 24.04.

Add smoke test for openjdk-8-jre-headless slice.

## Checklist

### DIFF
--- a/slices/ca-certificates-java.yaml
+++ b/slices/ca-certificates-java.yaml
@@ -1,0 +1,19 @@
+package: ca-certificates-java
+
+essential:
+  - ca-certificates-java_copyright
+
+slices:
+  data:
+    essential:
+      - ca-certificates_data
+      - libnss3_libs
+    contents:
+      /etc/default/cacerts:
+      /etc/ssl/certs/java/:
+      # we need to run `keytool` in postinst, but we can't yet do
+      # this from a mutation script
+
+  copyright:
+    contents:
+      /usr/share/doc/ca-certificates-java/copyright:
--- a/slices/fontconfig-config.yaml
+++ b/slices/fontconfig-config.yaml
@@ -1,0 +1,74 @@
+package: fontconfig-config
+
+essential:
+  - fontconfig-config_copyright
+
+slices:
+  config:
+    essential:
+      - fonts-dejavu-core_config
+      - fonts-dejavu-core_fonts
+    contents:
+      /etc/fonts/conf.avail/10-antialias.conf:
+      /etc/fonts/conf.avail/10-autohint.conf:
+      /etc/fonts/conf.avail/10-hinting-full.conf:
+      /etc/fonts/conf.avail/10-hinting-medium.conf:
+      /etc/fonts/conf.avail/10-hinting-none.conf:
+      /etc/fonts/conf.avail/10-hinting-slight.conf:
+      /etc/fonts/conf.avail/10-no-sub-pixel.conf:
+      /etc/fonts/conf.avail/10-scale-bitmap-fonts.conf:
+      /etc/fonts/conf.avail/10-sub-pixel-bgr.conf:
+      /etc/fonts/conf.avail/10-sub-pixel-rgb.conf:
+      /etc/fonts/conf.avail/10-sub-pixel-vbgr.conf:
+      /etc/fonts/conf.avail/10-sub-pixel-vrgb.conf:
+      /etc/fonts/conf.avail/10-unhinted.conf:
+      /etc/fonts/conf.avail/11-lcdfilter-default.conf:
+      /etc/fonts/conf.avail/11-lcdfilter-legacy.conf:
+      /etc/fonts/conf.avail/11-lcdfilter-light.conf:
+      /etc/fonts/conf.avail/20-unhint-small-vera.conf:
+      /etc/fonts/conf.avail/25-unhint-nonlatin.conf:
+      /etc/fonts/conf.avail/30-metric-aliases.conf:
+      /etc/fonts/conf.avail/40-nonlatin.conf:
+      /etc/fonts/conf.avail/45-generic.conf:
+      /etc/fonts/conf.avail/45-latin.conf:
+      /etc/fonts/conf.avail/49-sansserif.conf:
+      /etc/fonts/conf.avail/50-user.conf:
+      /etc/fonts/conf.avail/51-local.conf:
+      /etc/fonts/conf.avail/53-monospace-lcd-filter.conf:
+      /etc/fonts/conf.avail/60-generic.conf:
+      /etc/fonts/conf.avail/60-latin.conf:
+      /etc/fonts/conf.avail/65-fonts-persian.conf:
+      /etc/fonts/conf.avail/65-khmer.conf:
+      /etc/fonts/conf.avail/65-nonlatin.conf:
+      /etc/fonts/conf.avail/69-unifont.conf:
+      /etc/fonts/conf.avail/70-force-bitmaps.conf:
+      /etc/fonts/conf.avail/70-no-bitmaps.conf:
+      /etc/fonts/conf.avail/70-yes-bitmaps.conf:
+      /etc/fonts/conf.avail/80-delicious.conf:
+      /etc/fonts/conf.avail/90-synthetic.conf:
+      /etc/fonts/conf.d/10-antialias.conf:
+      /etc/fonts/conf.d/10-hinting-slight.conf:
+      /etc/fonts/conf.d/10-scale-bitmap-fonts.conf:
+      /etc/fonts/conf.d/11-lcdfilter-default.conf:
+      /etc/fonts/conf.d/20-unhint-small-vera.conf:
+      /etc/fonts/conf.d/30-metric-aliases.conf:
+      /etc/fonts/conf.d/40-nonlatin.conf:
+      /etc/fonts/conf.d/45-generic.conf:
+      /etc/fonts/conf.d/45-latin.conf:
+      /etc/fonts/conf.d/49-sansserif.conf:
+      /etc/fonts/conf.d/50-user.conf:
+      /etc/fonts/conf.d/51-local.conf:
+      /etc/fonts/conf.d/60-generic.conf:
+      /etc/fonts/conf.d/60-latin.conf:
+      /etc/fonts/conf.d/65-fonts-persian.conf:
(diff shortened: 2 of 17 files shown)


rebornplusplus:
@vpa1977 #302 is in.

vpa1977:
> @vpa1977 #302 is in.

Thank you!!! I have updated my branch.

cjdcordeiro (changes requested):
thanks. i've left some relevant comments in #313 .

It would also be good to address with the additional `libuuid1` dependency in a couple of SDFs, and why some many other dependencies are left out the jre SDF.

vpa1977:
> It would also be good to address with the additional `libuuid1` dependency in a couple of SDFs, and why some many other dependencies are left out the jre SDF.

Probably It would be better to split this MR into openjdk-8-jre dependencies and openjdk-17-jre-dependencies

cjdcordeiro:
it all lgtm, but as discussed offline, one change from 22.04 is the missing `ca-certificates-java`. it might be ok, but my pending question is: what about `/usr/share/ca-certificates-java/ca-certificates-java.jar`?

vpa1977:
> it all lgtm, but as discussed offline, one change from 22.04 is the missing `ca-certificates-java`. it might be ok, but my pending question is: what about `/usr/share/ca-certificates-java/ca-certificates-java.jar`?

`/usr/share/ca-certificates-java/ca-certificates-java.jar` is a maintainer script that updates Java certificate keystore.

tests/spread/integration/openjdk-8-jre-headless/task.yaml
cjdcordeiro, line 1: same comment as in https://github.com/canonical/chisel-releases/pull/309#discussion_r1731182668

vpa1977:
Note: CI fails due to 
```
2024-08-28 01:31:55 Found /home/runner/work/chisel-releases/chisel-releases/spread.yaml.
error: nothing matches provider filter
```

cjdcordeiro (approved):
Nice, thanks. This LGTM

slices/ca-certificates-java.yaml
```diff
@@ -0,0 +1,33 @@
+package: ca-certificates-java
+
+essential:
+  - ca-certificates-java_copyright
+
+slices:
+  bins:
```
rebornplusplus, line 7: Should this be named `data` in similarity with the one from 22.04?

slices/ca-certificates-java.yaml
```diff
@@ -0,0 +1,33 @@
...
+    essential:
+      - ca-certificates_bins
+      - ca-certificates_data-with-certs
+      # chisel tool does not allow per-architecture packages
+      # openjdk-8 is not published for RISCV in the archive
+      # - openjdk-8-jre-headless_security
```
rebornplusplus, line 13: I think the dependency should be the other way around? I mean, `ca-certificates-java` should not depend on `openjdk-..`, rather `openjdk-..` should depend on `ca-certificates-java`?

slices/fontconfig-config.yaml
```diff
@@ -0,0 +1,61 @@
...
+      /usr/share/fontconfig/conf.avail/51-local.conf:
+      /usr/share/fontconfig/conf.avail/60-generic.conf:
+      /usr/share/fontconfig/conf.avail/60-latin.conf:
+      /usr/share/fontconfig/conf.avail/65-fonts-persian.conf:
+      /usr/share/fontconfig/conf.avail/65-nonlatin.conf:
+      /usr/share/fontconfig/conf.avail/69-unifont.conf:
+      /usr/share/fontconfig/conf.avail/70-no-bitmaps-except-emoji.conf:
+      /usr/share/fontconfig/conf.avail/80-delicious.conf:
+      /usr/share/fontconfig/conf.avail/90-synthetic.conf:
+      /usr/share/xml/fontconfig/fonts.dtd:
```
rebornplusplus, line 57: Do we need this file?

slices/fonts-dejavu-core.yaml
```diff
@@ -0,0 +1,35 @@
+package: fonts-dejavu-core
```
rebornplusplus, line 1: It probably depends on `fonts-dejavu-mono` too, right? Do we need the mono fonts?

slices/libnss3.yaml
```diff
@@ -0,0 +1,18 @@
...
+
+slices:
+  libs:
+    essential:
+      - libc6_libs
+      - libnspr4_libs
+      - libsqlite3-0_libs
+    contents:
+      /usr/lib/*-linux-*/libnss3.so:
+      /usr/lib/*-linux-*/libnssutil3.so:
```
rebornplusplus, line 14: Do we need the other libs as well? (It's fine to leave them behind if we don't need them for now)

```
root@855047470eb3:~# dpkg -c libnss3_2%3a3.98-1build1_amd64.deb | grep "so$"
-rw-r--r-- root/root     14560 2024-04-08 22:14 ./usr/lib/x86_64-linux-gnu/libfreebl3.so
-rw-r--r-- root/root    833688 2024-04-08 22:14 ./usr/lib/x86_64-linux-gnu/libfreeblpriv3.so
-rw-r--r-- root/root   1229736 2024-04-08 22:14 ./usr/lib/x86_64-linux-gnu/libnss3.so
-rw-r--r-- root/root    584816 2024-04-08 22:14 ./usr/lib/x86_64-linux-gnu/libnssckbi.so
-rw-r--r-- root/root    137520 2024-04-08 22:14 ./usr/lib/x86_64-linux-gnu/libnssdbm3.so
-rw-r--r-- root/root    199128 2024-04-08 22:14 ./usr/lib/x86_64-linux-gnu/libnssutil3.so
-rw-r--r-- root/root    162208 2024-04-08 22:14 ./usr/lib/x86_64-linux-gnu/libsmime3.so
-rw-r--r-- root/root    358016 2024-04-08 22:14 ./usr/lib/x86_64-linux-gnu/libsoftokn3.so
-rw-r--r-- root/root    432864 2024-04-08 22:14 ./usr/lib/x86_64-linux-gnu/libssl3.so
```

slices/openjdk-8-jre-headless.yaml
```diff
@@ -0,0 +1,200 @@
...
+      /usr/lib/jvm/java-8-openjdk-*/jre/lib/tzdb.dat:
+
+  locale:
+    essential:
+      - openjdk-8-jre-headless_core
+    contents:
+      /usr/lib/jvm/java-8-openjdk-*/jre/lib/ext/cldrdata.jar:
+      /usr/lib/jvm/java-8-openjdk-*/jre/lib/ext/localedata.jar:
+
+  security:
```
rebornplusplus, line 63: It would be nice to have a comment here too.

rebornplusplus (commented):
Overall looks good! Only minor questions and suggestions.

slices/ca-certificates-java.yaml
vpa1977: This slice provides a runnable jar file that is used to populate Java keystore. I can split out data slice that will  create /etc/ssl/certs/java/ in order to keep the name.

slices/ca-certificates-java.yaml
vpa1977: This is  a chicken-and-egg packaging issue. 
ca-certificate-java used to have circular dependency with openjdk. 
The current state openjdk depends on ca-certificates-java and maintainer script checks if it can find java in the common locations. 

In the chisel story it does not make sense to make ca-certificates-java a dependency of openjdk - we do not need maintainer script installed in the chiselled rootfs unless we intend to update cacerts in place as per instructions below. 

Normal use-case would have something like https://github.com/canonical/rockcraft/pull/688/files#diff-e14fbacbe2674da891b0458de834490fc20512c6f00379284b659fce6ef0a927 where we copy the cacerts from the build host.

slices/fontconfig-config.yaml
vpa1977: The dtd is needed for validating conf files if the xml parser reading them has validation enabled. 
We probably want to leave it in.

slices/libnss3.yaml
vpa1977: Other libs are not needed for openjdk, but we may want to include them in a separate PR.

slices/ca-certificates-java.yaml
```diff
@@ -28,6 +28,12 @@ slices:
       /etc/ssl/certs/java/:
       /usr/share/ca-certificates-java/ca-certificates-java.jar:
 
+  data:
+    essential:
+      - ca-certificates_data-with-certs
+  contents:
+    /etc/ssl/certs/java/:
```
rebornplusplus, line 35: The indentation seems a bit odd there.

rebornplusplus (approved):
Looks good to me, thank you! Only have a nitpick below that I can probably update as well. :)

slices/ca-certificates-java.yaml
rebornplusplus: I am quite surprised that the workflows did not pick up on this. They are treating it as another slice which they should, but I am surprised that they did not complain about the unknown field.
https://github.com/canonical/chisel-releases/actions/runs/10673178458/job/29581840492#step:8:53
```
INFO: Installing ca-certificates-java_contents on amd64...
```

rebornplusplus: Pushed https://github.com/canonical/chisel-releases/pull/308/commits/8d5307ef4c4d5928adb47bbaf21e38acb02f5bba.

rebornplusplus (approved):

slices/ca-certificates-java.yaml
```diff
@@ -28,6 +28,12 @@ slices:
       /etc/ssl/certs/java/:
       /usr/share/ca-certificates-java/ca-certificates-java.jar:
 
+  data:
+    essential:
+      - ca-certificates_data-with-certs
```
cjdcordeiro, line 33: do you really need the certs? is it because you're going to run the `ca-certificates-java.jar` postinst? If so, then I would suggest calling this slice `data-with-certs` too, to be consistent with `ca-certificates`.

vpa1977: Yes, those are needed for ca-certificates-java.jar. 
Renamed the slice.

cjdcordeiro (approved):
tyvm :)

PR #309, opened 2024-08-07
feat(22.04): smoke test for openjdk-8-jre-headless

vpa1977:
# Proposed changes

Add a basic smoke test for openjdk-8-jre-headless. Install openjdk + busybox and ensure that core slice can create a Java machine.

Add missing s390x architecture to openjdk-8-jre-headless
### Forward porting

https://github.com/canonical/chisel-releases/pull/308

## Checklist


tests/spread/integration/openjdk-8-jre-headless/smoke.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/sh
+
+set -ex
+
+for java in `find /usr/lib/jvm -name java`; do
+# workaround missing /proc filesystem
```
cjdcordeiro, line 6: can you explain this? 

The test already runs inside a container, which is our typical and existing use case nonetheless. so why do we need this additional step here, given that `/proc` is a special directory inside the container?

vpa1977: We create a chroot with openjdk slice and this chroot does not have /proc. 
This breaks loading dynamic libraries with a relative path (e.g. $ORIGIN/...) as the loader checks /proc/self/exe to discover the value of $ORIGIN. 
The chroot should either have a /proc filesystem - for this we need to install mount slice and mount /proc after (install-slices) inside the chroot or do the workaround like this.

rebornplusplus (approved):
Looks good to me, thank you!

Will be good to merge two separate commits instead of squashing.

tests/spread/integration/openjdk-8-jre-headless/smoke.sh
```diff
@@ -0,0 +1,11 @@
...
+    mkdir -p ${rootfs}/proc/self
```
rebornplusplus, line 7: On second glance, you don't need the `rootfs` var here, do you?

tests/spread/integration/openjdk-8-jre-headless/smoke.sh
vpa1977: Also checked whether I can mount/bind-mount /proc - it seems that container lacks permissions for mounting /proc.

vpa1977:
Note: CI failed for an unrelated reason, needs a retry, but I probably do not have permissions to retry CI.

rebornplusplus (approved):
Looks good to me, thank you!

rebornplusplus:
> Note: CI failed for an unrelated reason, needs a retry, but I probably do not have permissions to retry CI.

Re-triggered the CI, passing now. It was probably failing before due to some issue on the Github runner.

tests/spread/integration/openjdk-8-jre-headless/task.yaml
```diff
@@ -0,0 +1,14 @@
+summary: Integration tests for openjdk-8-jre-headless
+
+environment:
+  SLICE/core: "core"
+
+execute: |
+  # Test different slice installations
```
cjdcordeiro, line 7: nit: this is just testing one

tests/spread/integration/openjdk-8-jre-headless/task.yaml
```diff
@@ -0,0 +1,14 @@
...
+  echo "SLICE=${SLICE}"
+  rootfs="$(install-slices busybox_bins openjdk-8-jre-headless_${SLICE})"
+  (cd ${rootfs}/bin && for applet in $(${rootfs}/bin/busybox --list); do \
+      ln -s busybox $applet ; \
+    done)
+  cp smoke.sh "${rootfs}/"
+  chroot ${rootfs} busybox sh /smoke.sh
```
cjdcordeiro, line 14: This feels like a lot of unnecessary setup...
 Why not

```
cd ${choot}
mkdir -p proc/self
for java in `find usr/lib/jvm -name java`; do
    ln -sf /${java} proc/self/exe
    chroot . ${java} -version
done
```

vpa1977: We need to set up applets because chroot contains only busybox executable.

cjdcordeiro: yes but why do you need busybox at all? shouldn't my snippet above work?

vpa1977: We need to install coreutils and bash for it.

vpa1977:
Note: CI seems to be broken with "Nothing matches provider filter error"

cjdcordeiro (approved):
just needed a rebase. lgtm thanks

PR #329, opened 2024-09-03
feat(24.04): add libcrack-runtime and its dependencies

Meulengracht:
# Proposed changes

Proposing libcrack-runtime and the dependencies it has. Unfortunately I was not able to write a mutation script that does what `update-cracklib` does due to limitations in how we write files. We need the capability to write raw file bytes instead of formatted as text, as it otherwise results in malaligned files.

### Forward porting

I can propose this for 24.10 when this is approved

## Checklist

## Additional Context

My attempt at the mutation script: https://paste.ubuntu.com/p/Jc5vvgk3tN/

### DIFF
--- a/slices/cracklib-runtime.yaml
+++ b/slices/cracklib-runtime.yaml
@@ -1,0 +1,46 @@
+package: cracklib-runtime
+
+essential:
+  - cracklib-runtime_copyright
+
+# Cracklib runs /usr/sbin/update-cracklib on install, and this
+# in turns generate the cracklib caches. This is mostly possible to do
+# as a mutation script, and is provided here:
+# https://paste.ubuntu.com/p/Jc5vvgk3tN/
+# however the files it writes are malformed as we cannot write raw bytes,
+# but instead we write files as text which results in larger files than
+# expected (i.e the generated cracklib_dict.hwm is 1618 bytes instead of 1024).
+# If we could use the above mutation script, then the contents of the 'bins'
+# slice would probably not be needed, together with most of the 'config'
+# slice.
+slices:
+  # the scripts use both 'file', 'find' and 'grep'
+  # so we ensure those dependencies are added.
+  # furthermore, libssl3t64 is required as well
+  bins:
+    essential:
+      - cracklib-runtime_config
+      - file_bins
+      - findutils_bins
+      - grep_bins
+      - libc6_libs
+      - libcrack2_libs
+      - libssl3t64_libs
+    contents:
+      /usr/sbin/cracklib-check:
+      /usr/sbin/cracklib-format:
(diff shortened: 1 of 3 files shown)


slices/libmagic-mgc.yaml
```diff
@@ -0,0 +1,15 @@
+package: libmagic-mgc
+
+essential:
+  - libmagic-mgc_copyright
+
+slices:
+  config:
```
clay-lake, line 7: Can we call this slice `data` since these files are binary not easily edited?

slices/cracklib-runtime.yaml
```diff
@@ -0,0 +1,46 @@
+package: cracklib-runtime
+
+essential:
+  - cracklib-runtime_copyright
+
+# Cracklib runs /usr/sbin/update-cracklib on install, and this
+# in turns generate the cracklib caches. This is mostly possible to do
+# as a mutation script, and is provided here:
+# https://paste.ubuntu.com/p/Jc5vvgk3tN/
```
rebornplusplus, line 9: Nice work there! It's a shame we cannot raw bytes yet.
On another note, could this link expire some day? If so, I would want there to be something that does not expire, Github Gist maybe?

tests/spread/integration/cracklib-runtime/task.yaml
```diff
@@ -0,0 +1,31 @@
+summary: Integration tests for cracklib-runtime
+
+systems:
+  - -ubuntu-24.04-ppc64le
+  - -ubuntu-24.04-s390x
```
rebornplusplus, line 5: Curious to know why you don't need these.

clay-lake (changes requested):

rebornplusplus (changes requested):
Nice changes! I left a few comments below.

In addition to the changes, could you please add a spread test for `file` as well? Just a smoke test will do. :)

tests/spread/integration/cracklib-runtime/task.yaml
```diff
@@ -0,0 +1,31 @@
...
+
+  if ! [ -e "$rootfs"/var/cache/cracklib/cracklib_dict.pwi ]; then
+    echo "expected /var/cache/cracklib/cracklib_dict.pwi to exist"
+    exit -1
+  fi
+
+  if ! [ -e "$rootfs"/var/cache/cracklib/src-dicts ]; then
+    echo "expected /var/cache/cracklib/src-dicts to exist"
+    exit -1
+  fi
```
linostar, line 31: Can you also add smoke tests for the rest of the binaries in cracklib-runtime_bins slice?

linostar (changes requested):

PR #337, opened 2024-09-04
slices: add bzip2 slice

Meulengracht:
# Proposed changes

Add the bzip2 package

### Forward porting

Can be proposed for 24.10 if approved

## Checklist

### DIFF
--- a/slices/bzip2.yaml
+++ b/slices/bzip2.yaml
@@ -1,0 +1,27 @@
+package: bzip2
+
+essential:
+  - bzip2_copyright
+
+slices:
+  bins:
+    essential:
+      - libbz2-1.0_libs
+      - libc6_libs
+    contents:
+      /usr/bin/bunzip2:
+      /usr/bin/bzcat: { symlink: /usr/bin/bunzip2 }
+      /usr/bin/bzcmp:
+      /usr/bin/bzdiff:
+      /usr/bin/bzegrep:
+      /usr/bin/bzexe:
+      /usr/bin/bzfgrep:
+      /usr/bin/bzgrep:
+      /usr/bin/bzip2: { symlink: /usr/bin/bunzip2 }
+      /usr/bin/bzip2recover:
+      /usr/bin/bzless:
+      /usr/bin/bzmore:
+
+  copyright:
+    contents:
+      /usr/share/doc/bzip2/copyright:
--- a/tests/spread/integration/bzip2/task.yaml
+++ b/tests/spread/integration/bzip2/task.yaml
@@ -1,0 +1,17 @@
+summary: Integration tests for bzip2
+
+execute: |
+  rootfs="$(install-slices bzip2_bins)"
+
+  # generate a test file
+  echo "this is a test file we use for compression with bzip2" > "${rootfs}/input.txt"
+
+  # compress file we've generated
+  chroot "${rootfs}" bzip2 -z -k input.txt
+  mv "${rootfs}/input.txt" "${rootfs}/input.original"
+
+  # decompress it
+  chroot "${rootfs}" bzip2 -d input.txt.bz2
+
+  # ensure its identical to the original file
+  cmp "${rootfs}/input.txt" "${rootfs}/input.original"


slices/bzip2.yaml
```diff
@@ -0,0 +1,27 @@
...
+
+essential:
+  - bzip2_copyright
+
+slices:
+  bins:
+    essential:
+      - libbz2-1.0_libs
+      - libc6_libs
+    contents:
```
clay-lake, line 11: Some of these bins are actually scripts and probably need dash_bins and one or more base-files slices. Could we separate the executables like this xz-utils PR? https://github.com/canonical/chisel-releases/pull/338/files


```
/usr/bin/bzdiff:       POSIX shell script, ASCII text executable
/usr/bin/bzexe:        POSIX shell script, ASCII text executable
/usr/bin/bzgrep:       POSIX shell script, ASCII text executable
/usr/bin/bzmore:       POSIX shell script, ASCII text executable
```

clay-lake (changes requested):
Thanks this is a good package to add, although I think the bins slice needs to be split.

slices/bzip2.yaml
Meulengracht: Why do we want to introduce dash as a dependency? What if people want to use bash?

Meulengracht: I asked questions in your PR so we can clarify this - I'm not sure I completely agree with this

clay-lake: No problem. btw that draft PR just contains some examples for another PR I was assisting, it probably wont get merged. Most of the work belongs to endersonmaia, Maybe you want to move some of the comments to -> https://github.com/canonical/chisel-releases/pull/338

clay-lake: dash is the default system shell on debian based systems. It is normally linked to /bin/sh. As far as I know, users can still install and execute bash form bash_bins

Here is an example from a relatively fresh multipass instance
```
ubuntu@noble:~$ head -n 10 /usr/bin/bzmore
#!/bin/sh

# Bzmore wrapped for bzip2, 
# adapted from zmore by Philippe Troin <phil@fifi.org> for Debian GNU/Linux.

PATH="/usr/bin:$PATH"; export PATH

prog=`echo $0 | sed 's|.*/||'`
case "$prog" in
	*less)	more=less	;;
ubuntu@noble:~$ ll /bin/sh
lrwxrwxrwx 1 root root 4 Mar 31 10:47 /bin/sh -> dash*
```

Meulengracht: I know it's the default shell in ubuntu, but I question whether or not it should be a forced dependency when alternatives exist *and* it's not a dependency for the official package. Some dependencies make sense to introduce as they are actually required to run and it's pretty binary (i.e coreutils vs not coreutils), but the choice of shell interpreter, `dash` is not actually a required choice.

I'm not saying I'm right here, but I just want to explain my opinion about why I think it's shouldn't be, if you and your team feel strongly otherwise I'll let it be

clay-lake: >  I question whether or not it should be a forced dependency when alternatives exist 
I agree, and that is a good point. Our team had a short discussion about this recently. What if someone wanted a busybox based image? I think we need a larger discussion this, its probably outside the scope of this PR. Especially since some slices use dash for the same reason.

Not to argue against my first point, but for the time being I think we should be consistent with expectations of a Ubuntu 24.04 distribution. This more reliable since it is widely used as a default shell, and it is what an end user probably expects to be there. 

I'd like to hear more voices on this, I've only been on the project for a few months. :)

Meulengracht: @rebornplusplus @cjdcordeiro I belive we need some more input here

clay-lake: I had a discussion with @cjdcordeiro yesterday on this. We'll get the user to link in the shell after chiseling. The current PR should be fine then. Apologies for the confusion.

clay-lake (approved):

slices/bzip2.yaml
cjdcordeiro: Y I think this is something that may get some peace of mind once we can also handle conflicts. But for now, I can actually suggest one thing, **if needed** (I won't stop the PR because of this): you can create shell-specific slices. If really needed, one can have `bins-dash`, etc. to create options. 

I'm not sure this is the ideal UX, but it surely is possible and avoids any post-inst hooks.

cjdcordeiro (changes requested):
lgtm but we need to test all binaries and scripts, just to make sure they're properly loaded.

And as alluded by the above discussion, if there are scripts that **assume** a post-inst operation, then at least leave a comment in the YAML

linostar:
Hey @Meulengracht 
Can you add simple tests for the rest of the bzip binaries/shell scripts, in addition to the comment regarding manuallylinking to /bin/sh?

clay-lake:
Currently waiting on less in PR #484

cjdcordeiro:
#484 is merged

slices/bzip2.yaml
cjdcordeiro: good enough to move on

cjdcordeiro (approved):
lgtm after some minor naming and linting fixes

PR #338, opened 2024-09-04
xz utils changes

clay-lake:


slices/xz-utils.yaml
```diff
@@ -0,0 +1,36 @@
...
+      /usr/bin/lzmainfo:
+      /usr/bin/unxz:
+      /usr/bin/xz:
+      /usr/bin/xzcat:
+  utils:
+    essential:
+      - base-files_bin
+      - base-files_tmp
+      - coreutils_bins
+      - dash_bins
```
Meulengracht, line 21: I disagree with adding dash as a dependency if not explicitly mentioned by the package itself. Dash is only one of the possible shells, and bash would also work. I don't think we should get to decide this

slices/xz-utils.yaml
```diff
@@ -0,0 +1,36 @@
+package: xz-utils
+
+essential:
+  - xz-utils_copyright
+
+slices:
+  bins:
```
Meulengracht, line 7: this should either depend on the shell scripts the package brings as they are executables as well, or we should introduce an `execs` slice that bundles all of them imo

slices/xz-utils.yaml
```diff
@@ -0,0 +1,36 @@
...
+      - libc6_libs
+      - liblzma5_libs
+    contents:
+      /usr/bin/lzmainfo:
+      /usr/bin/unxz:
+      /usr/bin/xz:
+      /usr/bin/xzcat:
+  utils:
+    essential:
+      - base-files_bin
```
Meulengracht, line 18: while I understand this, I don't think this is neccesarily needed, most of shell scripts use `#/bin/sh` which won't be solved by this anyway without any manual setup

slices/xz-utils.yaml
```diff
@@ -0,0 +1,36 @@
...
+      - base-files_tmp
```
Meulengracht, line 19: why is this depending on /tmp?

slices/xz-utils.yaml
```diff
@@ -0,0 +1,36 @@
...
+  bins:
+    essential:
+      - libc6_libs
+      - liblzma5_libs
+    contents:
+      /usr/bin/lzmainfo:
+      /usr/bin/unxz:
+      /usr/bin/xz:
+      /usr/bin/xzcat:
+  utils:
```
Meulengracht, line 16: This slice should *as a minimum* explain why these extra package dependencies are introduced, and also explain which scripts depend on what

slices/xz-utils.yaml
clay-lake: Let me fix that now that we are not going with any pre-defined shell. :)

slices/xz-utils.yaml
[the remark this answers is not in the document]
clay-lake: Sounds good to me. updated.

slices/xz-utils.yaml
clay-lake: Using the `strings` util, it does not look like these bins reference the scripts. That seems to fix expectations as it looks like the scripts extend common utils using the binaries. 

Here is a sample from xzgrep:
```
# xzgrep -- a wrapper around a grep program that decompresses files as needed
# Instead of unsetting XZ_OPT, just make sure that xz will use file format
xz='xz --format=auto'

```

slices/xz-utils.yaml
clay-lake: Good idea, added some inline comments.

slices/xz-utils.yaml
clay-lake: It is possible to install a shell without the link from  /usr/bin -> /bin. This slice only contains a few directories or links so I think it is ok to add for convenience.

slices/xz-utils.yaml
clay-lake: Looking around with grep, the xz diff utils make use of the /tmp directory.

tests/spread/integration/diffutils/task.yaml
```diff
@@ -0,0 +1,13 @@
+summary: Integration tests for diffutils
+
+execute: |
+  rootfs="$(install-slices diffutils_bins)"
+
+  apt install -y gcc
+  arch=$(gcc -dumpmachine)
```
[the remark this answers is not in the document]
clay-lake: Same for me, I removed both the lines.

tests/spread/integration/xz-utils/FILE1
```diff
@@ -0,0 +1 @@
+one
```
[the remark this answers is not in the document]
clay-lake: Sounds good to me. changed.

tests/spread/integration/diffutils/FILE1
```diff
@@ -0,0 +1 @@
+one
```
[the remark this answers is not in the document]
clay-lake: changed as per earlier comment.

rebornplusplus (approved):
Looks good to me, one comment for the future.

cjdcordeiro (changes requested):
(tests)

rebornplusplus:
> (tests)

The tests are passing actually after a re-trigger. Was failing previously due to unrelated issues.

slices/xz-utils.yaml
```diff
@@ -0,0 +1,33 @@
...
+      /usr/bin/unxz:
+      /usr/bin/xz:
+      /usr/bin/xzcat:
+  scripts:  # scripts which add xz compression to common utils
+    essential:
+      - base-files_tmp  # utils require temporary storage
+      # shebang in scripts reference interpreter in /bin not /usr/bin
+      - coreutils_bins
+      - diffutils_bins  # needed for zx cmp and diff
+      - grep_bins  # needed for zx grep utils
```
cjdcordeiro, line 22:
```suggestion
      - diffutils_bins  # needed for xz cmp and diff
      - grep_bins  # needed for xz grep utils
```
?

cjdcordeiro (changes requested):

slices/xz-utils.yaml
clay-lake: Whoops! Thanks :)

slices/xz-utils.yaml
```diff
@@ -0,0 +1,33 @@
...
+    contents:
+      /usr/bin/lzmainfo:
+      /usr/bin/unxz:
+      /usr/bin/xz:
+      /usr/bin/xzcat:
+  scripts:  # scripts which add xz compression to common utils
+    essential:
+      - base-files_tmp  # utils require temporary storage
+      # shebang in scripts reference interpreter in /bin not /usr/bin
+      - coreutils_bins
```
cjdcordeiro, line 20: just realized that coreutils brings a lot of stuff. do you need it all? the fact that it isn't a marked dependency makes it even more suspicious.

clay-lake: `xzcmp` and `xzdiff` depend on coreutils for `mkdir` and `expr` unfortunately. Undeclared dependencies in apt was the core issue in the original PR. This is a good talking point in the upcoming sprint. :)

Also I found a missing change just above line 20. Should be fixed now

cjdcordeiro: If it's only a subset of coreutils bins that is needed, I actually think this is a perfect use case to explore the slicing power -> just create the slices for those binaries, thus ditching any other coreutils binaries that are not needed

> This is a good talking point in the upcoming sprint. :)

let's do it! (add a slot for it pls :) )

clay-lake: Just pushed some changes for coreutils_mkdir and coreutils_expr slices + individual tests. The rootfs in the xzutils task should be much smaller now :)

zhijie-yang (approved):
LGTM!

cjdcordeiro (approved):
LGTM thanks for the coreutils changes

PR #340, opened 2024-09-05
feat(24.04): add dbus-user-session and libpam-systemd

Meulengracht:
# Proposed changes

Add dbus-user-session which is package that simply carries systemd units. It also depends on libpam-systemd which is a pure library, so added very basic test for the libpam-systemd as I'm a bit unsure how to best test this. What the user units does are executing dbus-daemon in user context instead of system context as we already test for dbus.

## Checklist

### DIFF
--- a/slices/dbus-user-session.yaml
+++ b/slices/dbus-user-session.yaml
@@ -1,0 +1,21 @@
+package: dbus-user-session
+
+essential:
+  - dbus-user-session_copyright
+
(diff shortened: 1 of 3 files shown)


Meulengracht:
This PR needs https://github.com/canonical/chisel-releases/pull/306

cjdcordeiro:
@Meulengracht #306 is in

PR #342, opened 2024-09-05
feat(24.04): add finalrd and add bins slice to libc-bin

Meulengracht:
# Proposed changes

Finalrd is needed by Ubuntu Core - added the package and extended libc-bin by adding the binaries it carries. Those are needed not only by the finalrd script (which invokes ldconfig), but also by our hooks in ubuntu core.

I added a test for finalrd, but as expected it will fail, so we just ensure it fails in the way we expect.

## Checklist

### DIFF
--- a/slices/finalrd.yaml
+++ b/slices/finalrd.yaml
@@ -5,9 +5,9 @@ essential:
 
 slices:
   # finalrd is a shell script that relies on a couple of packages to even
-  # work correctly: libc-bin, systemd, mount, grep and coreutils. If initramfs-tools
-  # is supplied (which is optional), it also needs findutils and debianutils.
-  # so rely on atleast the first set of binaries.
+  # work correctly: libc-bin, systemd, mount, grep and coreutils. 
+  # If initramfs-tools is supplied (which is optional), it also needs 
+  # findutils and debianutils. so rely on atleast the first set of binaries.
   bins:
     essential:
       - coreutils_bins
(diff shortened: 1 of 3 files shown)


slices/finalrd.yaml
```diff
@@ -0,0 +1,33 @@
...
+  - finalrd_copyright
+
+slices:
+  # finalrd is a shell script that relies on a couple of packages to even
+  # work correctly: libc-bin, systemd, mount, grep and coreutils.
+  # If initramfs-tools is supplied (which is optional), it also needs
+  # findutils and debianutils. so rely on atleast the first set of binaries.
+  bins:
+    essential:
+      - coreutils_bins
```
clay-lake, line 13: Since this is a shell script should we add `dash_bins` for the shell and `base-files_bin` for `/bin`?

clay-lake (changes requested):
This is a difficult slice to test. I'll ask around our team and see if anyone has an additional strategy. Other than that I only have the one question below.

Thanks! :)

slices/libc-bin.yaml
```diff
@@ -4,6 +4,33 @@ essential:
   - libc-bin_copyright
 
 slices:
+  bins:
```
cjdcordeiro, line 7: We need tests for this

slices/finalrd.yaml
cjdcordeiro: you probably don't need the entire `coreutils`, so I suggest creating a new coreutils slice(s) with just the bins you need for this.

cjdcordeiro (changes requested):

PR #345, opened 2024-09-05
feat(24.04): add libnss-{extrausers,mdns}

Meulengracht:
# Proposed changes

Add libnss extra libraries like extrausers and mdns:

* extrausers does not exist on i386

## Checklist

### DIFF
--- a/slices/libnss-extrausers.yaml
+++ b/slices/libnss-extrausers.yaml
@@ -1,0 +1,21 @@
+package: libnss-extrausers
+
+essential:
+  - libnss-extrausers_copyright
+
+slices:
+  libs:
+    essential:
+      - libc6_libs
+      - libnss-extrausers_var
+    contents:
+      /usr/lib/libnss_extrausers.so.2:
+      /usr/lib32/libnss_extrausers.so.2:
+
+  var:
+    contents:
+      /var/lib/extrausers/:
+
+  copyright:
+    contents:
+      /usr/share/doc/libnss-extrausers/copyright:
--- a/slices/libnss-extrausers.yaml
+++ b/slices/libnss-extrausers.yaml
@@ -9,8 +9,8 @@ slices:
       - libc6_libs
       - libnss-extrausers_var
     contents:
-      /usr/lib/libnss_extrausers.so.2:
-      /usr/lib32/libnss_extrausers.so.2: { arch: [amd64, s390x] }
+      /usr/lib/libnss_extrausers.so.*:
+      /usr/lib32/libnss_extrausers.so.*: { arch: [amd64, s390x] }
 
   var:
     contents:
--- a/slices/libnss-extrausers.yaml
+++ b/slices/libnss-extrausers.yaml
@@ -10,7 +10,7 @@ slices:
       - libnss-extrausers_var
     contents:
       /usr/lib/libnss_extrausers.so.2:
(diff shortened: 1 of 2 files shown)


slices/libnss-mdns.yaml
```diff
@@ -0,0 +1,28 @@
...
+
+  mdns4:
+    contents:
+      /usr/lib/*-linux-*/libnss_mdns4.so.*:
+      /usr/lib/*-linux-*/libnss_mdns4_minimal.so.*:
+
+  mdns6:
+    contents:
+      /usr/lib/*-linux-*/libnss_mdns6.so.*:
+      /usr/lib/*-linux-*/libnss_mdns6_minimal.so.*:
```
clay-lake, line 24: Should we split out the minimal lib of each mdns version into its own slice?

clay-lake (commented):
Thanks for the addition!

slices/libnss-extrausers.yaml
```diff
@@ -0,0 +1,21 @@
...
+  - libnss-extrausers_copyright
+
+slices:
+  libs:
+    essential:
+      - libc6_libs
+      - libnss-extrausers_var
+    contents:
+      /usr/lib/libnss_extrausers.so.2:
+      /usr/lib32/libnss_extrausers.so.2: { arch: [amd64, s390x] }
```
rebornplusplus, line 13: Let's be consistent with the other file with regards to the trailing version number in the path, `.2` in this case. Let's use globs like below, if they can change.

slices/libnss-mdns.yaml
```diff
@@ -0,0 +1,28 @@
...
+
+slices:
+  libs:
+    essential:
+      - libc6_libs
+      - libnss-mdns_mdns4
+      - libnss-mdns_mdns6
+    contents:
+      /usr/lib/*-linux-*/libnss_mdns.so.*:
+      /usr/lib/*-linux-*/libnss_mdns_minimal.so.*:
```
rebornplusplus, line 14: How about adding a `mdns` slice of it's own and then `libs` can include all of them together?

rebornplusplus (changes requested):
The slices look good! Few comments below.

rebornplusplus (approved):
Looks good to me, thank you very much!

slices/libnss-mdns.yaml
```diff
@@ -0,0 +1,51 @@
...
+  - libnss-mdns_copyright
+
+slices:
+  libs:
+    essential:
+      - libc6_libs
+      - libnss-mdns_full
+      - libnss-mdns_minimal
+
+  full:
```
cjdcordeiro, line 13: is `full` the right name? Did you come up with it or is it how the package references these libs?

linostar: The package doesn't reference those libs individually, but the lib file names indicate whether they are minimal or otherwise (aka full), and the version numbers, so it looks that this slice naming is reasonable.

linostar (approved):

slices/libnss-mdns.yaml
cjdcordeiro: @canonical/rocks what do you think we should make of this? `standard`?

zhijie-yang: I agree with `standard`, as `full` is a term that we use to indicate whether a rock is `full(-based)` or `bare(-based)`.

PR #379, opened 2024-10-23
Rework of PR #49 "[22.04] Add slices for libgnutls, libsoup and glib-networking"

clay-lake:
Rework of #49, as it has not been updated in a while.

This PD adds the following changes:
- The data slice in `publicsuffix` is more explicit following earlier comments in #49 
- `glib-networking` and dependent `gsettings-desktop-schemas` were removed. Since these packages related to the gnome desktop environment, I am hesitant to add them unless they can be tested in a minimal rootfs.


rebornplusplus:
The CI is failing.

PR #382, opened 2024-10-24
fix(22.04/libnss3): libs should contain all libraries

### LINKED ISSUE #381: libnss3 does not contain libsmime3.so
In file `libnss3.yaml` (both 22.04 and 24.04), the libs slice only take `libnss3.so` and `libnssutil3.so`. But the package `libnss3` contains many other .so files [https://ubuntu.pkgs.org/22.04/ubuntu-main-amd64/libnss3_3.68.2-0ubuntu1_amd64.deb.html](https://ubuntu.pkgs.org/22.04/ubuntu-main-amd64/libnss3_3.68.2-0ubuntu1_amd64.deb.html)

One of them is `libsmime3.so` which can only found in this package. We need it. How do you suggest to solve this case?

rebornplusplus:
This PR restructures the libnss3 slices. It renames the former `libs` slice to `nss`, to only include nss3-specific libraries. It also adds a new slice `smime` to include S/MIME libraries.

Finally, the `libs` slice is added back, which includes all of the slices that provide libraries. The rationale is that "libs" slices should contain all libraries coming from the package.

Note that the ca-certificates-java slice has also been updated to reflect this change.

Resolves #381

### DIFF
--- a/slices/ca-certificates-java.yaml
+++ b/slices/ca-certificates-java.yaml
@@ -7,7 +7,7 @@ slices:
   data:
     essential:
       - ca-certificates_data
-      - libnss3_libs
+      - libnss3_nss
     contents:
       /etc/default/cacerts:
       /etc/ssl/certs/java/:
--- a/slices/ca-certificates-java.yaml
+++ b/slices/ca-certificates-java.yaml
@@ -7,7 +7,6 @@ slices:
   data:
     essential:
       - ca-certificates_data
-      - libnss3_nss
     contents:
       /etc/default/cacerts:
       /etc/ssl/certs/java/:
--- a/slices/ca-certificates-java.yaml
+++ b/slices/ca-certificates-java.yaml
@@ -7,6 +7,7 @@ slices:
   data:
     essential:
       - ca-certificates_data
+      - libnss3_nss
     contents:
       /etc/default/cacerts:
       /etc/ssl/certs/java/:
--- a/slices/libnss3.yaml
+++ b/slices/libnss3.yaml
@@ -27,11 +27,6 @@ slices:
     contents:
       /usr/lib/*-linux-*/libnss3.so:
       /usr/lib/*-linux-*/libnssutil3.so:
-      # Checksum files (.chk) are required for the NSS softoken to operate in
-      # FIPS 140 mode.
-      # https://www-archive.mozilla.org/projects/security/pki/nss/tech-notes/tn6
-      /usr/lib/*-linux-*/nss/libsoftokn3.chk:
-      /usr/lib/*-linux-*/nss/libsoftokn3.so:
 
   smime:
     essential:
--- a/slices/libnss3.yaml
+++ b/slices/libnss3.yaml
@@ -23,6 +23,7 @@ slices:
     essential:
       - libc6_libs
       - libnspr4_libs
+      - libsqlite3-0_libs
     contents:
       /usr/lib/*-linux-*/libnss3.so:
       /usr/lib/*-linux-*/libnssutil3.so:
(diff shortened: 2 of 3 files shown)


slices/libnss3.yaml
```diff
@@ -1,18 +1,39 @@
...
+      # slices yet. Upon creation of new slices containing those libraries,
+      # make sure to add those slices in this list.
+      - libnss3_nss
+      - libnss3_smime
+
+  nss:
     essential:
       - libc6_libs
       - libnspr4_libs
-      - libsqlite3-0_libs
```
rebornplusplus, line 11: Turns out we never needed `libsqlite3-0_libs` for the `libnss3.so` and `libnssutil3.so` files.

rebornplusplus:
@vpa1977, will you please take a look at this PR? I could not add you to the reviewers list up there on the top-right.

slices/libnss3.yaml
rebornplusplus: Turns out I was wrong :facepalm: Added the slice back. Thanks @vpa1977.

slices/libnss3.yaml
```diff
@@ -23,6 +23,7 @@ slices:
     essential:
       - libc6_libs
       - libnspr4_libs
+      - libsqlite3-0_libs
     contents:
       /usr/lib/*-linux-*/libnss3.so:
       /usr/lib/*-linux-*/libnssutil3.so:
```
vpa1977, line 29: We might want to add libsoftokn.

slices/ca-certificates-java.yaml
```diff
@@ -7,7 +7,7 @@ slices:
   data:
     essential:
       - ca-certificates_data
-      - libnss3_libs
```
cjdcordeiro, line 10: I was about to complain about this change :p but it seems ok cause the contents are the same. At least for now. Are you sure `data` only depends on `nss` and not the other libraries?

cjdcordeiro (approved):
It LGTM. There's a small comment for clarification. I'll also wait for @vpa1977 's review.

vpa1977 (approved):
We might want to add libsoftokn slice, but this can be done separately.

slices/ca-certificates-java.yaml
rebornplusplus: The `ca-certificates-java_data` slice currently have a `cacerts` file and an empty directory. We do not need the `nss` libraries at all **directly**, but they are used by the `openjdk` slices.

Anyway, yeah, it should be alright to include only the `nss` slice here.

rebornplusplus:
> The `ca-certificates-java_data` slice currently have a `cacerts` file and an empty directory. We do not need the `nss` libraries at all **directly**, but they are used by the `openjdk` slices.

Hiya @vpa1977 and @cjdcordeiro, based on [this comment above](https://github.com/canonical/chisel-releases/pull/382#discussion_r1847952748), I have removed the `libnss3_nss` from `ca-certificates-java_data` and added it directly to `openjdk-8-jre-headless_security` slice. Please let me know what you think!

Also, I don't know if I was asleep or something before, but I finally added the `libsoftokn.so` files. :facepalm:

vpa1977 (changes requested):
I have added an integration test https://github.com/rebornplusplus/chisel-releases/pull/12 for the soft token and it is currently fails due to the provider error. 
Running it installing nss3 package normally succeeds. 

Maybe we should drop soft token changes and make a separate PR with the integration test?

rebornplusplus:
> Maybe we should drop soft token changes and make a separate PR with the integration test?

I agree. Let's add the soft token stuff in a later PR with passing tests.

slices/ca-certificates-java.yaml
cjdcordeiro: Thanks for adding it back. And ok for the coming PR. Feel free to open it already

cjdcordeiro (approved):

vpa1977 (approved):

cjdcordeiro:
Waiting for PRs for newer releases

PR #387, opened 2024-11-10
feat(24.04): add python3 slice

linostar:
# Proposed changes

Add python3 slice, which depends on python3.12 slices and add a couple of symlinks over that.

## Related issues/PRs

N/A

### Forward porting

N/A

## Checklist

## Additional Context

N/A


slices/python3.yaml
```diff
@@ -0,0 +1,25 @@
...
+
+  standard:
+    essential:
+      - python3.12_standard
+
+  utils:
+    essential:
+      - python3.12_utils
+    contents:
+      /usr/bin/pdb3:
```
clay-lake, line 19: Do these files need to be symlinked to their python 3.12 alternatives in python3.12_utils?
ex. /usr/bin/pdb3 -> /usr/bin/pdb3.12

slices/python3.yaml
```diff
@@ -0,0 +1,25 @@
+package: python3
+
+essential:
+  - python3_copyright
+
+slices:
```
clay-lake, line 6: Should we modify or add a slice to link python3.12 to python3 and possibly python?

slices/python3.yaml
linostar: They are already symlinks when installed from python3 package, and they are linked properly (pdb3 linked to pdb3.12, etc).

zhijie-yang (approved):
Looks good to me, thanks!

slices/python3.yaml
linostar: A link from python to python3.12 is not created by the postinst script of python3 package, so I didn't include that in this slice. Besides, linking python3 to a specific python3.x version will cause conflicts when there are multiple python3.x packages in the same release (e.g. python3.10 and python3.11 in 22.04).

clay-lake (approved):
LGTM Thanks!

slices/python3.yaml
cjdcordeiro: I also think we should have the `python3` symlink, as that might be the most useful thing about this SDF.

It is true that it isn't set by this pkg. The symlink is created by `python3-minimal` which is a dependency that is missing from this PR.

cjdcordeiro (changes requested):

slices/python3.yaml
cjdcordeiro: The latest change artificially creates the python3 symlink, but I don't think that's the right way to do it. The symlink is created by python3-minimal which is a dependency that should be included in this PR.

cjdcordeiro (changes requested):

linostar:
@cjdcordeiro python3-minimal is already included as an indirect dependency (see my comment above).
I'll move the symlink to python3-minimal SDF instead.

slices/python3.yaml
linostar: Done.

slices/libpython3-stdlib.yaml
```diff
@@ -0,0 +1,110 @@
+package: libpython3-stdlib
+
```
cjdcordeiro, line 2: missing the copyright

slices/python3-minimal.yaml
```diff
@@ -0,0 +1,29 @@
...
+slices:
+  bins:
+    essential:
+      - python3-minimal_config
+      - python3-minimal_modules
+      - python3.12-minimal_bins
+    contents:
+      /usr/bin/py3clean:
+      /usr/bin/py3compile:
+      /usr/bin/py3versions:
```
cjdcordeiro, line 15: do we need these? 

If we want to keep the `bins` slice, all of these must be tested.

In the end, we only care about the python3 symlink right? If so, I'd create a dedicated slice for it (and test it).

slices/python3-minimal.yaml
```diff
@@ -0,0 +1,29 @@
...
+      /usr/bin/python3:
+
+  config:
+    contents:
+      /usr/share/python3/debian_defaults:
+
+  modules:
```
cjdcordeiro, line 22: why are these called "modules"?

Does `py3versions.py` need `debpython`? The former is related to the bin py3versions symlink, so they could stay together.

slices/python3.yaml
```diff
@@ -0,0 +1,57 @@
...
+      - libpython3-stdlib_language
+      - libpython3-stdlib_markup-tools
+      - libpython3-stdlib_multimedia
+      - libpython3-stdlib_net-data
+      - libpython3-stdlib_numeric
+      - libpython3-stdlib_osx-support
+      - libpython3-stdlib_pydoc
+      - libpython3-stdlib_text
+      - libpython3-stdlib_unix
+      - python3-minimal_bins
```
cjdcordeiro, line 42: It is more maintainable if you replace `libpython3-stdlib_core` and `libpython3-stdlib_core` by `python3_core`

cjdcordeiro (changes requested):

slices/python3-minimal.yaml
linostar: debian_defaults needed by py3versions. The others are needed by py3compile/py3clean. Added more slices for those.

slices/python3-minimal.yaml
linostar: Added more slices.

slices/libpython3-stdlib.yaml
linostar: Added.

slices/python3-minimal.yaml
cjdcordeiro: Thanks.

We need to be consistent about `bins`.

As per other slices, `bins` typically deliver all binaries.

In this case, we want the `/usr/bin/python3` symlink, so we should have a dedicated slice for it.

You've created dedicated slices for the other bins too, which is ok.

cjdcordeiro (changes requested):
Need to address https://github.com/canonical/chisel-releases/pull/387#discussion_r1853655947 and the failing tests

cjdcordeiro (approved):
thanks for the changes. lgtm

PR #258, opened 2024-06-13
feat(24.04): add rabbitmq slices

linostar:
# Proposed changes
Add `rabbitmq-server` and `librabbitmq4` slices to 24.04.

Note that erlang* and logrotate packages are not available for i386 arch, so they have been commented out.

## Related issues/PRs
N/A

### Forward porting
N/A

## Testing

```bash
chisel cut --release ./ --root rootfs rabbitmq-server_bins                                                                                                                                        

2024/06/13 11:27:35 Processing ./ release...
2024/06/13 11:27:36 Selecting slices...
2024/06/13 11:27:36 Fetching ubuntu 24.04 noble suite details...
2024/06/13 11:27:38 Release date: Thu, 25 Apr 2024 15:10:33 UTC
2024/06/13 11:27:38 Fetching index for ubuntu 24.04 noble main component...
2024/06/13 11:27:38 Fetching index for ubuntu 24.04 noble universe component...
2024/06/13 11:27:38 Fetching ubuntu 24.04 noble-security suite details...
2024/06/13 11:27:39 Release date: Wed, 12 Jun 2024 13:31:47 UTC
2024/06/13 11:27:39 Fetching index for ubuntu 24.04 noble-security main component...
2024/06/13 11:27:39 Fetching index for ubuntu 24.04 noble-security universe component...
2024/06/13 11:27:39 Fetching ubuntu 24.04 noble-updates suite details...
2024/06/13 11:27:39 Release date: Thu, 13 Jun 2024  7:51:49 UTC
2024/06/13 11:27:39 Fetching index for ubuntu 24.04 noble-updates main component...
2024/06/13 11:27:39 Fetching index for ubuntu 24.04 noble-updates universe component...
2024/06/13 11:27:39 Fetching pool/main/a/adduser/adduser_3.137ubuntu1_all.deb...
2024/06/13 11:27:39 Fetching pool/main/a/audit/libaudit-common_3.1.2-2.1build1_all.deb...
2024/06/13 11:27:39 Fetching pool/main/g/glibc/libc6_2.39-0ubuntu8.2_amd64.deb...
2024/06/13 11:27:39 Fetching pool/main/libc/libcap-ng/libcap-ng0_0.8.4-2build2_amd64.deb...
2024/06/13 11:27:39 Fetching pool/main/a/audit/libaudit1_3.1.2-2.1build1_amd64.deb...
[...]

PR #384, opened 2024-11-08
feat(24.04) add erlang slices

linostar:
# Proposed changes

Add erlang-* slices to 24.04

Note that erlang packages are not available for i386 arch.

## Related issues/PRs

PR #258

## Forward porting

N/A\n\n

## Checklist

* [x] I have read the [contributing guidelines](\nhttps://github.com/canonical/chisel-releases/blob/main/CONTRIBUTING.md)
* [x] I have tested my changes ([see how](https://github.com/canonical/chisel-releases/blob/main/CONTRIBUTING.md#7-test-your-slices-before-opening-a-pr))
* [x] I have already submitted the [CLA form](\nhttps://ubuntu.com/legal/contributors/agreement)

## Additional Context

N/A

PR #394, opened 2024-11-21
feat(24.04): add erlang slices

linostar:
# Proposed changes

Add erlang-* slices to 24.04

Note that erlang packages are not available for i386 arch.

I tried to add tests for all erlang binaries , but some of them needs the erl server to be running in the background, and attempting to run it with `&` or with `nohup` has failed in the spread container (although it works locally).

## Related issues/PRs

PR #258

## Forward porting

N/A

## Checklist

* [x] I have read the [contributing guidelines](\nhttps://github.com/canonical/chisel-releases/blob/main/CONTRIBUTING.md)
* [x] I have already submitted the [CLA form](\nhttps://ubuntu.com/legal/contributors/agreement)

## Additional Context

N/A


cjdcordeiro (changes requested):
Following up from https://github.com/canonical/chisel-releases/pull/384#discussion_r1840788785

> bash is needed because there are several erlang packages that have shell scripts, and they all depend on erlang-base, so it makes sense to put bash as dependency in erlang-base.
> coreutils is needed because some bash scripts use /usr/bin/env to detect the path of escript, and env is found in coreutils
> as for libtinfo8t64, I found that some of the binaries in erlnag-snmp are trying to open libtinfo so file (from strace output), and I am not sure if there are others that use it as well, so I put it in erlang-base's essential as a precaution.

What are the several erlang packages that have shell scripts? If those are other than the `erlang-base`, then those should be the ones carrying this `bash` dependency. 

For `coreutils`, same as for `bash`. Additionally, if all you need is `env`, then let's not bring in all the binaries.

And `libtinfo8t64`, adding it just as a precaution isn't a good practice cause we may be unnecessarily bloating the slice, and once introduced, it's not something we can delete (not at least until the following Ubuntu release).

slices/coreutils.yaml
```diff
@@ -129,6 +129,12 @@ slices:
       /usr/bin/yes:
       /usr/sbin/chroot:
 
+  env:
```
cjdcordeiro, line 132: since this slice is being introduced and it's a binary, can you please add a test for it?

cjdcordeiro (changes requested):
LGTM. just missing a test and that would be it

zhijie-yang (approved):
LGTM, thanks!

PR #339, opened 2024-09-04
feat(24.10): forward-port 24.04 slices

rebornplusplus:
This PR forward-ports the 24.04 slices into the new 24.10 branch.

Highlights:
```git
55c2930 feat: Start ubuntu-24.10
8679103 feat: import slices from 24.04
f80c55f test: import and edit spread tests from 24.04
a6b8fb2 chore: use appropriate spread backends (oracular)
51d3e4e fix: replace libnode109 with libnode115
566b31d fix: remove librav1e0: does not exist
f4230c6 fix: remove libsvtav1enc1d1: does not exist
464de24 fix: remove libunibreak5: does not exist
97b9553 fix: remove libx265-199: does not exist
57ec378 fix: incorrect path in libcryptsetup12
5d0b487 fix: incorrect path in libelf1t64
0570f30 fix: remove missing path in passwd
102a953 fix: remove missing paths in sensible-utils_extras
b3fc7f3 fix: remove systemd slice definition file
39e7518 fix: remove missing paths in procps
ddfe33d fix: remove deprecated slice from ca-certificates
de02183 fix: add new deps in libssl3t64
d83b934 fix: remove lsb-base dependency from apparmor
9850b74 fix: add and remove deps in coreutils
fcb5308 fix: add missing deps in httping
815d7d0 feat: add libssh2-1t64 slices
4cdb7ba fix: replace old deps in libcurl4t64
4fbd1bf feat: add libldap-common slices
2397d0c fix: add new deps in libldap2
c7a977b feat: add libxxhash0 slices
197b5b8 fix: add missing deps in liblz4-1
d6e7de9 fix: add missing deps in libpam-modules
8b49ce9 fix: add missing paths and deps in libpython3.12-stdlib
96c7ae0 fix: remove unnecessary deps from libsystemd-shared
42261cc fix: remove unnecessary deps from libsystemd0
7e94b1d fix: use correct dep in libtiff6
87f1a48 fix: add missing deps in passwd
37075e3 fix: replace arch-specific keywords
7c99b2e fix: remove unnecessary deps from libcryptsetup12
```

<details>
<summary> Changes from 24.04 </summary>

```git
commit 7c99b2e187432c720787c0f0ba97cae9316c49f9

    fix: remove unnecessary deps from libcryptsetup12

diff --git a/slices/libcryptsetup12.yaml b/slices/libcryptsetup12.yaml
index f88d32e..8a48ad0 100644
--- a/slices/libcryptsetup12.yaml
[...]

PR #395, opened 2024-11-21
fix(24.04/libnss3): libs should contain all libraries

### LINKED ISSUE #381: libnss3 does not contain libsmime3.so

rebornplusplus:
This PR restructures the libnss3 slices. It renames the former "libs" slice to "nss", to only include nss3-specific libraries. It also adds a new slice "smime" to include S/MIME libraries.

Finally, the "libs" slice is added back, which includes all of the slices that provide libraries. The rationale is that "libs" slices should contain all libraries coming from the package.

Note that the openjdk-*-jre-headless slices have also been updated to reflect this change.

Resolves #381

## Forward-porting
- #382.

### DIFF
--- a/slices/libnss3.yaml
+++ b/slices/libnss3.yaml
@@ -1,10 +1,25 @@
+# Network Security Service libraries.
+#
+# This is a set of libraries designed to support cross-platform development of
+# security-enabled client and server applications. It can support SSLv2 and v4,
+# TLS, PKCS #5, #7, #11, #12, S/MIME, X.509 v3 certificates and other security
+# standards.
 package: libnss3
 
 essential:
   - libnss3_copyright
 
 slices:
+  # The libs slice contains all the libraries coming from this package.
   libs:
+    essential:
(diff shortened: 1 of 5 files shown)


linostar (approved):
LGTM, thanks.

The spread tests are failing because they are trying to install and test multiple java versions in parallel on the same worker.

cjdcordeiro (changes requested):
needs the fix for the java tests to be backported from ~~#339~~ #395

PR #402, opened 2024-11-28
fix(sbsigntool): missing top-level essential

cjdcordeiro:
Fix https://github.com/canonical/chisel-releases/actions/runs/12060324342/job/33630542601


linostar (approved):
LGTM.

PR #467, opened 2025-01-22
feat(24.04): add SDF for pwgen

eunufe:
# Proposed changes

feat(24.04): add SDF for pwgen
### Forward porting

#476 

## Checklist

### DIFF
--- a/tests/spread/integration/pwgen/task.yaml
+++ b/tests/spread/integration/pwgen/task.yaml
@@ -1,0 +1,14 @@
+summary: Integration tests for pwgen
+
+execute: |
+  rootfs="$(install-slices pwgen_bins)"
+
+  chroot "${rootfs}/" pwgen -h
+  # pwgen requires /dev/urandom & /dev/random
+  mkdir "${rootfs}/dev"
+  mount --bind /dev "${rootfs/dev}"
+
+  chroot "${rootfs}/" pwgen -h
+  test $(chroot "${rootfs}/" pwgen -C1 12 | wc -m) -eq 13
+
+  umount "${rootfs}/dev" 
--- a/tests/spread/integration/pwgen/task.yaml
+++ b/tests/spread/integration/pwgen/task.yaml
@@ -3,12 +3,9 @@ summary: Integration tests for pwgen
 execute: |
   rootfs="$(install-slices pwgen_bins)"
 
-  chroot "${rootfs}/" pwgen -h
   # pwgen requires /dev/urandom & /dev/random
   mkdir "${rootfs}/dev"
   mount --bind /dev "${rootfs/dev}"
-
-  chroot "${rootfs}/" pwgen -h
   test $(chroot "${rootfs}/" pwgen -C1 12 | wc -m) -eq 13
 
   umount "${rootfs}/dev" 
--- a/tests/spread/integration/pwgen/task.yaml
+++ b/tests/spread/integration/pwgen/task.yaml
@@ -1,7 +1,7 @@
 summary: Integration tests for pwgen
 
 execute: |
-  rootfs="$(install-slices pwgen_bins)"
+  rootfs="$(install-slices base-files_base pwgen_bins)"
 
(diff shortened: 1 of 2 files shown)


tests/spread/integration/pwgen/task.yaml
```diff
@@ -0,0 +1,11 @@
+summary: Integration tests for pwgen
+
+execute: |
+  rootfs="$(install-slices pwgen_bins)"
```
rebornplusplus, line 4:
```suggestion
  rootfs="$(install-slices base-files_base pwgen_bins)"
```

I suspect the spread tests are failing because of this. Maybe `libc6_libs` should come up with `base-files_base` or `base-files_lib`. Anyway, until that happens, let's please install `base-files_base` too.

rebornplusplus (commented):
The slices look good to me. The spread test is failing though. I have a guess below why.

tests/spread/integration/pwgen/task.yaml
```diff
@@ -0,0 +1,11 @@
+summary: Integration tests for pwgen
+
+execute: |
+  rootfs="$(install-slices base-files_base pwgen_bins)"
+
+  # pwgen requires /dev/urandom & /dev/random
+  mkdir "${rootfs}/dev"
+  mount --bind /dev "${rootfs/dev}"
```
zhijie-yang, line 8:
```suggestion
  mount --rbind /dev "${rootfs}/dev"
```

tests/spread/integration/pwgen/task.yaml
```diff
@@ -0,0 +1,11 @@
...
+  test $(chroot "${rootfs}/" pwgen -C1 12 | wc -m) -eq 13
+
+  umount "${rootfs}/dev"
```
zhijie-yang, line 11:
```suggestion
  umount -l "${rootfs}/dev" 
```

zhijie-yang (changes requested):
You need either: 1. use `--rbind` to mount `/dev` to make `/dev/urandom` work, or 2. read and copy some random bytes long enough to the `/dev/urandom` to fake the entropy. Here I would suggest use an `--rbind` to avoid hacks.

tests/spread/integration/pwgen/task.yaml
norrisjeremy: Hi @zhijie-yang,

I'm not sure I understand why it needs to use `--rbind` instead of `--bind`?
Other spread tests are successfully using `--bind` current.
See:
- dbus
- perl-base
- python3-pyftpdlib

Thanks,
Jeremy

tests/spread/integration/pwgen/task.yaml
norrisjeremy: Hi @zhijie-yang,

I'm not sure I understand why it needs to use `umount -l`?
Other spread tests are successfully using `umount` without the `-l` argument.
See:
- systemd
- dbus

Thanks,
Jeremy

tests/spread/integration/pwgen/task.yaml
zhijie-yang: Hi Norris,

This is because `/dev/urandom` in LXD is essentially a mount instead of a direct hardware mapping. This can be verified by running `lxc launch ubuntu:24.04 noble && lxc shell noble` then `mount | grep urandom`, from which you will get
```
udev on /dev/urandom type devtmpfs (rw,nosuid,relatime,size=16216556k,nr_inodes=4054139,mode=755,inode64)
```
 
This would effectively require a `--rbind`.

tests/spread/integration/pwgen/task.yaml
zhijie-yang: Hi Norris,

Using the `-l` avoids the `umount` failing with `target is busy` error, when other processes are preventing the `--rbind` mount from being unmounted.

zhijie-yang (approved):
Looks good to me. Thanks!

rebornplusplus (approved):
Looks good to me, thanks for the changes!

PR #477, opened 2025-01-28
feat(24.04): add python3-pytest slices

rebornplusplus:
# Proposed changes
Add the slices for python3-pytest and dependencies.

### Forward porting

 - TODO for 24.10

## Checklist


slices/python3-importlib-metadata.yaml
```diff
@@ -0,0 +1,18 @@
+package: python3-importlib-metadata
+
+essential:
+  - python3-importlib-metadata_copyright
+
+slices:
+  libs:
+    essential:
+      - python3-typing-extensions_libs
```
clay-lake, line 9: Looks like python3-typing-extensions is only required for python >> 3.8. Should we remove it? I think this might only apply to Jammy, w/o LTS updates.

clay-lake (commented):
Looks good, but I added a few comments. Lets review again in our 1:1

slices/python3-importlib-metadata.yaml
rebornplusplus: That's a great point. We might just not need it.

I just checked the `/usr/lib/python3/dist-packages/typing_extensions.py` file and found this line which might indicate that we should be fine by not including the extensions slice.

```py
if sys.version_info >= (3, 11):
    from typing import Any
```

Let's talk more about this in our 1:1!

rebornplusplus: Let's just ditch the slice since noble will have python >> 3.8 anyway. And add another spread test which imports `typing` and does something with it.

slices/python3-typing-extensions.yaml
```diff
@@ -0,0 +1,16 @@
...
+
+essential:
+  - python3-typing-extensions_copyright
+
+slices:
+  libs:
+    essential:
+      - python3_standard
+    contents:
+      /usr/lib/python3/dist-packages/typing_extensions-*.dist-info/**:
```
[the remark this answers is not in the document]
cjdcordeiro: Btw, I think it's safe to keep `**` here because it is dedicated to `typing_extensions-*`, but IMO this thinking - "future-proof" - shouldn't be the general rule here. 

Because of conflicts and lazy globs, we should advocate for conservative globs rather than "just in case" globs, cause the latter can backfire later.

cjdcordeiro (changes requested):
Thanks for this. Since all the python libs depend on python3, can you actually add a test for each one of them, where you simply `import` the corresponding libs and do something very basic boilerplate call as a test?

PR #331, opened 2024-09-03
feat(24.04): add gzip

Meulengracht:
# Proposed changes

Propose tools gzip and tar. I did not propose tar alone, as tar invokes gzip as the underlying compression tool.

### Forward porting

#483 

## Checklist

* [x] I have read the [contributing guidelines](
https://github.com/canonical/chisel-releases/blob/main/CONTRIBUTING.md)
* [x] I have tested my changes ([see how](https://github.com/canonical/chisel-releases/blob/main/CONTRIBUTING.md#7-test-your-slices-before-opening-a-pr))
* [x] I have already submitted the [CLA form](
https://ubuntu.com/legal/contributors/agreement)

PR #484, opened 2025-02-05
feat(24.04): add less

zhijie-yang:
# Proposed changes

Add `less`

## Related issues/PRs

Dependent by #331 

### Forward porting

Will open once approved.

## Checklist

### DIFF
--- a/slices/less.yaml
+++ b/slices/less.yaml
@@ -13,7 +13,7 @@ slices:
       /usr/bin/lessecho:
       # Note that lesskey is deprecated since less version 582.
       /usr/bin/lesskey:
-  
+
   # pager conflicts with the one generated in util-linux_generated slice.
   # Manually link pager to less if required.
   # TODO: Add `prefer` to /usr/bin/pager when conflict resolution
--- a/slices/less.yaml
+++ b/slices/less.yaml
@@ -11,19 +11,11 @@ slices:
     contents:
       /usr/bin/less:
       /usr/bin/lessecho:
-      # lesskey is deprecated since version 582.
+      # Note that lesskey is deprecated since less version 582.
+      /usr/bin/lesskey:
 
-  # scripts slice is not added since `lessfile`
-  # is a script contains unlisted dependencies
-  # making it difficult to test thoroughly.
-  # scripts:
-  #   essential:
-  #     - less_bins
-  #   contents:
-  #     /usr/bin/lessfile:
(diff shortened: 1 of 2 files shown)


slices/less.yaml
```diff
@@ -0,0 +1,32 @@
...
+  bins:
+    essential:
+      - libc6_libs
+      - libtinfo6_libs
+    contents:
+      /usr/bin/less:
+      /usr/bin/lessecho:
+      # lesskey is deprecated since version 582.
+
+  scripts:
```
clay-lake, line 16: Should we add this if we are not including/testing all dependencies? Its a very long script

clay-lake (changes requested):
Looks good except the scripts slice.

norrisjeremy:
Hi @zhijie-yang,

We also put together PRs independently to add less to 24.04 & 24.10 (see #486 & #487).
Our versions also include more extensive integration tests, and also includes SDFs for terminfo data needed to make `less` actually functional.
How would you like to proceed?

Thanks,
Jeremy

zhijie-yang:
Hi @norrisjeremy ,

Yes, we are pushing this for Ubuntu 24.04 now. However, for `less`, the scripts in the package contain many unlisted dependencies (if you read the script), which makes writing the tests cumbersome. We are leaning a bit towards holding the `_scripts` slices back as comments since they are just "scripts" combined with other commands, and the binaries are still the major components of this package. If you have a specific use case for the scripts, please point it out, and we can find a way to add proper tests for them. There are some reviews pending for this PR now. Once this once gets merged, we can continues on the other slices that depends on `less` (e.g., `gzip`.)

norrisjeremy:
Hi @zhijie-yang,

We don't feel strongly about the `scripts` slice, so those can be left out.
However, have you actually tested that the `less` utility itself works?
We found that we needed to create an SDF for `ncurses-base` to get basic terminfo definitions in place in order for it to actually function (which we also included in our PRs).
We also constructed a simple integration test for `less` itself.
I'm not sure if you want to incorporate those pieces from our PRs?

Thanks,
Jeremy

rebornplusplus (approved):
Looks good to me.

Let's merge this first, and @norrisjeremy @eunufe, if you guys want to include the `scripts` slices in #486, then we might need to add some more package slices. Because, the script invokes quite a number of binaries not from this package. Let us know what you think!

rebornplusplus (approved):

PR #485, opened 2025-02-05
feat(24.04): rework rabbitmq 24.04

zhijie-yang:
# Proposed changes

This is a rework of #258 to reduce the rebase overhead. The original author of these files is @linostar.

## Related issues/PRs
#258 

### Forward porting
#514 

## Checklist


slices/rabbitmq-server.yaml
```diff
@@ -0,0 +1,106 @@
...
+      - base-files_etc
+      - base-files_home
+      - base-files_var
+      - base-passwd_data
+      - bash_bins
+      - coreutils_bins
+      # Uncomment the following lines if not on i386
+      # - erlang-base_bins
+      # - erlang-inets_scripts
+      # - erlang-os-mon_bins
```
cjdcordeiro, line 25: it's been a long time since I last reviewed those, but, wouldn't you need to add these in your test? You're only installing `erlang-base_bins`, and it seems to work (?)

zhijie-yang: The other `erlang-*` slices are for i386. For platforms other than i386, they work without installing these slices.

cjdcordeiro: but then maybe I'm misunderstanding the comment (which btw is a double negative so it should be cleaned up for readability)
 - **uncomment**, if arch != i386 - IoW install these slices if on amd64.

zhijie-yang: After some investigation, it's me having the misunderstandings. 

The `rabbitmq-server` relies on `erlang-*` to work. However, `erlang-*` are not available on `i386`. Now we are commenting them out to pass the CI for i386, but in the real use cases, we still need them. 

In other words: `rabbitmq-server` does not work for `i386` from the perspective of the Ubuntu Archives.

zhijie-yang:
I have a feeling that, if we skew the actual dependencies for the platforms other than `i386` just for passing the CI, we are actually departing from the purpose of having CI. 

I suggest that we uncomment all the dependencies that are currently commented out, add comments at the top of the slice definitions, indicating that this slice does not work for i386, and let CI just fail on this before chisel supports marking unsupported slices in SDF. We can override the failed CI checks upon merging, as the failing installability test for `i386` is expected.

Also, the installability test workflow for [`erlang—*`](https://github.com/canonical/chisel-releases/actions/runs/11971661668/job/33376954823) magically succeeded. We can't access the logs anymore, but it's definitely worth investigatingn.

cjdcordeiro (changes requested):
Unfortunately, as it is now, we'll start getting CI errors on a daily basis (we install all slices daily to check to upstream regressions).

Although less intuitive and potentially more impactful, given that we'll be working on arch-specific slices/deps in the future, I'd suggest the opposite approach: let's comment it, with a `# NOTE`, an on installation+spread tests, ensure that those commented essentials are explicitly added to the `cut`

cjdcordeiro:
Also, please create the FPorting PR already, even if it needs changes later, so that we don't forget about it

zhijie-yang:
Updated. The `/usr/bin/login` binary inside package `login` seems not testable. It fails consistently when run in spread. However, the same code works in a `spread -debug` session without any errors.

tests/spread/integration/rabbitmq-server/task.yaml
```diff
@@ -1,7 +1,7 @@
 summary: Integration tests for RabbitMQ
 
 execute: |
-  rootfs="$(install-slices rabbitmq-server_bins)"
+  rootfs="$(install-slices rabbitmq-server_bins erlang-base_bins)"
```
cjdcordeiro, line 4: what about the rabbitmq-server_modules and erlang-inets_scripts, erlang-os-mon_bins?

zhijie-yang: They are not necessary to run the most fundamental test case, i.e., the `rabbitmqctl version`. These erlang packages are required by the further modules/plugins in rabbitmq.

cjdcordeiro (commented):
@zhijie-yang check my proposed structure at ~~https://github.com/canonical/chisel-releases/pull/485/commits/b711494fa6837ab56e8837323db640cf0abc2817~~ https://github.com/canonical/chisel-releases/commit/b6cd285b650cedc92e6c87b30c02853d47a724c0. wdyt?

(not sure if the tests will pass yet...passed locally, but TBC with Spread)

cjdcordeiro (approved):

clay-lake (approved):
LGTM! Thanks

PR #486, opened 2025-02-07
feat(24.04): add SDF for ncurses-base and zstd

eunufe:
# Proposed changes

feat(24.04): add SDF for ncurses-base, less and zstd
### Forward porting

#487 
## Checklist


rebornplusplus:
Hi! We merged #484. Please see https://github.com/canonical/chisel-releases/pull/484#pullrequestreview-2615032393. We cherry-picked the tests from here. 

Regarding https://github.com/canonical/chisel-releases/pull/484#issuecomment-2652126241, we just tested that `less` does work but it prints out a warning message.
So what we can do in this PR, is just add a comment over `less_bins` suggesting the user to include `ncurses-base_terminfo` for a better experience.

Let us know what you think!

slices/zstd.yaml
```diff
@@ -0,0 +1,31 @@
...
+      - libstdc++6_libs
+      - zlib1g_libs
+    contents:
+      /usr/bin/pzstd:
+      /usr/bin/unzstd: {symlink: /usr/bin/zstd}
+      /usr/bin/zstd:
+      /usr/bin/zstdcat: {symlink: /usr/bin/zstd}
+      /usr/bin/zstdmt: {symlink: /usr/bin/zstd}
+
+  scripts:
```
cjdcordeiro, line 22: I understand the logic here, and I'm not opposed to this call.

But following the logic from other existing slices, I'd prefer these two scripts to (also) be in `bins`, given that they are located in `/usr/bin/`.

tests/spread/integration/zstd/task.yaml
```diff
@@ -0,0 +1,26 @@
...
+execute: |
+  rootfs="$(install-slices base-files_bin bash_bins dash_bins grep_bins ncurses-base_terminfo \
+    zstd_bins less_bins zstd_scripts)"
+  
+  chroot "${rootfs}/" pzstd -h
+  chroot "${rootfs}/" unzstd -H
+  chroot "${rootfs}/" zstd -H
+  chroot "${rootfs}/" zstdcat -H
+  chroot "${rootfs}/" zstdless --version
+  chroot "${rootfs}/" zstdmt -H
```
cjdcordeiro, line 12: can you also please add a smoke test this this? it can be similar to `zstd`

cjdcordeiro (changes requested):

slices/zstd.yaml
norrisjeremy: Hi @cjdcordeiro,

That doesn't seem to be the logic we see in other slices?
For example, in `xz-utils` & `unzip`, the shell scripts are kept in a separate `scripts` slice that is not essential for the `bins` slice.
And frankly, we would rather the `scripts` slice not be `essential` for the `bins`, as it pulls in additional cruft that isn't strictly needed for the `bins` to function.

norrisjeremy: Hi @cjdcordeiro,

I see now that the `xz-utils_scripts` slice includes `essentials` for all the utilities it requires that aren't actually declared as dependencies in the `.deb` package (except `dash_bins` for some reason?).
Would you like us to follow suit here (something like this)?
```diff
diff --git a/slices/zstd.yaml b/slices/zstd.yaml
index 716b05b..d9cc182 100644
--- a/slices/zstd.yaml
+++ b/slices/zstd.yaml
@@ -21,6 +21,12 @@ slices:
 
   scripts:
     essential:
+      # shebang in scripts reference interpreter in /bin not /usr/bin
+      - base-files_bin
+      - bash_bins  # needed for zstdless
+      - dash_bins  # needed for zstdgrep and zstdless
+      - grep_bins  # needed for zstdgrep
+      - less_bins  # needed for zstdless
       - zstd_bins
     contents:
       /usr/bin/zstdgrep:
```

slices/zstd.yaml
```diff
@@ -0,0 +1,31 @@
...
+      - liblz4-1_libs
+      - liblzma5_libs
+      - libstdc++6_libs
+      - zlib1g_libs
+    contents:
+      /usr/bin/pzstd:
+      /usr/bin/unzstd: {symlink: /usr/bin/zstd}
+      /usr/bin/zstd:
+      /usr/bin/zstdcat: {symlink: /usr/bin/zstd}
+      /usr/bin/zstdmt: {symlink: /usr/bin/zstd}
```
norrisjeremy, line 20: Another question we have: we see in some other slice definitions that some files that are symlinks are just listed directly in the contents without the `symlink:` attribute being marked in the SDF.

For example, in xz-utils.yaml, /usr/bin/unxz, /usr/bin/xzcat, /usr/bin/xzcmp, /usr/bin/xzegrep & /usr/bin/xzfgrep are all symlinks pointing to either xz, xzdiff or xzgrep, but are not marked with the `symlink:` attribute in the SDF.

Should we follow suit here and drop the `symlink:` attribute in this SDF for zstd for /usr/bin/unzstd, /usr/bin/zstdcat & /usr/bin/zstdmt?

slices/zstd.yaml
cjdcordeiro: Hi @norrisjeremy . Yes for the latter please, makes sense.

As for the `scripts` slice name, that is true - there are in fact some precedents in this release. 
Now that we have [docs](https://documentation.ubuntu.com/chisel) for Chisel, we shall soon come up with a comprehensive slicing guide that shall address these slice design decisions. Until then, given that there's a precedent in this release, and that it's easier to "add" than it is to "remove", let's keep it as you propose ;)

slices/zstd.yaml
cjdcordeiro: Yes, you can drop `symlink` when the file is already a `symlink` in the deb.

In short:
 - the deb contents will be extracted while preserving their type, mode and permissions,
 - `symlink` can be used to explicitly create a symlink that doesn't exist in the deb's contents (this is useful for reproducing the outcomes of maintainer scripts)

slices/zstd.yaml
norrisjeremy: Hi @cjdcordeiro,

Ok, just to clarify, we should retain the `bins` & `scripts` slices as we proposed?
And we should go ahead and add `essentials` for the `scripts` slice as follows?
```diff
diff --git a/slices/zstd.yaml b/slices/zstd.yaml
index 716b05b..9bf2d31 100644
--- a/slices/zstd.yaml
+++ b/slices/zstd.yaml
@@ -21,6 +21,11 @@ slices:
 
   scripts:
     essential:
+      # shebang in scripts reference interpreter in /bin not /usr/bin
+      - base-files_bin
+      - bash_bins  # needed for zstdless
+      - grep_bins  # needed for zstdgrep
+      - less_bins  # needed for zstdless
       - zstd_bins
     contents:
       /usr/bin/zstdgrep:
```

tests/spread/integration/zstd/task.yaml
norrisjeremy: Hi @cjdcordeiro,

To clarify, you are referring to `zstdmt`?

slices/zstd.yaml
norrisjeremy: Hi @cjdcordeiro,

Great, thank you for clarifying! We'll drop the `symlink:` attribute here in this case.

slices/zstd.yaml
cjdcordeiro: Yes for both :)

tests/spread/integration/zstd/task.yaml
cjdcordeiro: Yes. There's only the `-H` test, but we could also test some real functionality since it should provide a comparable output to `zstd`

clay-lake (approved):
LGTM. We had some issues with CI, but I re-triggered the workflow it should be good now.
@cjdcordeiro is OOO for the next week, but I'll discuss what we can do to merge this in.

zhijie-yang (approved):
LGTM, thanks!

PR #475, opened 2025-01-26
feat(24.10): Add SDF for `libgl1` and dependencies

HadrienPatte:
# Proposed changes

Add SDF for `libgl1` and dependencies.

See https://github.com/canonical/chisel/issues/93 for the dependency diff.

## Related issues/PRs

24.04 PR: #474 (dropped)
24.04 PR: #508 (current, back-ported)

### Forward porting
N/A

## Checklist

* [x] I have read the [contributing guidelines](
https://github.com/canonical/chisel-releases/blob/main/CONTRIBUTING.md)
* [x] I have tested my changes ([see how](https://github.com/canonical/chisel-releases/blob/main/CONTRIBUTING.md#7-test-your-slices-before-opening-a-pr))
* [x] I have already submitted the [CLA form](
https://ubuntu.com/legal/contributors/agreement)

PR #508, opened 2025-02-28
feat(24.04): slices for libgl1

clay-lake:
# Proposed changes
This PR backports the changes for `libgl1` from #475. There is an existing PR for noble, however this PR is consistent with the dependencies from noble-updates archive which differ significantly from the noble archive alone.

### DIFF
--- a/slices/libgbm1.yaml
+++ b/slices/libgbm1.yaml
@@ -1,0 +1,21 @@
+package: libgbm1
+
+essential:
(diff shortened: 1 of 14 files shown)


zhijie-yang (approved):
Looks good to me. Thanks!

HadrienPatte (approved):
Thanks for the port, LGTM :+1:

PR #305, opened 2024-08-02
feat(24.04) add adduser slice

linostar:
# Proposed changes

Adds `adduser` slice to 24.04.

## Related issues/PRs

N/A

### Forward porting

N/A

## Checklist

* [x] I have read the [contributing guidelines](
https://github.com/canonical/chisel-releases/blob/main/CONTRIBUTING.md)
* [x] I have tested my changes ([see how](https://github.com/canonical/chisel-releases/blob/main/CONTRIBUTING.md#7-test-your-slices-before-opening-a-pr))
* [x] I have already submitted the [CLA form](
https://ubuntu.com/legal/contributors/agreement)

PR #511, opened 2025-03-11
feat(24.04): add influxdb slice

alesancor1:
# Proposed changes
Slice definition for influxdb `1.6.7~rc0-2ubuntu0.24.04.2` located in apt **universe**.

## Checklist

## Additional Context

This is an older version of influxdb (v1). Influxdb is currently in v2, which is available in [InfluxData package repository](https://repos.influxdata.com/debian).


slices/influxdb.yaml
```diff
@@ -0,0 +1,28 @@
...
+    # influxdb actually depends on adduser as well to create and remove
+    # a user during package install/removal - however we don't support this
+    # currently so adduser is not added here.
+    essential:
+      - influxdb_config
+      - libc6_libs
+      - lsb-base_bins
+    contents:
+      /usr/bin/influxd:
+      /usr/lib/systemd/system/influxdb.service:
```
cjdcordeiro, line 17: This isn't a bin but rather a systemd service file.

rule of thumb: if you don't need it for your use case, don't slice it. it will be hard to test and, if wrong, impossible to remove (we don't remove `contents` (due to regressions) but can always add more).

slices/influxdb.yaml
```diff
@@ -0,0 +1,28 @@
...
+      /var/lib/influxdb/:
+      /var/log/influxdb/:
```
cjdcordeiro, line 19: same here, not bins

slices/influxdb.yaml
```diff
@@ -0,0 +1,28 @@
...
+  config:
+    contents:
+      /etc/influxdb/influxdb.conf:
+      /etc/init.d/influxdb:
```
cjdcordeiro, line 23: this is not config

slices/influxdb.yaml
```diff
@@ -0,0 +1,28 @@
...
+      /etc/logrotate.d/influxdb:
```
cjdcordeiro, line 24: might be better to have a dedicated "logging" slice for this and `/var/log/influxdb`

slices/influxdb.yaml
```diff
@@ -0,0 +1,28 @@
+package: influxdb
+
+essential:
+  - influxdb_copyright
+
+slices:
+  bins:
+    # influxdb actually depends on adduser as well to create and remove
```
cjdcordeiro, line 8: well, the postinst uses it right? not influxdb at runtime? if the latter, this will help: https://github.com/canonical/chisel-releases/pull/305/files

There's also more stuff in the maintainer scripts that you can enforce in this SDF. E.g. there's a `/var/run/influxdb` dir.

cjdcordeiro (changes requested):
Thanks for this. It needs some love - left a few comments below. In general, more is less - try to be conservative when slicing and only declare things that you need for your use case (e.g. leave systemd stuff out for now)

slices/influxdb.yaml
alesancor1: addressed in https://github.com/canonical/chisel-releases/pull/511/commits/b84135990590eb63eef0b2c95101865aac5cc53f - removed systemd related files

slices/influxdb.yaml
alesancor1: addressed in https://github.com/canonical/chisel-releases/pull/511/commits/b84135990590eb63eef0b2c95101865aac5cc53f 

* moved `/var/log/influxdb` to logging slice
* moved `/var/lib/influxdb` to data slice

NOTE: influx creates /var/lib/influxdb on startup - should still be on the SDF? I included it since it's part of the debian pkg and this way we dont need to `mkdir` it in case we want to have `influxdb` as a user and owner of that dir

slices/influxdb.yaml
alesancor1: Addressed in https://github.com/canonical/chisel-releases/pull/511/commits/b84135990590eb63eef0b2c95101865aac5cc53f - is used for the systemd service so it has been removed

slices/influxdb.yaml
alesancor1: Addressed in https://github.com/canonical/chisel-releases/pull/511/commits/b84135990590eb63eef0b2c95101865aac5cc53f - added a logging slice

cjdcordeiro (changes requested):

slices/influxdb.yaml
```diff
@@ -14,14 +14,19 @@ slices:
       - lsb-base_bins
     contents:
       /usr/bin/influxd:
```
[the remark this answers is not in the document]
alesancor1: Yes and no - it does depend on it but actually creates the contents of `data` if they don't exist. I assume we'd want to include it, since technically is required for bins to work.

slices/influxdb.yaml
alesancor1: `bins` don't really require anything else, it is able to run as it is currently. It creates `/var/lib/influxdb/` if doesn't exist. I'd be more inclined to take the first approach so we would have:

```yaml
bins:
  essential:
    - influxdb_config
    - influxdb_vars
   ...

vars:
  contents:
       /var/log/influxdb/:
       /var/lib/influxdb/:

config:
  contents:
    /etc/influxdb/influxdb.conf:
    /etc/logrotate.d/influxdb:
```

This means that `logging` dissapears and goes to `vars` and `configs`, then `bins` requires both of them. 

Semantically speaking, this would be the one which makes most sense, as /var/lib/influxdb and /var/log/ingluxdb would be created runtime at some point.

cjdcordeiro: ok go for it

slices/influxdb.yaml
```diff
@@ -0,0 +1,33 @@
...
+
+slices:
+  bins:
+    # influxdb actually depends on adduser as well to create and remove
+    # a user during package install/removal - however we don't support this
+    # currently so adduser is not added here.
+    essential:
+      - influxdb_config
+      - libc6_libs
+      - lsb-base_bins
```
clay-lake, line 14: I think `lsb-base` is only needed for `/etc/init.d/influxdb` but if we do not include init script we could probably drop this slice.

clay-lake (changes requested):
Looks good! Just had one question about `lsb-base`

slices/influxdb.yaml
alesancor1: Good shout, tested without it and works perfectly. I removed `lsb-base` in https://github.com/canonical/chisel-releases/pull/511/commits/67ee0889d98aca99111c6697f9e054088df3d572

cjdcordeiro (approved):
lgtm thanks

clay-lake (approved):
Looks good!

PR #526, opened 2025-04-11
Fix: Inject correct ref for testing feature branches

clay-lake:
# Included in this PR:
- When testing specific branches, swap the ref in the RELEASE JSON with the current feature branch to be merged in

This should fix the instability tests we are currently seeing.

### DIFF
--- a/.github/workflows/install-slices.yaml
+++ b/.github/workflows/install-slices.yaml
@@ -81,8 +81,9 @@ jobs:
 
             echo "install_all=true" >> $GITHUB_OUTPUT
           else
-            # Filter the releases to only the affected branch.
-            export RELEASES=$(echo "$RELEASES" | jq '[.[] | select(.ref == "${{ github.base_ref }}")]')
+            # Filter the releases to only the affected branch, then swap the ref
+            # such that the correct branch is tested.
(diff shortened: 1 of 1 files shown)


clay-lake:
Here is an example workflow run on my fork. It wont run the ARM tests since I have no GH runner for that, but it demonstrates that the installability tests are working correctly.
https://github.com/clay-lake/chisel-releases/actions/runs/14617738468/job/41009782333?pr=3

When we make changes to CI all of our slices are installed together, but this seems to cause problems on its own resulting in timeouts seen [here](https://github.com/canonical/chisel-releases/actions/runs/14407004035/job/40405854365). Maybe we could refactor the installability test to only spoof installation to verify SDFs can be sliced together, then only run real installability test on individual SDFs when the package they are slicing is updated in the archives.

clay-lake:
@norrisjeremy That is a good point; there is no restriction as far as I know of other than github-hosted ARM runners are only previews for now, but I modified our spread workflows to select github-hosted runners outside of the Canonical org,

https://github.com/canonical/chisel-releases/pull/534

cjdcordeiro (approved):
LGTM thanks. the context deadline CI fail is unrelated, so let's merge

PR #291, opened 2024-07-24
feat(24.04): add perl

zhijie-yang:
# Proposed changes
Add Perl for 24.04

### Forward porting
Not applicable

## Testing

```bash
mkdir rootfs
chisel cut --release ./ --root rootfs/ perl_bins bash_bins base-files_base
sudo chroot rootfs perl
print "Hello world"
```

Output:

```
perl: warning: Setting locale failed.
perl: warning: Please check that your locale settings:
        LANGUAGE = (unset),
        LC_ALL = (unset),
        LC_ADDRESS = "de_DE.UTF-8",
        LC_IDENTIFICATION = "de_DE.UTF-8",
        LC_MEASUREMENT = "de_DE.UTF-8",
        LC_MONETARY = "de_DE.UTF-8",
        LC_NAME = "de_DE.UTF-8",
        LC_NUMERIC = "de_DE.UTF-8",
        LC_PAPER = "de_DE.UTF-8",
        LC_TELEPHONE = "de_DE.UTF-8",
        LC_TIME = "de_DE.UTF-8",
        LANG = "en_US.UTF-8"
    are supported and installed on your system.
perl: warning: Falling back to the standard locale ("C").
Hello world%
```

## Checklist

* [x] I have read the [contributing guidelines](
https://github.com/canonical/chisel-releases/blob/main/CONTRIBUTING.md)
* [x] I have tested my changes ([see how](https://github.com/canonical/chisel-releases/blob/main/CONTRIBUTING.md#7-test-your-slices-before-opening-a-pr))
* [x] I have already submitted the [CLA form](
https://ubuntu.com/legal/contributors/agreement)

PR #532, opened 2025-04-21
feat(24.04): add SDFs for apache2, apache2-bin, apache2-data, apache-utils and dependencies

eunufe:
# Proposed changes

feat(24.04): add SDFs for apache2, apache2-bin, apache2-data, apache-utils and dependencies
## Checklist

### DIFF
--- a/slices/apache2-bin.yaml
+++ b/slices/apache2-bin.yaml
@@ -1,0 +1,1038 @@
+package: apache2-bin
+
+essential:
+  - apache2-bin_copyright
+
+slices:
+  apache2:
+    essential:
+      - libapr1t64_libs
+      - libaprutil1t64_libs
+      - libpcre2-8-0_libs
+    contents:
+      /usr/sbin/apache2:
+
+  bins:
+    essential:
+      - apache2-bin_apache2
+      - apache2-bin_modules
+      - apache2-bin_state
+
+  modules:
+    essential:
+      - apache2-bin_mod-access-compat
+      - apache2-bin_mod-actions
+      - apache2-bin_mod-alias
+      - apache2-bin_mod-allowmethods
+      - apache2-bin_mod-asis
+      - apache2-bin_mod-auth-basic
+      - apache2-bin_mod-auth-digest
+      - apache2-bin_mod-auth-form
+      - apache2-bin_mod-authn-anon
+      - apache2-bin_mod-authn-core
+      - apache2-bin_mod-authn-dbd
+      - apache2-bin_mod-authn-dbm
+      - apache2-bin_mod-authn-file
+      - apache2-bin_mod-authn-socache
+      - apache2-bin_mod-authnz-fcgi
+      - apache2-bin_mod-authnz-ldap
+      - apache2-bin_mod-authz-core
+      - apache2-bin_mod-authz-dbd
+      - apache2-bin_mod-authz-dbm
+      - apache2-bin_mod-authz-groupfile
+      - apache2-bin_mod-authz-host
+      - apache2-bin_mod-authz-owner
+      - apache2-bin_mod-authz-user
+      - apache2-bin_mod-autoindex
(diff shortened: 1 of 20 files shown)


norrisjeremy:
FYI, a few notes:

1. Something appears to be broken with how the `CI / Instability tests / Install` parts work in the Github Actions workflow, that is incorrectly causing failures: I think this is what #526 is trying to remedy?
2. Something unusual seems to happen with how the `CI / Integration tests / Run Spread tests` work that is causing it to neglect to run the tests we built for the `apache2` SDF: it only runs them for the the `apache2-bin` & `apache2-utils` SDFs, skipping the `apache2` SDF.
3. We need the Perl SDFs to be forward ported to 24.10 in order to in turn forward port these new Apache SDFs: it appears this was missed when #291 was merged.
4. The event MPM appears to cause segfaults when run as part of the spread test suite in Github Actions, because Apache claims `libgcc_s.so.1 must be installed for pthread_exit to work`, however this library is definitely present in the chroot. When we run Apache locally in our own containers, this does not seem to occur, so we are unable to determine what is causing the event MPM to fail in as part of the spread test suite in Github Actions.

slices/nginx-common.yaml
```diff
@@ -9,6 +9,8 @@ slices:
   # handled differently in Chisel.
   # There's also no need (yet) for any init- and systemd-related files.
   config:
+    essential:
+      - apache2_documentroot
```
norrisjeremy, line 13: This is ugly, but it was the best solution we could find due to the fact that both the Apache & NGINX packages claim ownership of the `/var/www/html` directory, and chisel does not allow multiple SDFs to claim ownership of the same directory.
We are open to suggestions for a better solution.

slices/apache2-bin.yaml
cjdcordeiro, line 1: This is an amazing piece of work @eunufe ! Thanks.

Before proceeding with the review, I'd like to understand better your use case and raise the need to have tests for each module slice.

For example, we must ensure that if a user runs something like: `chisel cut ... apache2-bins_mod-auth-basic`, that the resulting filesystem will actually be capable to performing basic authentication.

We had a similar challenge with Perl: https://github.com/canonical/chisel-releases/blob/ubuntu-24.04/slices/perl-modules-5.38.yaml. In that case, some level of testing automation was achievable in order to ensure minimum functionality (e.g. that the Perl modules could be loaded). So we'd need to think about doing the same here! Otherwise we risk shipping broken slice definitions.

I'd suggest adopting one of the following strategies (from most to least time consuming):
 1. keep these slices, but test them all. This can be time consuming, but the resulting SDF will offer the most granular split we can have, which is amazing for building tiny apache2 applications
 2. attempt to re-slice these modules, but grouping modules into higher abstraction slices (e.g. all the authn-related modules into a single authn slice). I assume you are following https://httpd.apache.org/docs/2.4/mod/, so this option would create some artificial groupings, which isn't great, but still somewhat acceptable as a trade off, as long as the slices make sense
 3. just slice what you need. If you have a use case for these slices, just define the slices for the modules that you'll be using. This way you leave further splitting for future PRs (e.g. this is what we did in https://github.com/canonical/chisel-releases/blob/ubuntu-24.04/slices/coreutils.yaml#L132 - just added slices for contents that were explicitly requested, instead of trying to achieve full granularity from the get-go)

cjdcordeiro (changes requested):

slices/apache2-bin.yaml
norrisjeremy: Hi @cjdcordeiro,

Just wanted to follow-up and see what your thoughts are on the above proposal?

Thanks,
Jeremy

cjdcordeiro: Hi @norrisjeremy ,

I did reply but I'm just now realizing that my comment is in a "Pending" state :/ 

image

I'm sorry for not noticing the glitch earlier.

PR #540, opened 2025-04-25
ci(lint): do not produce warnings

rebornplusplus:
The lint workflow currently produces a ton of warnings for slices names not being sorted. In practice, we do not abide by that rule much and it's tough to find the actual errors within the sea of warnings. This PR disables emitting those warnings.


clay-lake (approved):
This is a good UX improvement. Thanks for bringing it up.

PR #556, opened 2025-05-05
feat(22.04): podman - WIP

clay-lake:
**This is WIP and is for demonstration only**


slices/conmon.yaml
```diff
@@ -0,0 +1,17 @@
+package: conmon
```
norrisjeremy, line 1: I think this overlaps with #547?

norrisjeremy: Nvm, I didn't see the earlier conversation there.

lczyk:
@clay-lake, looks old. can this be closed?

PR #564, opened 2025-05-21
ci: replace install-slices with new concurrent tool

rebornplusplus:
The new tool can install slices concurrently and adds a few new options.

---

TODO: more description :)


alesancor1 (commented):
Hey Rafid, thanks for this! This is very good!

Leaving a couple of comments below, mostly to be addressed after moving the new tool to the toolbox

PR #571, opened 2025-06-06
feat(25.04): rabbitmq-server

clay-lake:
# Included in this PR:
- Portforward of RabbitMQ Server from 24.10


clay-lake:
After investigation with @alesancor1, it appears that i386 is not supported by rabbitmq-server. Which also begs the question, should we drop support for i386 all together since support was dropped since 19.10(?)

PR #581, opened 2025-07-04
feat(24.04): add busybox-static slice

endersonmaia:
Proposed changes

Add busybox-static slice using the new chisel:1.2.0 `prefer` functionality to avoid conflicts.

I hope this is the way it was designed to be used, I tested installing it locally and it works.

## Checklist


endersonmaia:
Seems the CI still uses chisel 1.0.0, is that it?

cjdcordeiro (commented):
Thanks for this @endersonmaia !

I'm being a bit cautious with the introduction of `prefer`. I'd like to allow for a grace period to give enough time for existing users to update their Chisel versions.

So my question is: do you need this for 24.04? Cause if not, I'd rather start with 25.04 (the most recent, and interim, release) as a way to slowly introduce the disruption for those using <1.2.0

cjdcordeiro:
Noted @endersonmaia . If you don't mind then, I'd rather start by merging the slices for 25.04 first, wait a few days to let the community react, and then merge this one. Wdyt?

cjdcordeiro (commented):
@endersonmaia I'd like to get back to this now that we have opened the 26.04 release, where we plan to drop support for Chisel v1.1.0.

I understand your original goal was to use this slice with 24.04, but given that we've already opened the new LTS (still in development for the next ~5 months), I wanted to ask you if **relying on it would be an option for you**?

I'm still reluctant to introduce this in previous releases because even after 11 months of it being released, Chisel v1.1.0 is still the most downloaded version of Chisel, with ~25% of all downloads and 50% more downloads than v1.2.0. Just for snaps alone, I just confirmed there are still devices with an active version of v1.1.0. 

Having said that, I'm confident there would be a reasonable number of people affected by the backporting of `prefer` into older releases.

cjdcordeiro:
brilliant @endersonmaia .

In that case, see https://github.com/canonical/chisel-releases/pull/800

PR #582, opened 2025-07-07
fix(22.04): add wildcard to libssl3 path for FIPS compatibility

clay-lake:
# Included in this PR:
- Added wildcard to account for FIPS enabled package in `esm-updates` archive.

Note: This fix only applies to Jammy. Noble and Focal FIPS enabled `openssl` have different dependencies.


cjdcordeiro (approved):
Nice. thanks

PR #589, opened 2025-07-14
feat(20.04): add python3.9

cjdcordeiro:
# Proposed changes

Slice Python3.9, based on Python3.8. The diff is really small, and noted in the form of comments throughout the SDFs (e.g. " ... new in 3.9" or  "gone in 3.9")

### Forward porting

None available.

## Checklist

### DIFF
--- a/slices/libpython3.9-stdlib.yaml
+++ b/slices/libpython3.9-stdlib.yaml
@@ -1,0 +1,399 @@
+package: libpython3.9-stdlib
+
+# The slices in this package have been grouped with inspiration from the Python
+# 3.9 Standard Library (https://docs.python.org/3.9/library/index.html).
+# Aside from the "core" slice which contains the minimal libraries that should
+# come from this package and the "extra" slice which contains miscellaneous
+# libraries, all the other slice definitions are sorted by their names. The
+# "core" slice is placed at the very beginning and the "extra" slice at the
+# very end.
+slices:
+  # The "core" slice provides a very minimal libpython3.9-stdlib
+  core:
+    essential:
+      - libbz2-1.0_libs
+      - libc6_libs
+      # Adding libgcc-s1 here because it is required for the arm arch.
+      # In practice, libc6 should bring it anyway, BUT, there is a circular
+      # dependency in the archives, where libc6 depends on libgcc-s1 and
+      # vice versa. We don't allow that circular dependency to exist in Chisel
+      # and libgcc-s1 already depends on libc6, so we'll have to include this
+      # line here, explicitly.
+      - libgcc-s1_libs
+      - liblzma5_libs
+      - libpython3.9-minimal_libs
+      - mime-support_mime-types
+    contents:
+      # _aix_support and _bootsubprocess are new in 3.9
+      /usr/lib/python3.9/_aix_support.py:
+      /usr/lib/python3.9/_bootsubprocess.py:
+      /usr/lib/python3.9/_compression.py:
+      /usr/lib/python3.9/bz2.py:
+      /usr/lib/python3.9/contextvars.py:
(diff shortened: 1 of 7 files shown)


slices/libpython3.9-stdlib.yaml
```diff
@@ -0,0 +1,399 @@
...
+# come from this package and the "extra" slice which contains miscellaneous
+# libraries, all the other slice definitions are sorted by their names. The
+# "core" slice is placed at the very beginning and the "extra" slice at the
+# very end.
+slices:
+  # The "core" slice provides a very minimal libpython3.9-stdlib
+  core:
+    essential:
+      - libbz2-1.0_libs
+      - libc6_libs
```
gaborbk, line 15: `libc6_libs` is already listed under the `libs` of `libpython3.9-minimal`, does it need to be present here as well?

slices/python3.9-minimal.yaml
```diff
@@ -0,0 +1,20 @@
+package: python3.9-minimal
+
+essential:
+  - python3.9-minimal_copyright
+
+slices:
+  bins:
+    essential:
+      - libc6_libs
```
gaborbk, line 9: This lib is already included under `libpython3.9-minimal_libs`, does it need to be listed again here?

gaborbk (commented):
Perhaps just due to my lack of Chisel expertise, I have a few questions

slices/libpython3.9-stdlib.yaml
cjdcordeiro: it's a harmless redundancy.

slices/libpython3.9-stdlib.yaml
```diff
@@ -0,0 +1,399 @@
...
+      # Adding libgcc-s1 here because it is required for the arm arch.
+      # In practice, libc6 should bring it anyway, BUT, there is a circular
+      # dependency in the archives, where libc6 depends on libgcc-s1 and
+      # vice versa. We don't allow that circular dependency to exist in Chisel
+      # and libgcc-s1 already depends on libc6, so we'll have to include this
+      # line here, explicitly.
+      - libgcc-s1_libs
```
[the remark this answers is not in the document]
cjdcordeiro: Good catch. I actually cannot find this requirement for this package. I'll remove it for now, and reconsider introduce it again if the tests fail or we learn otherwise later on

slices/python3.9-minimal.yaml
gaborbk: I see, thanks!

slices/libpython3.9-stdlib.yaml
```diff
@@ -0,0 +1,392 @@
...
+  core:
+    essential:
+      - libbz2-1.0_libs
+      - libc6_libs
+      - liblzma5_libs
+      - libpython3.9-minimal_libs
+      - mime-support_mime-types
+    contents:
+      # _aix_support and _bootsubprocess are new in 3.9
+      /usr/lib/python3.9/_aix_support.py:
```
lengau, line 21: Odd that this is even included in the deb file since it should never even be imported on a Linux system, but I suppose it's harmless there.

Since [the only time this is imported](https://github.com/python/cpython/blob/865270262ade38b96fce88e32a321caac5332a87/Lib/sysconfig.py#L680-L682) is on an AIX system though, do we want to lump it in with `_osx_support.py` in an `other-os` slice?

lengau: Actually it looks like this is done as its own slice in Python 3.10 (https://github.com/canonical/chisel-releases/blob/edde443041a628d3b77a9e475a61afd3d5e36e84/slices/libpython3.10-stdlib.yaml#L53-L58), so mimicking that is probably the way to go.

PR #601, opened 2025-07-21
feat(24.04): add postgresql-client-16 slice + spread tests

taurus-forever:
# Proposed changes

At the moment PostgreSQL slices are missing for both PG client and server apps
(at the same time, the slice for libpq5 has been added in PR#233 already). 
This is the first educational PR to add slice `postgresql-client-16` based on
Ubuntu/Debian APT package `postgresql-client-16`.

The PR container the slice for most common used PostgreSQL CLI tools:
 * psql (PostgreSQL CLI client)
 * createdb / dropdb (CLI tool to create / drop PostgreSQL DBs)
 * pg_dmp (CLI tool to dump content of PostgreSQL DB)

## Related issues/PRs

Reference ticket: [DPE-7291](https://warthogs.atlassian.net/browse/DPE-7291)

### Forward porting

The most of newer Ubuntu versions have newer PostgreSQL version already.
The PostgreSQL 16 was officially released for Ubuntu 24.04 and 24.10 [only](https://packages.ubuntu.com/search?keywords=postgresql-client-16).
Therefore back-porting could be considered to Ubuntu 24.10 only (which is EOL).
The Ubuntu 25.04+ will receive new PR with PostgreSQL 17, once we agree here. 

## Checklist

## Additional Context

[DPE-7291]: https://warthogs.atlassian.net/browse/DPE-7291?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


taurus-forever:
Hi, it is a first PR to practice slicing before touching PostgreSQL server.

Q1: should this be backported to EOL 24.10 ?

Q2: should this slice `postgresql-client-16` be split into smaller slicer shipping each CLI tool independently?
E.g. I might need CLI client `psql` only, but pg_dump ships unnecessary for me dependencies: liblz4-1_libs/zlib1g_libs, etc.

Also, the Ubuntu package has significant list of CLI tools https://packages.ubuntu.com/noble/amd64/postgresql-client-16/filelist (which at some point should be added here). Are there some good example for follow for such "collection of CLI tools" debian package? Tnx!

cjdcordeiro:
Thanks for this @taurus-forever (and for the nicely written description :) )

> Q1: should this be backported to EOL 24.10 ?

No. It's EOL so we should go straight to 25.04 and then 25.10 (once it's opened)

---

> Q2: should this slice postgresql-client-16 be split into smaller slices shipping each CLI tool independently?

This is a pertinent question. The best practice is to "slice according to your use case". There's a trade-off between (A) slice an entire package, and (B) slice just what your use case needs.

With (A), you're creating a Slice Definitions File (SDF) that is more generic and potentially more useful to a wider range of users. At the same time, it requires more in-depth knowledge of the entire package, including for uses that may not be yours, but you'll need to create the spread tests for them anyway. Another consideration to have is, the more generic your SDF is, the more diligent you'll need to be with the design of your slices. For example, see https://github.com/canonical/chisel-releases/blob/068fa6b7e903f7e91d7df9bb2d505c8c4d39d8cd/slices/libpython3.13-stdlib.yaml, and how methodical the design of that SDF was, respecting the actual Python documentation and groupings of modules. 

With (B), you're reducing your scope, making it easier to get your slices reviewed and merged upstream. Sure, depending on your use case, they may not cover all the use cases out there, but at the same time, it's always easier for add new slices/contents, than it is to try to fix a mistake or flawed design introduced by (A).

I'd propose you start with (B) as an approach, assuring there is a good enough SDF to cover the most fundamental use cases (including yours), and then improve (and further slice) on top of that. Concretely, if you don't need `pg_dump`, create the specific slices to encompass just the other CLI tools.

> Also, the Ubuntu package has significant list of CLI tools https://packages.ubuntu.com/noble/amd64/postgresql-client-16/filelist (which at some point should be added here). Are there some good example to follow for such "collection of CLI tools" debian package? Tnx!

I like what we have done with `coreutils`: https://github.com/canonical/chisel-releases/blob/ubuntu-24.10/slices/coreutils.yaml

lczyk:
to add a little bit regarding slicing specific deps out of a package, what you then need to do is to establish which of the deps of the whole package are appropriate for the sliced out tool. what i found useful in this case is something like this:

```bash
apt install postgresql-client-16
ls /usr/lib/postgresql/16/bin/ | xargs -I{} -n1 bash -c 'echo {}; ldd {};'
```

which shows you what libraries do each of the cli tools links against. Something like this:

```
pg_config
  linux-vdso.so.1 (0x0000ffff871ac000)
  libc.so.6 => /lib/aarch64-linux-gnu/libc.so.6 (0x0000ffff86f80000)
  /lib/ld-linux-aarch64.so.1 (0x0000ffff87170000)
```

means that it links only with `libc6`. `linux-vdso.so` is the [virtual dynamic shared object](https://man7.org/linux/man-pages/man7/vdso.7.html) which will be in every binary, and `ld-linux-aarch64.so` is the dynamic linker which will be in every dynamically linked file.

in this case it was pretty easy to figure out where each file is from, but in some cases it's not obvious. amongst the libraries which `vacuumdb` links with is `libcrypto.so.3`. where is it from? `dpkg --search` to the rescue:

```
$ dpkg --search $(find / -name libcrypto.so.3)
libssl3t64:arm64: /usr/lib/aarch64-linux-gnu/libcrypto.so.3
```

aka from `libssl3t64` :+1:

lczyk:
btw, can you change the title to `feat(24.04): add postgresql-client-16 slice + spread tests` for consistency with other PRs? (as in, adding the `(24.04)`)

slices/postgresql-client-16.yaml
```diff
@@ -0,0 +1,31 @@
+package: postgresql-client-16
+
+essential:
+  - postgresql-client-16_copyright
+
+slices:
+  bins:
+    contents:
+      /usr/bin/createdb: {symlink: /usr/lib/postgresql/16/bin/createdb}
+      /usr/bin/dropdb: {symlink: /usr/lib/postgresql/16/bin/dropdb}
```
lengau, line 10: These two commands seem like they might be needed less than `psql` or `pg_dump`, so it might be worth considering if we want to put those into separate slices.

lengau (approved):

PR #604, opened 2025-07-31
feat(24.04): add apt slices

zhijie-yang:
# Proposed changes

This PR adds the slices definition to provide the functionalities `update` and `install` for `apt`.

## Related issues/PRs

N/A

### Forward porting

#617 

## Checklist


tests/spread/integration/apt/task.yaml
```diff
@@ -0,0 +1,25 @@
...
+  umount "${rootfs}/dev"
+
+  rootfs="$(install-slices apt_placeholder)"
+
+  mkdir -p "${rootfs}/dev"
+  mount --bind /dev "${rootfs}/dev"
+
+  chroot "${rootfs}/" apt | grep -q "chisel cut --root / apt_"
+
+  umount "${rootfs}/dev"
```
clay-lake, line 25:
```suggestion

  chroot "${rootfs}/" apt | grep -q "chisel cut --root / apt_"

```
Not sure we need `/dev` for this task.

slices/apt.yaml
```diff
@@ -0,0 +1,102 @@
...
+    contents:
+      /etc/apt/apt.conf.d/01-vendor-ubuntu:
+      /etc/apt/apt.conf.d/01autoremove:
+      /etc/apt/auth.conf.d/:
+      /etc/apt/keyrings/:
+      /etc/apt/preferences.d/:
+      /etc/apt/sources.list.d/:
+      /etc/apt/trusted.gpg.d/:
+
+  get-bins:
```
lengau, line 17: Why are these not named `bins` and `libs`?

cjdcordeiro: Possibly redundant, but explicit at least. 
And I don't think we need to distinguish, because when would you want one without the other?

zhijie-yang: Fixed in [0b2aff5](https://github.com/canonical/chisel-releases/pull/604/commits/0b2aff5a41b7730bd3468a1d30a3eb96da804777)

cjdcordeiro:
@zhijie-yang thanks. Let me know once the FP is updated so it can be merged too

slices/apt.yaml
```diff
@@ -0,0 +1,104 @@
...
+      /etc/apt/sources.list.d/:
+      /etc/apt/trusted.gpg.d/:
+
+  apt-get:
+    essential:
+      - apt_config
+      - apt_default-source-list
+      - apt_var
+      - base-files_bin
+      - coreutils_bins
```
lczyk, line 23: see comments about `coreutils_bins` and `sed_bins` in my review of #617

slices/gpgv.yaml
```diff
@@ -0,0 +1,23 @@
+package: gpgv
+
+essential:
+  - gpgv_copyright
+
+slices:
+  bins:
+    essential:
+      - base-files_tmp
+      - coreutils_mktemp
```
lczyk, line 10: why? there is no mention of `mktemp` in the whole dep. even `strings /usr/bin/gpgv | grep mktemp` yields nothing.

slices/dpkg.yaml
```diff
@@ -6,7 +6,9 @@ essential:
 slices:
   bins:
     essential:
+      - diffutils_bins  # used by dpkg -i
```
lczyk, line 9: from what i can tell it's just `cmp` (not all of `diffutils`) and only used by `dpkg-db-backup`. i'd subslice it.

lczyk (commented):
i'd leave a 'request changes' review but too late i guess

slices/apt.yaml
lczyk: just noticed that this PR slices `apt-key`, which is a shell script, in addition to other `apt-...` bits which are bins. lemme investigate some more

lczyk: ok. looks like it dopes use `sed`, so nvm that

lczyk: and `apt-key` contains about 22 packages from `coreutils` so, while i would still advocate to subslice it, its fine as is.

PR #607, opened 2025-08-02
Rocks 2356/coreutils for php 25.04 parallel test

lczyk:
Experimental PR to test #580. pls don't merge


lczyk:
CI failure unrealted. we only care about the time Installability tests take in this PR

PR #613, opened 2025-08-05
ci: forward porting workflow

lczyk:
# Proposed changes

A workflow to automatically label PRs with `forward port missing`

example of a successful run: https://github.com/lczyk/chisel-releases/actions/runs/23120943107/job/67154757937

## Related issues/PRs

- ~~requires `add-pr-comment` from https://github.com/canonical/chisel-releases/pull/720~~ _(posting comments will be added in later commit)_

### Forward porting
N/A

## Checklist

## Additional Context

- ~~I've included all the development files (pyproject, readme, etc) but it's possible let ought to live somewhere else, for example on rocks-toolbox, and the co should just pull the script file from there. tbd~~
- ~~Examples of successful runs can be found [here](https://github.com/lczyk/chisel-releases-workflow-testing/actions)~~
- ~~There are a bunch more tests which can be written, and a bunch more which i can port over from some [proofs-of-principle](https://github.com/lczyk/availability_matrix) which i've written, but i didn't want to make this an even larger of a PR.~~


lczyk:
a couple more extra requirements / changes came up for this came up, so marked as draft for now. internal ticket numbers ROCKS-2441 and ROCKS-2442.

lczyk:
cla check failing on the server side... either way, [example of a successful run](https://github.com/lczyk/chisel-releases-workflow-testing/actions/runs/17543849653) + changing out of draft

alesancor1 (commented):
Looks pretty good on the surface, but it definitely needs a more in depth pass - it is quite a long PR to review!

I feel the toolbox is a better place for the scripts, so we only push the workflow here, at least imo.

In any case, it'd be good to try push this so we can merge sooner than later, it is a great tool to help the review process :+1: 

Ping @canonical/rocks

.github/scripts/forward-port-missing/src/chisel_releases_data.py
```diff
@@ -0,0 +1,999 @@
...
+
+################################################################################
+
+_T = TypeVar("_T")
+
+
+class _TypeConversionDict(dict[str, object]):
+    def get_path(self, path: str, t: type[_T]) -> _T:
+        """Get a value from a nested dictionary using a dot-separated path."""
+        keys = path.split(".")
```
clay-lake: I'd also look at str.partition, it makes it somewhat simpler to handle paths via recursion.

.github/scripts/forward-port-missing/src/chisel_releases_data.py
```diff
@@ -0,0 +1,999 @@
...
+        )
+    if repo not in ("", "security", "updates", "backports"):
+        raise ValueError(f"Invalid repo: {repo}. Must be one of '', 'security', 'updates', or 'backports'.")
+
+    logging.debug("Fetching packages for %s, component=%s, repo=%s", release, component, repo or "<none>")
+
+    name = release.codename
+    name = f"{name}-{repo}" if repo else name
+
+    package_url = f"https://archive.ubuntu.com/ubuntu/dists/{name}/{component}/binary-amd64/Packages.gz"
```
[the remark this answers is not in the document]
clay-lake: This might fail when/if we pull in packages that only exist in pro repositories like the openssl-fips-module3 package

clay-lake (commented):
Just had a first pass review. Nice looking code, but I left a few comments. I'm going to have another pass on the arch later on.

.github/workflows/forward-port-missing.yaml
```diff
@@ -0,0 +1,108 @@
...
+        run: |
+          ln -s "${{ env.files_main }}/${{ env.src_dir }}/chisel_releases_data.py" .
+          ./chisel_releases_data.py --log-level=debug -j=-1 "${{ env.DATA_FILE }}"
+
+      - name: "Check for forward port"
+        if: env.RUN_FORWARD_PORT_MISSING == 'true'
+        run: |
+          ln -s "${{ env.files_main }}/${{ env.src_dir }}/forward_port_missing.py" .
+          ./forward_port_missing.py "${{ env.DATA_FILE }}" --log-level=debug --format=json 1> results.json
+          cat results.json
```
clay-lake:
```suggestion
          ./forward_port_missing.py "${{ env.DATA_FILE }}" --log-level=debug --format=json | tee results.json
```
nit:  we could use `tee` here.

.github/workflows/forward-port-missing.yaml
```diff
@@ -0,0 +1,108 @@
...
+      
+      #== MAIN =================================================================
+
+      - name: "Gather data"
+        env:
+          GITHUB_TOKEN: ${{ secrets.ROCKSBOT_PR_TOKEN }}
+          CHISEL_RELEASES_URL: http://github.com/canonical/chisel-releases
+        run: |
+          ln -s "${{ env.files_main }}/${{ env.src_dir }}/chisel_releases_data.py" .
+          ./chisel_releases_data.py --log-level=debug -j=-1 "${{ env.DATA_FILE }}"
```
clay-lake: nit: we probably do not need to use a variable for the data file since we only reference it twice.

.github/workflows/forward-port-missing.yaml
```diff
@@ -0,0 +1,99 @@
+name: "Forward port missing"
+
+# This workflow checks whether there exists a forward port for the slices and
+# applies the `forward port missing` label if necessary.
+
+on:
+  schedule:
+    # Run at 1 min past every hour
+    # Ref: https://man7.org/linux/man-pages/man5/crontab.5.html
+    - cron: "1 * * * *"
```
cjdcordeiro: this is still every hour, so might as well have "0 * ..."

I think it's fine. If we hit rate limits, we can then think about a more elaborate approach with either 2 crons, or something like `0,30 0-23/3,1-23/3 * * *`

```suggestion
    - cron: "0 * * * *"
```

.github/scripts/forward-port-missing/chisel_releases_data.py
```diff
@@ -0,0 +1,368 @@
...
+                logging.debug("Using a thread pool of size %d.", getattr(executor, "_max_workers", -1))
+                results_head = list(executor.map(get_slices_head, _prs))
+                results_base = list(executor.map(get_slices_base, _prs))
+
+            slices_in_head_by_pr = {pr: slices for pr, slices in zip(_prs, results_head)}
+            slices_in_base_by_pr = {pr: slices for pr, slices in zip(_prs, results_base)}
+
+    total_slices = sum(len(slices) for slices in slices_in_head_by_pr.values())
+    total_slices += sum(len(slices) for slices in slices_in_base_by_pr.values())
+    logging.info("Fetched %d slices for %d PRs in %.2f seconds.", total_slices, len(prs), elapsed())
```
cjdcordeiro: related to https://github.com/canonical/chisel-releases/pull/613/changes#r2792791262

why is this valuable?

.github/scripts/forward-port-missing/chisel_releases_data.py
```diff
@@ -0,0 +1,368 @@
...
+    package_url = f"https://archive.ubuntu.com/ubuntu/dists/{name}/{component}/binary-amd64/Packages.gz"
+    headers = {
+        "User-Agent": "forward-port-missing-script",
+        "Referer": f"https://archive.ubuntu.com/ubuntu/dists/{name}/{component}/binary-amd64",
+    }
+    response = requests.get(package_url, headers=headers)
+
+    if response.status_code != 200:
+        # retry with old-releases if not found in archive
+        package_url = f"https://old-releases.ubuntu.com/ubuntu/dists/{name}/{component}/binary-amd64/Packages.gz"
```
cjdcordeiro: why do you need this if you're only running this script for SUPPORTED_RELEASES?

.github/scripts/forward-port-missing/chisel_releases_data.py
```diff
@@ -0,0 +1,368 @@
...
+            slices_in_base_by_pr = {pr: slices for pr, slices in zip(_prs, results_base)}
+
+    total_slices = sum(len(slices) for slices in slices_in_head_by_pr.values())
+    total_slices += sum(len(slices) for slices in slices_in_base_by_pr.values())
+    logging.info("Fetched %d slices for %d PRs in %.2f seconds.", total_slices, len(prs), elapsed())
+
+    return slices_in_head_by_pr, slices_in_base_by_pr
+
+
+def _get_packages_by_release(
```
cjdcordeiro: note that this won't be able to find stuff like fips packages. just a note

.github/scripts/forward-port-missing/chisel_releases_data.py
```diff
@@ -0,0 +1,368 @@
...
+            f"Failed to download package list from '{package_url}'. HTTP status code: {response.status_code}"
+        )
+
+    with gzip.GzipFile(fileobj=io.BytesIO(response.content)) as f:
+        content = f.read().decode("utf-8")
+
+    return set(m.group(1) for m in _PACKAGE_RE.finditer(content))
+
+
+def _group_prs_by_ubuntu_release(prs: set[PR], ubuntu_releases: list[UbuntuRelease]) -> dict[UbuntuRelease, set[PR]]:
```
cjdcordeiro: [1]

.github/scripts/forward-port-missing/forward_port_missing.py
```diff
@@ -0,0 +1,384 @@
...
+            if pr not in grouped_comparisons:
+                grouped_comparisons[pr] = {}
+
+    # For each of the dicts in grouped_comparisons, we want all of the future releases, even if there are no comparisons
+    for pr, comparisons_by_future_release in grouped_comparisons.items():
+        ubuntu_release = pr.ubuntu_release
+        future_releases = [r for r in prs_by_ubuntu_release if r > ubuntu_release]
+        for future_release in future_releases:
+            if future_release not in comparisons_by_future_release:
+                comparisons_by_future_release[future_release] = set()
```
cjdcordeiro: couldn't this be done above, alongside `        grouped_comparisons[pr][future_release].add(comparison)`, to avoid another loop?

cjdcordeiro (commented):
Also
 - see copilot comments. some are quite valid
 - provide better documentation and rationale explanations. that includes the naming of steps, which atm can be quite high level (eg. "Gather data"..._what data?_)
 
 Btw, as a side note and not a change request (given how advanced this is), my spidey sense is kinda telling me that Python was the wrong tool for this job, given the amount of `git` and distro manipulation we have in this PR. Relying on tools like `git`, `gh`, `distro-info`, `{j,y}q` and `rmadison`, could've potentially reduced the size of the PR significantly...but this is just a feeling - I have no evidence of that 🕷️ 

---

[1]
I see a design red flag:
- the splitting of python scripts leads to a lot of duplication (e.g. same functions being called, for doing the same job)

This splitting also increases the complexity of your code, leading you to have to do things like schema validation (e.g. in `load_data_from_json`).

My suggestion is: regardless of whether these scripts are meant to be used separately (i.e. is the `chisel-releases-navigator` using `chisel_releases_data.py` on its own?), I'd merge them into a single python program. You can then adjust the CLI to instruct the code on what the output should be (e.g. in case you really need the ability to get a `data.json` data file).

.github/scripts/forward-port-missing/chisel_releases_data.py
```diff
@@ -0,0 +1,368 @@
...
+            merge_bases_by_pr = {pr: _get_merge_base(pr.base, pr.head) for pr in prs}
+        else:
+            _prs = list(prs)  # we want list for zipping with results
+            with ThreadPoolExecutor(max_workers=jobs) as executor:
+                logging.debug("Using a thread pool of size %d.", getattr(executor, "_max_workers", -1))
+                results = list(executor.map(lambda pr: _get_merge_base(pr.base, pr.head), _prs))
+            merge_bases_by_pr = {pr: mb for pr, mb in zip(_prs, results)}
+
+    logging.info("Fetched merge bases for %d PRs in %.2f seconds.", len(prs), elapsed())
+    for pr, mb in merge_bases_by_pr.items():
```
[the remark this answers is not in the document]
lczyk: i've decided not to remove this in the rework since it's a small function and it helped a lot with the dev, but it's really easy to rip this out if you want, especially now its a single file

.github/scripts/forward-port-missing/forward_port_missing.py
```diff
@@ -0,0 +1,384 @@
...
+            continue
+        _prs_by_ubuntu_release[pr.ubuntu_release].add(pr)
+    prs_by_ubuntu_release: dict[UbuntuRelease, frozenset[PR]] = {
+        k: frozenset(v) for k, v in _prs_by_ubuntu_release.items()
+    }
+
+    # Make sure we have all the ubuntu_releases as keys, even if they have no PRs
+    for ubuntu_release in ubuntu_releases:
+        if ubuntu_release not in prs_by_ubuntu_release:
+            prs_by_ubuntu_release[ubuntu_release] = frozenset()
```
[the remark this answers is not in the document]
lczyk: resolving as outdated

.github/scripts/forward-port-missing/chisel_releases_data.py
lczyk: yep. good point. not an issue for now but worth keeping in mind

.github/scripts/forward-port-missing/forward_port_missing.py
```diff
@@ -0,0 +1,384 @@
...
+        if future_release not in grouped_comparisons[pr]:
+            grouped_comparisons[pr][future_release] = set()
+        grouped_comparisons[pr][future_release].add(comparison)
+
+    # We may not have all the PRs in the grouped_comparisons, since they may not have had any PRs to compare to.
+    # We need to add them with empty dicts.
+    for prs_in_release in prs_by_ubuntu_release.values():
+        for pr in prs_in_release:
+            if pr not in grouped_comparisons:
+                grouped_comparisons[pr] = {}
```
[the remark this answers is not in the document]
lczyk: resolving as outdated

.github/scripts/forward-port-missing/forward_port_missing.py
```diff
@@ -0,0 +1,417 @@
...
+            warn(
+                f"Rate limit exceeded when fetching diff for PR #{pr_number}. Skipping."
+            )
+            sys.exit(1)
+        return pr_number, Diff(diff_text)
+
+    with timing_context() as elapsed:
+        with requests.Session() as s, ThreadPoolExecutor(max_workers=5) as executor:
+            _diffs = list(executor.map(partial(_fetch_diff, s), results))
+    diffs: dict[int, Diff] = dict(_diffs)
```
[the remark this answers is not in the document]
lczyk: lemme look into that...

.github/scripts/forward-port-missing/requirements.txt
```diff
@@ -0,0 +1,3 @@
+requests>=2.32.5
+pyyaml>=6.0.0
+diff-parser>=1.1
```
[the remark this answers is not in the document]
cjdcordeiro: copilot has a point here. Having seen the ever increasing amount of supply chain attacks these past few months, i think it doesn't harm to pin these to a specific version

.github/scripts/forward-port-missing/forward_port_missing.py
```diff
@@ -0,0 +1,371 @@
...
+import datetime
+import gzip
+import io
+import logging
+import os
+import re
+from concurrent.futures import ThreadPoolExecutor
+from itertools import product
+import subprocess as sub
+import sys
```
cjdcordeiro: unused?

.github/scripts/forward-port-missing/forward_port_missing.py
```diff
@@ -0,0 +1,371 @@
...
+            )
+            if not end_of_life:
+                continue
+            if end_of_life < datetime.date.today():
+                continue
+
+            # Parse the short codename from the "archives" field in chisel.yaml.
+            # This is needed to quarry the archive
+            archives: list[str] = (
+                chisel_yaml.get("archives", {}).get("ubuntu", {}).get("suites", [])
```
cjdcordeiro, line 280: nit: it's ok for now. But I'd add a NOTE comment, to say that "it doesn't necessarily need to be `.get("ubuntu")`. This is an arbitrary key, and we don't really care about its value. With Pro archives, more entries can exist, and all of them will have a `suites` field, **where the codename should always match**"

.github/scripts/forward-port-missing/forward_port_missing.py
```diff
@@ -0,0 +1,371 @@
...
+            capture_output=True,
+            text=True,
+        ).stdout.splitlines()
+        out = [b.strip().strip("'") for b in out]
+        out = [b.removeprefix("origin/") for b in out]
+        branches = set(b for b in out if b.startswith("ubuntu-"))
+        if not branches:
+            raise Exception("No ubuntu branches in chisel-releases")
+
+        _branches = sorted(map(lambda b: b.removeprefix("ubuntu-"), branches))
```
cjdcordeiro: where is this used? it's being redefined in line 290 below

.github/scripts/forward-port-missing/forward_port_missing.py
```diff
@@ -0,0 +1,371 @@
...
+
+    # filter down to PRs into branches named "ubuntu-XX.XX"
+    results = [pr for pr in results if pr["base"]["ref"].startswith("ubuntu-")]
+
+    # filter out draft PRs
+    results = [pr for pr in results if not pr.get("draft", False)]
+
+    # filter down to PRs into supported branches if specified
+    if supported_branches is not None:
+        results = [pr for pr in results if pr["base"]["ref"] in supported_branches]
```
cjdcordeiro: Wouldn't it be more efficient to simply loop over `results` once?

Something like

```python
for pr in results[:]:
  if pr["base"]["ref"] not in supported_branches or not pr.get("draft", False):
    results.remove(pr)
    continue
```

.github/scripts/forward-port-missing/forward_port_missing.py
```diff
@@ -0,0 +1,371 @@
...
+            content = f.read().decode("utf-8")
+
+        return (
+            short_codename,
+            component,
+            repo,
+            set(m.group(1) for m in _PACKAGE_RE.finditer(content)),
+        )
+
+    _components = ("main", "restricted", "universe", "multiverse")
```
cjdcordeiro, line 207: nit: this may be conservative on purpose, but if you want to be aligned with Chisel, you can get the `components` from chisel.yaml too. 

(maybe for a future PR)

.github/scripts/forward-port-missing/forward_port_missing.py
```diff
@@ -0,0 +1,371 @@
...
+    _repos = ("", "security", "updates", "backports")
```
cjdcordeiro, line 208: similar to the above - but this one has an additional consideration: Pro archives. Again, not for now, especially because I think our other CI workflows aren't really fully ready for Pro tests, but still, note for future self

.github/scripts/forward-port-missing/forward_port_missing.py
```diff
@@ -0,0 +1,371 @@
...
+            )
+
+    info(f"Fetched packages for {len(codenames)} releases in {elapsed():.2f} seconds.")
+
+    # Union all components and repos for each release
+    packages_by_release: dict[str, set[str]] = {
+        release: set() for release in codenames.keys()
+    }
+    _codenames = {short: release for release, short in codenames.items()}
+    for short_codename, component, repo, packages in results:
```
cjdcordeiro:
```suggestion
    for short_codename, _, _, packages in results:
```

cjdcordeiro: If you don't need these, maybe `_fetch_packages` doesn't need to return them

.github/scripts/forward-port-missing/forward_port_missing.py
```diff
@@ -0,0 +1,371 @@
...
+    info(f"Fetching packages for {len(codenames)} releases...")
+
+    def _fetch_packages(args: tuple[str, str, str]) -> tuple[str, str, str, set[str]]:
+        """Fetch the list of packages for a given release, component, and repo"""
+        short_codename, component, repo = args
+        name = f"{short_codename}-{repo}" if repo else short_codename
+
+        url = f"https://archive.ubuntu.com/ubuntu/dists/{name}/{component}/binary-amd64/Packages.gz"
+
+        with requests.Session() as s:
```
cjdcordeiro, line 194: nit: you could reuse a session instead of creating new ones (feel free to ignore if it's too much work though - not really worried about performance here)

.github/scripts/forward-port-missing/apply_pr_labels.sh
```diff
@@ -0,0 +1,36 @@
...
+
+main() {
+    if [ "$#" -ne 1 ]; then echo "Usage: $0 <results.txt>"; exit 1; fi
+    # support both GITHUB_TOKEN and GH_TOKEN for flexibility. `gh` expects GH_TOKEN
+    export GH_TOKEN="${GITHUB_TOKEN:-${GH_TOKEN:-}}"
+    command -v gh >/dev/null 2>&1 || { echo >&2 "gh is required but it's not installed. Aborting."; exit 1; }
+
+    # read the add line from the input file
+    add_line=$(grep "^add:" "$1" || true)
+    remove_line=$(grep "^remove:" "$1" || true)
```
cjdcordeiro: I'm not sure what to think about this contract between forward_port_missing and this script :p 

It feels like the whole functionality is very fragile and depends on the parsing of a redirect of another script's logs onto a text file, assuming a specific writing convention.

Given that this is a CI-specific thing, I'm trying not to overthink this, but it does feel like a code smell. 

(pause for reflection xD)

let's move on for now and improve later if needed (this PR has suffered too much already)

.github/scripts/forward-port-missing/apply_pr_labels.sh
```diff
@@ -0,0 +1,36 @@
...
+    add_numbers=$(echo "$add_line" | cut -d: -f2 | tr -d ' ')
+    remove_numbers=$(echo "$remove_line" | cut -d: -f2 | tr -d ' ')
+
+    for number in ${add_numbers//,/ }; do
+        if [ -n "$number" ]; then
```
cjdcordeiro: this check may be unnecessary since you already `set -u`?

.github/scripts/forward-port-missing/apply_pr_labels.sh
```diff
@@ -0,0 +1,36 @@
...
+            echo "Adding label to PR #$number"
+            gh pr edit "$number" --add-label "forward port missing"
```
cjdcordeiro: it is important to note that this should be in run in the context of the repo, thus assuming there was a checkout done prior to executing this script. Should `GH_REPO` be set/used?

cjdcordeiro (changes requested):
Much much nicer! TYVM

I think it's a win. Just some questions and minor comments. Please get another reviewer and let's get it merged ;)

.github/scripts/forward-port-missing/forward_port_missing.py
lczyk: https://github.com/canonical/chisel-releases/pull/613/commits/aa4674a6350fd4c400c340274dca847ee3190222

.github/scripts/forward-port-missing/forward_port_missing.py
lczyk: noo! :flushed: mutating `results` while looping over them is hard to reason about and potentially buggy / weird behaviour. list comprehensions are safe in that regard. `foo = [x**x for x in foo]` is always fine.

lczyk: you can make this only loop once like this though: https://github.com/canonical/chisel-releases/pull/613/commits/eac5ee0e6f98b5920f82d576d27d2b046d58eb11 and this is very efficient (and very pythonic) :)

.github/scripts/forward-port-missing/forward_port_missing.py
lczyk: hmm... yeah, future PR i think :+1:

.github/scripts/forward-port-missing/forward_port_missing.py
lczyk: https://github.com/canonical/chisel-releases/pull/613/commits/ad5523bc966e58983347fd6cc7e9a4fc34e0ab00

.github/scripts/forward-port-missing/forward_port_missing.py
lczyk: https://github.com/canonical/chisel-releases/pull/613/commits/ad5523bc966e58983347fd6cc7e9a4fc34e0ab00

.github/scripts/forward-port-missing/forward_port_missing.py
lczyk: https://github.com/canonical/chisel-releases/pull/613/commits/f1ea7498ec6b2ab77a7e734ac754f926d82d9397

.github/scripts/forward-port-missing/forward_port_missing.py
lczyk: nope. sessions are not thread save so i must make one per thread. i could create a pool of sessions and pull/reuse them like that but that would be lots of work and code complication for what i'm not convinced would even be any speedup.

i *did* try using just one session and sharing that between all the threads and it does work and it is a little faster, but again, `requests.Sessions` is not officially thread-safe (something about a cookie jar) and i thought better not risk "undefined behaviour". the speed penalty is really small anyway so we're not loosing much

.github/scripts/forward-port-missing/apply_pr_labels.sh
lczyk: happy to change to whatever else. literally just typed the first simple thing that came to mind which was borderline as simple as it gets. i'm not attached to it in the slightest ;p

let's see what the other reviewers say. i'm not gonna change it yet, not to keep changing the code back and forth, but if we agree on what to change it to / come up with anything better, then i'm all for

.github/scripts/forward-port-missing/apply_pr_labels.sh
lczyk: https://github.com/canonical/chisel-releases/pull/613/commits/90cdfc2338d28f8bbe14f5e80066beb07b57ef2d

.github/scripts/forward-port-missing/apply_pr_labels.sh
lczyk: https://github.com/canonical/chisel-releases/pull/613/commits/a6ddc02a56e596820df0d73f6b93d243664e28f9

.github/scripts/forward-port-missing/test_forward_port_missing.py
```diff
@@ -0,0 +1,290 @@
...
+        ]
+
+    @patch("forward_port_missing.requests.Session")
+    def test_basic(self, mock_session: MagicMock) -> None:
+
+        side_effects: list[MagicMock] = self.make_side_effects(
+            self.json_response, self.diff_text
+        )
+
+        get = mock_session.return_value.__enter__.return_value.get
```
zhijie-yang: This part seems to be a boilerplate shared by many test cases. Can we make it reusable?

.github/scripts/forward-port-missing/test_forward_port_missing.py
```diff
@@ -0,0 +1,290 @@
...
+            prs=prs,
+            slices_per_branch=slices_per_branch,
+            packages_by_release=packages_by_release,
+        )
+
+        assert to_add == set()
+        assert to_remove == ({1} if labels else set())
+
+    @with_and_without_labels
+    def test_slices_missing_but_discontinued_only_in_some(
```
zhijie-yang, line 289: Should we also add another test case, where `foo` exists in `ubuntu-20.04` and in `ubuntu-24.04`, but is missing in `ubuntu-22.04`?

.github/scripts/forward-port-missing/apply_pr_labels.sh
zhijie-yang: I'd say it's better to combine these logics into the Python scripts, as we don't need to do the serialization and deserialization to set up the interface between the Python script and the shell script.

By adding a flag, e.g., `--apply` to the Python script, we can also easily control the read-only behavior of the script.

zhijie-yang (changes requested):

.github/scripts/forward-port-missing/apply_pr_labels.sh
lczyk: yep. that's neater, i agree. see https://github.com/canonical/chisel-releases/pull/613/commits/13865316582b78bbfdb62848beab2eeb8e92d891

.github/scripts/forward-port-missing/test_forward_port_missing.py
lczyk: https://github.com/canonical/chisel-releases/pull/613/commits/2cbc117308efce22926891cbc756c192e4252af6

.github/scripts/forward-port-missing/test_forward_port_missing.py
lczyk: https://github.com/canonical/chisel-releases/pull/613/commits/340d3c933edfc79a33f8ff143b1cb495c8cea1e0

.github/workflows/forward-port-missing.yaml
```diff
@@ -38,12 +38,5 @@ jobs:
         env:
           # we authenticate ourselves with the actions token to avoid hitting the unauthenticated rate limit
           GITHUB_TOKEN: ${{ github.token }}
```
cjdcordeiro, line 40: for a next PR, we should revisit the use of this token across all workflows

.github/scripts/forward-port-missing/forward_port_missing.py
cjdcordeiro: PS this is not looping over the `results`, but a shallow copy of it.

cjdcordeiro (approved):
LGTM, thanks. Note that there was a comment in https://github.com/canonical/chisel-releases/pull/613#discussion_r2917157635 that I believe might have been misunderstood. I'm not blocking the PR on that, but please reply to that comment just to make sure we're on the same page

.github/scripts/forward-port-missing/forward_port_missing.py
lczyk: mmm... gochya. ok, less _"noo!"_ ;)

https://github.com/canonical/chisel-releases/commit/eac5ee0e6f98b5920f82d576d27d2b046d58eb11 is also looping once too though, and so if whether this is better or worse, at this point i'd leave it for future consideration / improvement

.github/workflows/forward-port-missing.yaml
lczyk: yep. absolutely. for this specific line even a very unprivileged token with read only of public data would work. it's here, as the comment says, just for the rate limit reasons

.github/scripts/forward-port-missing/forward_port_missing.py
```diff
@@ -350,7 +351,40 @@ def determine_forward_porting_status(
     return to_add_label, to_remove_label
 
 
+def apply_labels(to_add: set[int], to_remove: set[int]) -> None:
+    """Apply or remove the 'forward port missing' label on PRs using the gh CLI."""
+    env = os.environ.copy()
+    gh_token = os.getenv("GITHUB_TOKEN") or os.getenv("GH_TOKEN", "")
+    if gh_token:
+        env["GH_TOKEN"] = gh_token
```
zhijie-yang: This logic is redundant. GH will default to `GITHUB_TOKEN` if `GH_TOKEN` is not set.

https://cli.github.com/manual/gh_help_environment

.github/scripts/forward-port-missing/forward_port_missing.py
```diff
@@ -350,7 +351,40 @@ def determine_forward_porting_status(
...
+            env=env,
+        )
+
+    for number in sorted(to_remove):
+        info(f"Removing label from PR #{number}")
+        sub.run(
+            ["gh", "pr", "edit", str(number), "--remove-label", label],
+            check=True,
+            env=env,
+        )
```
zhijie-yang: We'd like to have an error handling here, since this action can return code 1, if the label is not exist. Adding a warning for the failed operation should suffice.

zhijie-yang (commented):
Almost good to go! Just some minor comments here.

zhijie-yang (approved):
Looks good. Thanks!

PR #621, opened 2025-08-13
ci: update compatibility matrix with newer Chisel versions

cjdcordeiro:
v1.0.0 is slow when compared with more recent versions of Chisel (example: https://github.com/canonical/chisel-releases/actions/runs/16933587495/job/47984768366?pr=577)

It is also old enough that we can restrict our compatibility matrix to the newer minor versions and `main`.

### DIFF
--- a/.github/workflows/install-slices.yaml
+++ b/.github/workflows/install-slices.yaml
@@ -24,10 +24,10 @@ env:
   ARCHES: '["amd64","arm64","armhf","ppc64el","riscv64","s390x"]'
   RELEASES: |
     [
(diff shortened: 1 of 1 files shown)


cjdcordeiro:
> Looks good to me. Qq: Do we have a policy that how many minor versions behind `main` we should explicitly support?

Not really. We discussed this throughout #509 , but for normal, it's a loose matrix based on the version's age, feature set and impact of the CI

letFunny:
What we should avoid is removing 1.1.0 which doesn't have hardlink support (https://github.com/canonical/chisel/commit/9d9d2144cf47586d02ff7566b8764e70057987a4). That is the main use-case for the matrix.

This PR is good to go, thanks Cris!

PR #610, opened 2025-08-04
feat(25.04): add rustc-1.84 slices

lczyk:
# Proposed changes

Slices for `rustc-1.84` and its deps

## Related issues/PRs

...

## Checklist

* [x] I have read the [contributing guidelines](
https://github.com/canonical/chisel-releases/blob/main/CONTRIBUTING.md)
* [x] I have tested my changes ([see how](https://github.com/canonical/chisel-releases/blob/main/CONTRIBUTING.md#7-test-your-slices-before-opening-a-pr))
* [x] I have already submitted the [CLA form](
https://ubuntu.com/legal/contributors/agreement)

## Additional Context

...

PR #598, opened 2025-07-18
feat(25.04): add golang slices

zhijie-yang:
# Proposed changes

This PR adds SDF for `golang` and its dependencies for 25.04.

## Related issues/PRs

#590 

### Forward porting

N/A

## Checklist

* [x] I have read the [contributing guidelines](
https://github.com/canonical/chisel-releases/blob/main/CONTRIBUTING.md)
* [x] I have tested my changes ([see how](https://github.com/canonical/chisel-releases/blob/main/CONTRIBUTING.md#7-test-your-slices-before-opening-a-pr))
* [x] I have already submitted the [CLA form](
https://ubuntu.com/legal/contributors/agreement)

## Additional Context

See #590

PR #633, opened 2025-09-03
feat(25.04): libc6-dev slices

lczyk:
# Proposed changes
A bunch of other PRs (see below) depend on this, and it's a bit tricky slice. This PR cherrypicks the `libc6-dev` SDF in order to make other PRs smaller.

## Related issues/PRs

This PR is a member of a chain of PRs which depend on one another. The first commit is a squashed version of everything below it. Please review the 'root' PRs first.

#633 - ROCKS-2270/libc6-dev-for-25.04 **(this PR)**
└ #644 - ROCKS-2270/binutils-for-25.04
....└ #648 - ROCKS-2270/cpp-for-25.04
........└ #649 - ROCKS-2270/gcc-14-for-25.04
............└ #610 - ROCKS-2271/rust-1.84-25.04
................└ #634 - ROCKS-2271/cargo-1.84-25.04

Cherrypicked from #610, #598

### Forward porting

#664

## Checklist

### DIFF
--- a/slices/libc6-dev.yaml
+++ b/slices/libc6-dev.yaml
@@ -255,7 +255,7 @@ slices:
       /usr/include/*-linux-*/bits/socket_type.h:
       /usr/include/*-linux-*/bits/spawn_ext.h:
       /usr/include/*-linux-*/bits/ss_flags.h:
-      /usr/include/*-linux-*/bits/stab.def:
+      /usr/include/*-linux-*/bits/stab.def:  # debug symbol definitions
       /usr/include/*-linux-*/bits/stat.h:
       /usr/include/*-linux-*/bits/statfs.h:
       /usr/include/*-linux-*/bits/statvfs.h:


slices/libc6-dev.yaml
```diff
@@ -0,0 +1,120 @@
...
+      /usr/lib/*-linux-*/libm.a:
+      /usr/lib/*-linux-*/libmcheck.a:
+      /usr/lib/*-linux-*/libmvec.a:
+        arch: [ amd64, arm64 ]
+      /usr/lib/*-linux-*/libpthread.a:
+      /usr/lib/*-linux-*/libpthread_nonshared.a:
+      /usr/lib/*-linux-*/libresolv.a:
+      /usr/lib/*-linux-*/librt.a:
+      /usr/lib/*-linux-*/libutil.a:
+        arch: [ amd64, arm64, ppc64el, riscv64, s390x]
```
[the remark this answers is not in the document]
lczyk: lol. cheers copilot. :laughing:

lczyk:
added `priority` since its related to a lot of active PRs and would help a lot with pushing them forward

slices/libcrypt-dev.yaml
```diff
@@ -0,0 +1,32 @@
...
+    contents:
+      /usr/lib/*-linux-*/libcrypt.a:
+
+  symlink-libcrypt:
+    essential:
+      - libcrypt1_libs
+
+    # All of these are symlinks to so's in the libcrypt1 package
+    contents:
+      /usr/lib/*-linux-*/libcrypt.so:
```
zhijie-yang, line 28: I believe you can just put it together with `libcrypt.a` in the same slice, and put `libcrypt1_libs` in the essential list.

slices/libc6-dev.yaml
```diff
@@ -0,0 +1,120 @@
...
+      # and if we tried to glob *.o files, this slice would conflict with oh-so-many
+
+      /usr/lib/*-linux-*/Mcrt1.o:
+      /usr/lib/*-linux-*/Scrt1.o:
+      /usr/lib/*-linux-*/crt1.o:
+      /usr/lib/*-linux-*/crti.o:
+      /usr/lib/*-linux-*/crtn.o:
+      /usr/lib/*-linux-*/gcrt1.o:
+      /usr/lib/*-linux-*/grcrt1.o:
+        arch: [ amd64, arm64, riscv64, s390x ]
```
zhijie-yang, line 29:
```suggestion
        arch: [amd64, arm64, riscv64, s390x]
```

Let's make the formatting consistent across different files. We don't put extra spaces inside the brackets.

zhijie-yang (changes requested):
Thanks for the PR. Here are some points we can talk about.

slices/libcrypt-dev.yaml
```diff
@@ -0,0 +1,32 @@
...
+      - libcrypt-dev_headers
+      - libcrypt-dev_symlink-libcrypt
+
+  headers:
+    contents:
+      /usr/include/crypt.h:
+
+  core-libs:
+    contents:
+      /usr/lib/*-linux-*/libcrypt.a:
```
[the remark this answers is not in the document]
lczyk: sure to both this and the other comment on `libcrypt-dev`. i've basically just modeled this quickly on how `libc6-dev` looks, but have no problems with putting the two together here

slices/libc6-dev.yaml
```diff
@@ -0,0 +1,135 @@
...
+      - libc6_libs
+    contents:
+      /usr/lib/*-linux-*/Scrt1.o:
+      /usr/lib/*-linux-*/crt1.o:
+      /usr/lib/*-linux-*/crti.o:
+      /usr/lib/*-linux-*/crtn.o:
+      # Note: libc.so here is an ld script that tells the linker to look up the
+      # functions from libc_nonshared.a if they are not found in the libc.so.6
+      # from libc6.
+      /usr/lib/*-linux-*/libc.so:
```
clay-lake, line 24: Should we include `ld` if we are including a linker script? I think this depends on the use case, and perhaps the user wants to use a different linker.

slices/libcrypt-dev.yaml
```diff
@@ -0,0 +1,32 @@
+package: libcrypt-dev
```
clay-lake, line 1: Should we include the package config files? 
- /usr/lib/x86_64-linux-gnu/pkgconfig/libxcrypt.pc
- /usr/lib/x86_64-linux-gnu/pkgconfig/libcrypt.pc

clay-lake (approved):
Few minor questions below but otherwise LGTM

slices/libc6-dev.yaml
```diff
@@ -0,0 +1,135 @@
...
+
+essential:
+  - libc6-dev_copyright
+
+slices:
+  copyright:
+    contents:
+      /usr/share/doc/libc6-dev/copyright:
+
+  core-libs:
```
cjdcordeiro, line 11: We need some further explanation to what this is. During a functional split, `minimum`, `core`, etc., have a meaning, as defined in the chisel-docs. `core-libs` is a new thing. I'm not saying it's wrong, but it begs the question about its purpose and relation to `libs` (if there's even one)

slices/libcrypt-dev.yaml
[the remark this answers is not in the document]
cjdcordeiro: same comment as above, and highlighting @zhijie-yang comment too about the possibly unnecessary split

cjdcordeiro (changes requested):
let's have @zhijie-yang also re-review this

slices/libcrypt-dev.yaml
lczyk: ok. added the pklgconfig slice, but not as essential of libs

slices/libc6-dev.yaml
lczyk: talked with @zhijie-yang and proposed a new version. will mark this comment as resolved and re-request review from @clay-lake too

slices/libc6-dev.yaml
```diff
@@ -0,0 +1,153 @@
...
+      - linux-libc-dev_linux-headers
+    contents:
+      /usr/include/*-linux-*/bits/**.h:
+      /usr/include/*-linux-*/fpu_control.h:
+      /usr/include/*-linux-*/gnu/**.h:
+      /usr/include/*-linux-*/ieee754.h:
+      /usr/include/*-linux-*/sys/**.h:
+      /usr/include/a*.h:
+      /usr/include/arpa/*.h:
+      /usr/include/byteswap.h:
```
alesancor1, line 100: Why are we using the glob with `a*.h`, `d*.h`, `e*.h` but directly referencing other headers like `byteswap.h`, `complex.h`, `cpio.h`, etc.? Why not just include all .h files under /include? `/usr/include/*.h`

slices/libc6-dev.yaml
```diff
@@ -0,0 +1,153 @@
+package: libc6-dev
+
+essential:
+  - libc6-dev_copyright
+
+slices:
```
lczyk, line 6: ^

slices/libc6-dev.yaml
alesancor1: Right that's a good point. However, being this the case I'd rather list all the files individually because:
* Every time someone creates a slice that puts something under `/usr/include` a conflict will happen:
   * This slice would need to be modified as part of the PR
   * Even if the new slice is not depending on this `libc6-dev` ones (meaning no conflict happens on CI), the conflicting files are still there.
* If at some point we wanted to refactor and glob this files instead, it is safer as globing will add files rather than remove them, avoiding potential regressions.

Note that this is more a limitation with chisel on how conflicts are handled, so an issue could be raised there (if not yet raised).

lczyk: i've pushed the full unglob to test it in CI while i look for some deeper paths (deeper than `/usr/include/` for which it would make sense to glob. i will push that separately so if we decide to go back to full unglob, it's just a couple of reverted commits away

lczyk:
based on the overall consensus i've renamed the `posix-libs` slice to `core` and i've added it as essential for `libutil`, `librt` etc. ( good spot @cjdcordeiro ) some of the tests for the behavior there will need to come in #644 since we don't have the linker yet, and using the host linker (from outside of the chroot) for test would be tricky and error-prone imo. will work on that now and add these to #644, as well as updating it to use the new slice names.

`core` slice represents the libraries one will always need, since they are the basic things you need to create a binary. the tests should check that + check that, for example, if we install the `libm` slice but then manually delete `core`, we cannot compile a working binary -- aka test asserting / showing that `core` libs are indeed core.

lczyk:
based on `dpkg --search` on a full installation of ubuntu 25.04 (`find -maxdepth 1 -type f -printf "$PWD/%P\n" | xargs -n1 dpkg --search`) i found that it seems to be safe to glob `bits` and `src`. i've globbed these, but left arch-specific globs explicitly specified. i also left all shallower things unglobbed.

its possible that `/usr/include/net` or `/usr/include/netinet` could be globbed too, but the savings there are mich smaller.

lczyk:
just so it does not get lost in a response to the review comment, [here](https://github.com/lczyk/libc6-dev-deb-data) is the data dump of `dpkg -c` of `libc6-dev` .h files in all supported archs.

alesancor1 (approved):
This is amazing work, seems you got to understand libc6 quite in depth, which is great! (some bits are a bit over explained in comments I'd say, but not against it)

Apart from what others said I have no further comments, LGTM! Let's get those nits fixed quick and we good to go on my end :+1:

slices/libc6-dev.yaml
```diff
@@ -0,0 +1,659 @@
...
+      /usr/lib/*-linux-*/libnss_hesiod.so:
+      /usr/lib/*-linux-*/libthread_db.so:
+
+  # All the standard C library headers
+  headers:
+    # We need `libcrypt-dev_headers` for the crypt.h header which is referenced from unistd.h
+    essential:
+      - libcrypt-dev_headers
+      - linux-libc-dev_linux-headers
+    contents:
```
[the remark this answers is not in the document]
lczyk: good catch on that btw. thanks heaps :)

zhijie-yang (approved):
Looks good to me now! Many thanks!

clay-lake (approved):
Nice work 💪

PR #635, opened 2025-09-04
feat(25.10): open ubuntu-25.10 release

cjdcordeiro:
Set the CI and README for opening `ubuntu-25.10`

### DIFF
--- a/.github/workflows/install-slices.yaml
+++ b/.github/workflows/install-slices.yaml
@@ -94,13 +94,13 @@ jobs:
           ]]; then
             # This happens when a new branch/tag is created, so we'll want to
             # install all the slices for that release only
-            export RELEASES="[{\"ref\": \"${{ github.ref_name }}\", \"chisel-versions\": \"${CHISEL_VERSIONS}\"}]"
+            export RELEASES="[{\"ref\": \"${{ github.ref_name }}\", \"chisel-versions\": ${CHISEL_VERSIONS}}]"
             MATRIX=$(./version-matrix)
             echo "matrix={\"include\": $MATRIX}" >> $GITHUB_OUTPUT
           else
             # Filter the releases to only the affected branch, then swap the ref
             # such that the correct branch is tested.
-            export RELEASES="[{\"ref\": \"\", \"chisel-versions\": \"${CHISEL_VERSIONS}\"}]"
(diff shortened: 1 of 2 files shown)


cjdcordeiro:
> Qq: Since we are adding 25.10 to the `RELEASES` env var in `install-slices.yaml`, will this keep the cron workflow `install-slices` failing until the `ubuntu-25.10` branch is present?

We're hitting a chicken-egg problem. When the new branch is created, it's gonna trigger its `ci.yaml`, which will call this workflow. So if this workflow doesn't have the new release branch, that CI will fail.

At the same time, when we create this PR, the release branch doesn't exist yet, for the PR's CI will fail :/

clay-lake (approved):
LGTM. Do we then just override the checks to merge this, then create the release branch?

.github/workflows/install-slices.yaml
```diff
@@ -75,7 +72,18 @@ jobs:
...
             # enough.
-            export RELEASES=$(echo "$RELEASES" | jq 'map(.["chisel-versions"] |= ["main"])')
+            all_branches="$(git branch --all | grep -E "remotes.*ubuntu-[0-9]{2}\.[0-9]{2}$" | cut -d '/' -f 3)"
+            for unmaintained in $UNMAINTAINED_RELEASES; do
+              all_branches="$(echo "${all_branches//$unmaintained}")"
+            done
+
+            RELEASES="[]"
+            for branch in $all_branches; do
+              RELEASES="$(echo "$RELEASES" | jq --arg ref "$branch" '. += [{"ref": $ref, "chisel-versions": ["main"]}]')"
```
zhijie-yang, line 82: In the previous implementation, the `|= ["main"]` is set as the default value of the chisel version to be tested, since in the `RELEASES` array, the `chisel-versions` can be empty. However, in the lines below, we are forcing the `chisel-versions` to be the `CHISEL_VERSIONS` defined at the beginning of this code, which renders this `"chisel-versions": ["main"]` useless.

cjdcordeiro:
> > The changes look good to me, but do you want to try and run the checks again to ensure everything passes?
> 
> y I've been re-running

@clay-lake they are passing now

.github/workflows/install-slices.yaml
zhijie-yang: Ah I see. I've missed the `elif` on line 91 below... Then everything looks good.

PR #592, opened 2025-07-16
feat(24.04): add php slices

alesancor1:
# Proposed changes
Add slice definition files for php 8.3 on 24.04

### Forward porting
* https://github.com/canonical/chisel-releases/pull/749
* #594 
* https://github.com/canonical/chisel-releases/pull/593

## Checklist

* [x] I have read the [contributing guidelines](
https://github.com/canonical/chisel-releases/blob/main/CONTRIBUTING.md)
* [x] I have tested my changes ([see how](https://github.com/canonical/chisel-releases/blob/main/CONTRIBUTING.md#7-test-your-slices-before-opening-a-pr))
* [x] I have already submitted the [CLA form](
https://ubuntu.com/legal/contributors/agreement)

PR #637, opened 2025-09-05
feat(25.04): tomcat10 slices

vpa1977:
# Proposed changes

## Checklist

* [ ] I have read the [contributing guidelines](
* [ ] I have tested my changes ([see how](https://github.com/canonical/chisel-releases/blob/main/CONTRIBUTING.md#7-test-your-slices-before-opening-a-pr))
* [ ] I have already submitted the [CLA form](

### DIFF
--- a/slices/libeclipse-jdt-core-java.yaml
+++ b/slices/libeclipse-jdt-core-java.yaml
@@ -1,0 +1,13 @@
+package: libeclipse-jdt-core-java
+
+essential:
+  - libeclipse-jdt-core-java_copyright
+
+slices:
+  libs:
+    contents:
+      /usr/share/java/eclipse-jdt-core*.jar:
+
+  copyright:
+    contents:
+      /usr/share/doc/libeclipse-jdt-core-java/copyright:
--- a/slices/tomcat10-common.yaml
+++ b/slices/tomcat10-common.yaml
@@ -1,0 +1,62 @@
+package: tomcat10-common
+
+essential:
+  - tomcat10-common_copyright
+
+slices:
+  # additional shell scripts
+  bins:
+    essential:
+      - tomcat10-common_core
+    contents:
+      /usr/share/tomcat10/bin/ciphers.sh:
+      /usr/share/tomcat10/bin/configtest.sh:
+      /usr/share/tomcat10/bin/daemon.sh:
+      /usr/share/tomcat10/bin/digest.sh:
+      /usr/share/tomcat10/bin/makebase.sh:
+      /usr/share/tomcat10/bin/migrate.sh:
+      /usr/share/tomcat10/bin/shutdown.sh:
+      /usr/share/tomcat10/bin/startup.sh:
+      /usr/share/tomcat10/bin/tool-wrapper.sh:
+      /usr/share/tomcat10/bin/version.sh:
+
+  core:
+    essential:
+      - libtomcat10-java_i18n-common
+      - libtomcat10-java_libs
+    contents:
+      /usr/share/tomcat10/bin/bootstrap.jar:
+      /usr/share/tomcat10/bin/catalina-tasks.xml:
+      /usr/share/tomcat10/bin/catalina.sh:
+      /usr/share/tomcat10/bin/setclasspath.sh:
+      /usr/share/tomcat10/bin/tomcat-juli.jar:
+      /usr/share/tomcat10/lib/annotations-api.jar:
+      /usr/share/tomcat10/lib/catalina-ant.jar:
+      /usr/share/tomcat10/lib/catalina-ha.jar:
+      /usr/share/tomcat10/lib/catalina-storeconfig.jar:
+      /usr/share/tomcat10/lib/catalina-tribes.jar:
+      /usr/share/tomcat10/lib/catalina.jar:
+      /usr/share/tomcat10/lib/el-api.jar:
+      /usr/share/tomcat10/lib/jasper-el.jar:
+      /usr/share/tomcat10/lib/jasper.jar:
+      /usr/share/tomcat10/lib/jaspic-api.jar:
+      /usr/share/tomcat10/lib/jsp-api.jar:
+      /usr/share/tomcat10/lib/servlet-api.jar:
+      /usr/share/tomcat10/lib/tomcat-api.jar:
+      /usr/share/tomcat10/lib/tomcat-coyote.jar:
+      /usr/share/tomcat10/lib/tomcat-dbcp.jar:
+      /usr/share/tomcat10/lib/tomcat-i18n-es.jar:
+      /usr/share/tomcat10/lib/tomcat-i18n-fr.jar:
+      /usr/share/tomcat10/lib/tomcat-i18n-ja.jar:
+      /usr/share/tomcat10/lib/tomcat-i18n-ru.jar:
+      /usr/share/tomcat10/lib/tomcat-jdbc.jar:
+      /usr/share/tomcat10/lib/tomcat-jni.jar:
+      /usr/share/tomcat10/lib/tomcat-util-scan.jar:
+      /usr/share/tomcat10/lib/tomcat-util.jar:
+      /usr/share/tomcat10/lib/tomcat-websocket.jar:
+      /usr/share/tomcat10/lib/websocket-api.jar:
+      /usr/share/tomcat10/lib/websocket-client-api.jar:
+
+  copyright:
+    contents:
+      /usr/share/doc/tomcat10-common/copyright:
--- a/slices/tomcat10.yaml
+++ b/slices/tomcat10.yaml
@@ -1,0 +1,56 @@
+package: tomcat10
+
+essential:
+  - tomcat10_copyright
+
+slices:
+  core:
+    essential:
+      - tomcat10-common_core
+    contents:
+      /etc/tomcat10/Catalina/:
+        make: true
+      /etc/tomcat10/catalina.properties:
+        symlink: /usr/share/tomcat10/etc/catalina.properties
+      /etc/tomcat10/context.xml:
+        symlink: /usr/share/tomcat10/etc/context.xml
+      /etc/tomcat10/jaspic-providers.xml:
+        symlink: /usr/share/tomcat10/etc/jaspic-providers.xml
+      /etc/tomcat10/logging.properties:
+        symlink: /usr/share/tomcat10/etc/logging.properties
+      /etc/tomcat10/policy.d/01system.policy:
+      /etc/tomcat10/policy.d/02debian.policy:
+      /etc/tomcat10/policy.d/03catalina.policy:
+      /etc/tomcat10/policy.d/04webapps.policy:
+      /etc/tomcat10/policy.d/50local.policy:
+      /etc/tomcat10/server.xml:
+        symlink: /usr/share/tomcat10/etc/server.xml
+      /etc/tomcat10/tomcat-users.xml:
+        symlink: /usr/share/tomcat10/etc/tomcat-users.xml
+      /etc/tomcat10/web.xml:
+        symlink: /usr/share/tomcat10/etc/web.xml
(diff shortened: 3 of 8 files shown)


slices/libeclipse-jdt-core-java.yaml
```diff
@@ -0,0 +1,15 @@
...
+
+essential:
+  - libeclipse-jdt-core-java_copyright
+
+slices:
+  libs:
+    essential:
+      - libeclipse-jdt-core-compiler-batch-java_libs
+    contents:
+      /usr/share/java/eclipse-jdt-core-3.*.jar:
```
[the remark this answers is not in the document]
vpa1977: This will create a conflict with `/usr/share/java/eclipse-jdt-core-compiler-batch-*.jar:`

slices/libeclipse-jdt-core-compiler-batch-java.yaml
```diff
@@ -0,0 +1,13 @@
+package: libeclipse-jdt-core-compiler-batch-java
+
+essential:
+  - libeclipse-jdt-core-compiler-batch-java_copyright
+
+slices:
+  libs:
+    contents:
+      /usr/share/java/eclipse-jdt-core-compiler-batch-3.*.jar:
```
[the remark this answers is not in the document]
vpa1977: The `3.` is added to be in line with `slices/libeclipse-jdt-core-java.yaml` definition to resolve the conflict.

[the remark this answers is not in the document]
clay-lake: I see, that is an unfortunate collision. 😄

tests/spread/integration/tomcat10-common/task.yaml
```diff
@@ -0,0 +1,16 @@
+summary: Integration tests for tomcat10-common
+
+execute: |
+  # tomcat needs an openjdk and shell
+  rootfs="$(install-slices tomcat10-common_core base-files_base coreutils_bins dash_bins openjdk-21-jdk-headless_standard)"
```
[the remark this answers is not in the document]
vpa1977: This can not be done as tomcat can run on openjdk-8...-25.  The user should be able to configure openjdk to use. 

Dash and coreutils is a choice for testing, the user may also opt for busybox to provide shell and coreutils. Depending on the rock we may have one or another.

slices/tomcat10.yaml
```diff
@@ -0,0 +1,56 @@
...
+      /var/cache/tomcat10/:
+        make: true
+      /var/lib/tomcat10/conf:
+        symlink: /etc/tomcat10
+      /var/lib/tomcat10/lib/:
+        make: true
+      /var/lib/tomcat10/logs:
+        symlink: /var/log/tomcat10/
+      /var/lib/tomcat10/webapps/:
+        make: true
```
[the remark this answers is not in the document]
vpa1977: `error: cannot extract from package "tomcat10": no content at /var/lib/tomcat10/webapps`

tests/spread/integration/tomcat10-common/task.yaml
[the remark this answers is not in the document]
vpa1977: As discussed. I have created a minimal slice without dependencies, and the core one with a set of default dependencies.

lczyk:
nice. looks good on first glance. once minor thing i wanna check (dependencies of scripts) but will be a fery fast 'approve' after that :)

alesancor1 (commented):
This is almost ready to go, though I suggest some changes when splitting `tomcat10` and `tomcat10-common`, please see below whether my approach makes sense for you:

tests/spread/integration/tomcat10-common/task.yaml
```diff
@@ -0,0 +1,39 @@
+summary: Integration tests for tomcat10-common
+
+environment:
+  SLICE/core: "core"
```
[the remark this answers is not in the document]
vpa1977: Probably the whole test will be gone and replaced by tests in tomcat10 slice.

slices/tomcat10-common.yaml
```diff
@@ -0,0 +1,89 @@
...
+
+# NOTE: all the jar files in are symlinks
+# to their corresponding jars from libtomcat10-java except
+# /usr/share/tomcat10/bin/bootstrap.jar
+
+slices:
+  minimal:
+    essential:
+      - tomcat10-common_jars
+      - tomcat10-common_scripts
```
alesancor1, line 14: As mentioned before, there is no need for `minimal`/`core`/`standard`/`dev` separation here as this package only provides files for the parent `tomcat10` package, so it is enough with having that separation there (in the parent) pointing to the slices here in `essential`

slices/tomcat10-common.yaml
```diff
@@ -0,0 +1,89 @@
...
+
+  scripts:
+    contents:
+      /usr/share/tomcat10/bin/catalina.sh:
+      /usr/share/tomcat10/bin/setclasspath.sh:
```
alesancor1, line 19: These should be functional, so I'd add the required slices from `coreutils`, `dash`, `grep`, etc. here

slices/tomcat10-common.yaml
```diff
@@ -0,0 +1,89 @@
...
+    essential:
+      - libtomcat10-java_i18n
+    contents:
+      /usr/share/tomcat10/lib/tomcat-i18n-es.jar:
+      /usr/share/tomcat10/lib/tomcat-i18n-fr.jar:
+      /usr/share/tomcat10/lib/tomcat-i18n-ja.jar:
+      /usr/share/tomcat10/lib/tomcat-i18n-ru.jar:
+
+  # additional shell scripts
+  extra-scripts:
```
alesancor1, line 59: Same here, this should be functional, so if `coreutils`, `grep`, `debianutils`, etc. is required, add them here.

slices/tomcat10.yaml
```diff
@@ -0,0 +1,77 @@
...
+      - coreutils_expr
+      - coreutils_output-of-entire-files
+      - coreutils_rm-utility
+      - coreutils_special-file-types
+      - coreutils_system-context
+      - coreutils_touch
+      - coreutils_working-context
+      - dash_bins
+      - debianutils_which
+      - grep_bins
```
alesancor1, line 27: Following above comments, if these are in the scripts already they can be removed from here.

alesancor1 (commented):
Getting close to (what I believe) would be the ideal approach, just clarifying some of the comments I've left before:

slices/tomcat10-common.yaml
vpa1977: That's actually breaks the intended purpose of those slices - provide tomcat files without extra dependencies, so that those could be set in the rock. 
- tomcat10_minimal - non-functional slice that provides tomcat10 files
- tomcat10_core - functional slice with default dependencies.

slices/tomcat10-common.yaml
vpa1977: This one is useful for the scenario when I am building the rock and just need CATALINA_HOME files.

slices/tomcat10-common.yaml
alesancor1: This is a really similar case to this one in `php` slices, we ended up adding those dependencies despite the `php-common` package serving the same purpose as this one, providing files. Check this [conversation](https://github.com/canonical/chisel-releases/pull/592#discussion_r2226049352)

slices/tomcat10-common.yaml
alesancor1: In that case you can use the parent `minimal` slice for it, or just add those two slices to your rock can't you?

slices/tomcat10-common.yaml
vpa1977: This still wont make the slice functional as tomcat will not start without configuration defined in tomcat10 package. 
@cjdcordeiro

lczyk: > provide tomcat files without extra dependencies, so that those could be set in the rock.

sure, but the basic dependencies to run the script must be set here. `catalina.sh` depends on:

```
coreutils: /usr/bin/cat
coreutils: /usr/bin/dirname
coreutils: /usr/bin/expr
coreutils: /usr/bin/ls
coreutils: /usr/bin/mkfifo
coreutils: /usr/bin/rm
coreutils: /usr/bin/sleep
coreutils: /usr/bin/touch
coreutils: /usr/bin/uname
dash: /usr/bin/dash
grep: /usr/bin/grep
procps: /usr/bin/ps
```

and `setclasspath.sh` depends on:

```
coreutils: /usr/bin/dirname
dash: /usr/bin/dash
debianutils: /usr/bin/which
```

without those, yes, you are allowing the user to install these deps however and from whatever they want, but you're also asking them to figure all of them out and do that installation. At the very least you end up with something like:

```bash
chisel cut --rootfs . coreutils_bins dash_bins grep_bins procps_bins debianutils_which tomcat10-common_scripts ...
```

just to install the scripts, and that's not even as good as it can be because, for example, `coreutils_bins` brings in all the `coreutils` (100+) while you need 9, so now the user needs to go and figure all of that out and filter out the unneeded files either with some clever `find` or through the stage/prime filter or by finding the subslices.

it would be waaay better if the script slices had all the basic deps in them, and sure, they might not be fully functional, but at least they have all the basic stuff to run and the user can just do

```bash
chisel cut --rootfs . tomcat10-common_scripts ...
```

and if you're also cutting coreutils_bins anyway, no loss. Chisel will handle that just fine.

it's a balance really: on one hand side there is how much freedom we want to give to users, and on the other hand side how prescriptive we want to be. at the end of the day, rest assured, entrepreneurial user *will* get their way in the end and install all the things however they please, and we're not here to bar them from it. we do want to provide tastefully crafted path of least resistance.

btw, [here](https://github.com/lczyk/rocks-toolbox/tree/bash_dep_finder/bash_dep_finder) is my bash dependency analysis tool. not anything official and might change / move without notice so don't build your CIs with it ;))

lczyk (changes requested):
i agree with @alesancor1 re slice functionality. i've left a big comment here: https://github.com/canonical/chisel-releases/pull/637#discussion_r2402539692

slices/tomcat10-common.yaml
vpa1977: > it's a balance really: on one hand side there is how much freedom we want to give to users, and on the other hand side how prescriptive we want to be. at the end of the day, rest assured, entrepreneurial user will get their way in the end and install all the things however they please, and we're not here to bar them from it. we do want to provide tastefully crafted path of least resistance.

I am finding myself in the shoes of that user. But lets do it this way - move dependencies here from tomcat10 slices. 

In the rock I can chisel cut into a temporary directory and then copy out only the things that I need. 
Definitely not the most elegant solution, but I do not foresee tomcat packaging changing installation directories.

lczyk: > I am finding myself in the shoes of that user.

i really don't see why that woudl be the case here.. is it that you want to install `catalina.sh` and `setclasspath.sh` but not have any of the bits to run it?

> In the rock I can chisel cut into a temporary directory and then copy out only the things that I need.

You can use stage [filesets](https://documentation.ubuntu.com/rockcraft/stable/explanation/filesets/) for that too. Just FYI, in case you need to do these kind of operations on your rock.

alesancor1 (approved):
Thanks for the changes, the slicing LGTM.

As mentioned in https://github.com/canonical/chisel-releases/pull/637#discussion_r2405654372 I'm approving this with the caveat of `tomcat10-common_scripts` and `tomcat10-common_extra-scripts` not being functional out of the box.

I'll mark this as ready for merge, though @cjdcordeiro has the last word :+1:

slices/tomcat10-common.yaml
vpa1977: Ok, lets go with the functional approach with caveats described here[1].

[1] https://github.com/canonical/chisel-releases/pull/637#issuecomment-3378507555

PR #634, opened 2025-09-03
feat(25.04): cargo-1.84 slices

lczyk:
# Proposed changes
This pr introduces slices for cargo-1.84.

Internal ticket number ROCKS-2422.

## Related issues/PRs

This PR is continuously rebased on top of #633. #633 should be reviewed and merged first.

This PR is related to #610

### Forward porting
N/A

## Checklist

* [x] I have read the [contributing guidelines](
https://github.com/canonical/chisel-releases/blob/main/CONTRIBUTING.md)
* [x] I have tested my changes ([see how](https://github.com/canonical/chisel-releases/blob/main/CONTRIBUTING.md#7-test-your-slices-before-opening-a-pr))
* [x] I have already submitted the [CLA form](
https://ubuntu.com/legal/contributors/agreement)

PR #644, opened 2025-09-11
feat(25.04): binutils slices

lczyk:
# Proposed changes
Slices of binutils assembler and linker cherrypicked from #598.

## Related issues/PRs

This PR is a member of a chain of PRs which depend on one another. The first commit is a squashed version of everything below it. Please review the 'root' PRs first.

~~#633 - ROCKS-2270/libc6-dev-for-25.04~~ *(merged)*
└ #644 - ROCKS-2270/binutils-for-25.04 **(this PR)**
....└ #648 - ROCKS-2270/cpp-for-25.04
........└ #649 - ROCKS-2270/gcc-14-for-25.04
............└ #610 - ROCKS-2271/rust-1.84-25.04
................└ #634 - ROCKS-2271/cargo-1.84-25.04

### Forward porting

- https://github.com/canonical/chisel-releases/pull/703

## Checklist


alesancor1 (commented):
Looks good overall, just a couple of nitpicks down:

*Will probably require a second pass once #633 is merged

slices/binutils-aarch64-linux-gnu.yaml
```diff
@@ -0,0 +1,109 @@
+package: binutils-aarch64-linux-gnu
+
+essential:
+  - binutils-aarch64-linux-gnu_copyright
+
+slices:
+  # NOTE: You might want to create a symlink from /usr/bin/as to
+  #       /usr/bin/aarch64-linux-gnu-as in your rootfs to use these slices.
+  #       ln -s "/usr/bin/${arch}-as" "${rootfs}/usr/bin/as"
```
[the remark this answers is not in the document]
lczyk: it would normally be created by symlink in `binutils` package, but that's tricky/impossible to do at the moment for multiarch. the way to use ls/as/ar is therefore to manually figure out arch-sepcific slice name and then make the symlink oneself. see, e.g. [the override-build for the rust rock](https://github.com/canonical/rust-rock/blob/ROCKS-2271/rust-rock-25.04/rust-rock/1.84-25.04/rust_override_build.sh)).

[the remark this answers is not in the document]
lczyk: the situation is analogous for cpp (#648) and gcc (#649)

alesancor1 (approved):
LGTM, please create the FP for it

slices/binutils-aarch64-linux-gnu.yaml
```diff
@@ -0,0 +1,87 @@
...
+  assembler:
+    essential:
+      - libbinutils_libs
+      - libc6_libs
+      - libzstd1_libs
+      - zlib1g_libs
+    contents:
+      /usr/bin/aarch64-linux-gnu-as:
```
lengau, line 17: Relevant to the comment above, what about the symlinks for these binaries? It's very nice to be able to prepend `$PATH` for cases like this.

```suggestion
      /usr/bin/aarch64-linux-gnu-as:
      /usr/aarch64-linux-gnu/bin/as:
```

slices/binutils-aarch64-linux-gnu.yaml
```diff
@@ -0,0 +1,87 @@
+package: binutils-aarch64-linux-gnu
+
+essential:
+  - binutils-aarch64-linux-gnu_copyright
+
+slices:
+  # NOTE: You might want to create a symlink from /usr/bin/as to
+  #       /usr/bin/aarch64-linux-gnu-as in your rootfs to use these slices.
+  #       ln -s "/usr/bin/${arch}-as" "${rootfs}/usr/bin/as"
```
lengau, line 9:
```suggestion
  #       ln -s "${arch}-as" "${rootfs}/usr/bin/as"
```

Should this be a relative link?

slices/binutils-common.yaml
lengau, line 1: As an aside, the fact that `binutils-common` isn't built for `Architecture: all` but is instead built for each architecture (despite the contents being identical) is surprising. I'm sure there's some deep, interesting Debian lore about that.

tests/spread/integration/binutils/hello-aarch64-linux-gnu.S
```diff
@@ -0,0 +1,33 @@
...
+	mov	w0, 0
+	ldp	x29, x30, [sp], 16
+	.cfi_restore 30
+	.cfi_restore 29
+	.cfi_def_cfa_offset 0
+	ret
+	.cfi_endproc
+.LFE0:
+	.size	main, .-main
+	.ident	"GCC: (Ubuntu 13.3.0-6ubuntu2~24.04) 13.3.0"
```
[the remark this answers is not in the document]
lczyk: nah :sweat_smile: 

```bash
cat > main.c << EOF
#include<stdio.h>
int main() { printf(\"hi\"); return 1; }
EOF
```

and then

```
gcc -S main.c
```

or

```
/usr/libexec/gcc/aarch64-linux-gnu/14/cc1 -I/usr/include/aarch64-linux-gnu/ main.c
```

for style points (might need to adjust paths for non-arch) :v:

but yeah, hand-rolled assembly would definitely be way more style points.

alesancor1:
> @alesancor1 presumably [canonical/rust-rock#3 (comment)](https://github.com/canonical/rust-rock/pull/3#discussion_r2381694228) and [canonical/rocks-template#18](https://github.com/canonical/rocks-template/pull/18) PR also apply to this PR (and the ones which follow) ?

We have not yet decided about it for chisel-releases. If we go for this layout we should have it changed for all the other slices, for consistency - To be discussed

lczyk:
@lengau incorporated your feedback. decided to add `ldscripts` (doe snot collide with anything so just globbed) and bfd-plugins too. the `ldscripts` slice does not have the linker as essential on purpose, since the scripts *are* functional without the gnu linker -- they can be used by a different linker.

tests/spread/integration/binutils/run.sh
zhijie-yang, line 1: I'm personally against using this entry point for the spread test.

Firstly, this pattern will make the `run.sh` file duplicated in many other test directories. If we run into including this entry point, this file should be in `tests/spread/lib` instead of individual tests.

Secondly, we already have `spread` as the backend, and there is no reason for us to include a new wheel in the project, which serves a similar purpose to `spread`.

Thirdly, this script has no real effect when run in `spread`. These prettified texts won't get output when the `spread` test runs normally, and all the tests are gathered as only one case, which is a regression from the previous user experience we have.

```
2025-10-07 17:37:38 Preparing lxd:ubuntu-plucky (lxd:ubuntu-plucky)...
2025-10-07 17:37:39 Preparing lxd:ubuntu-plucky:tests/spread/integration/cpp-14 (lxd:ubuntu-plucky)...
2025-10-07 17:37:39 Executing lxd:ubuntu-plucky:tests/spread/integration/cpp-14 (lxd:ubuntu-plucky) (1/1)...
2025-10-07 17:38:47 Discarding lxd:ubuntu-plucky...
```

tests/spread/integration/binutils/setup.sh
zhijie-yang, line 1: Same comment for this file as for `run.sh`. We already have `spread` here. There is no need to include this wheel.

tests/spread/integration/binutils/task.yaml
```diff
@@ -0,0 +1,2 @@
+summary: Integration tests for binutils
+execute: ./run.sh --spread
```
zhijie-yang, line 2: What is the blocker of using the following pattern? In this way, `spread` can enumerate and report the tests correctly, which is a better and long-used UX.

```
environment:
    SLICE/AR: ar
...
execute: |
    # setup
    arch=...
    rootfs=$(install-slices binutils-${arch}_${SLICE})
    case ${SLICE} in
        ar)
            ./test_archiver.sh
            ./test_archiver_cross.sh
        ;;
```

tests/.gitignore
zhijie-yang, line 1: Why this? `spread` is expected not to pollute the test code directories.

zhijie-yang (changes requested):

tests/spread/integration/binutils/task.yaml
lczyk: i'm trying to remove excessive control logic from test files. i've talked this over with @alesancor1 already and i long-term this was going to be refactored to something more like [this](https://github.com/canonical/rust-rock/tree/main/rust/1.84-25.04/spread/general) but at the moment we only support one task.yaml in spread tests: https://github.com/canonical/chisel-releases/issues/701

having said that, will refactor this PR to use the [variants](https://github.com/canonical/spread/?tab=readme-ov-file#variants) as you suggest for now, but there is definitely more discussion to be had.

to collect some thought together: i think, for example calling `install_slices` directly in the `execute` snippet is an anti-pattern. i think that `task.yaml` should, most of the time, immediately pop out to a script file (as I suggested [here](https://github.com/canonical/rocks-template/pull/18/files)) and do it's logic there. tests like [this](https://github.com/canonical/chisel-releases/blob/03e6da2271a2a0f2fbe7ec4c6f2024593da6a540/tests/spread/integration/coreutils/smoke.sh#L16) are missing part of the point of testing -- to have a bunch of clear and simply written out cases / examples which are obvious to read.

anyway, just some thoughts for a different discussion. again, as to the review comment, will refactor to use variants :+1: (and yeah, run.sh and setup.sh probably ought not to be here)

tests/spread/integration/binutils/run.sh
lczyk: yep. all good points. already replied below but tldr; 1) will change and 2) more team discussion at some point :)

tests/spread/integration/binutils/task.yaml
lengau: I dislike using `install_slices` inside of `execute`, but for different reasons. We have installability tests already, so these spread tests should take for granted that the slice is installable.

Ideally, the entire suite would have a `prepare-each` and `restore-each` that would create and delete the chroot directory, so the actual execution would simply be able to run the script for that variant.

This would hide a certain amount of "magic" in spread.yaml, but updating the documentation should be sufficient to handle that. In the end, I imagine we could have a `task.yaml` that looks something like the follows:

```yaml
summary: Integration tests for binutils

environment:
  SLICE/ar,ar_cross: binutils-${TARGET_ARCH_TRIPLET}_archiver
  EXTRA_SLICES/ar_cross: binutils-${TARGET_ARCH_TRIPLET}_cross-libbfd

execute: bash "test_${SPREAD_VARIANT}.sh"
```

PR #732, opened 2025-10-21
chore(22.04): use v3-essential for rabbitmq-server

zhijie-yang:
# Proposed changes

This PR uses the `v3-essential` to include the `erlang` dependencies of `rabbitmq-server`, since `erlang` is not available to `i386`, while `rabbitmq-server` is available to all the architectures.

### Forward porting

#733 
#734 (closed, EOL)
#735 

## Checklist

* [x] I have read the [contributing guidelines](
https://github.com/canonical/chisel-releases/blob/main/CONTRIBUTING.md)
* [x] I have tested my changes ([see how](https://github.com/canonical/chisel-releases/blob/main/CONTRIBUTING.md#7-test-your-slices-before-opening-a-pr))
* [x] I have already submitted the [CLA form](
https://ubuntu.com/legal/contributors/agreement)

## Additional Context

PR in Chisel: https://github.com/canonical/chisel/pull/246

PR #735, opened 2025-10-21
chore(25.10): use v3-essential for rabbitmq-server

zhijie-yang:
# Proposed changes

This PR uses the `v3-essential` to include the `erlang` dependencies of `rabbitmq-server`, since `erlang` is not available to `i386`, while `rabbitmq-server` is available to all the architectures.

## Related issues/PRs

#732 
#733 
#734 

### Forward porting

N/A

## Checklist

## Additional Context

PR in Chisel: https://github.com/canonical/chisel/pull/246

### DIFF
--- a/slices/fonts-ipafont-gothic.yaml
+++ b/slices/fonts-ipafont-gothic.yaml
@@ -1,0 +1,17 @@
+package: fonts-ipafont-gothic
+
+essential:
+  - fonts-ipafont-gothic_copyright
+
+slices:
+  fonts:
+    contents:
+      /usr/share/fonts/opentype/ipafont-gothic/*.ttf:
+
+  config:
+    contents:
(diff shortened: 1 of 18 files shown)


upils:
@zhijie-yang I suspect the list of slices explicitly installed in the spread test in `tests/spread/integration/rabbitmq-server` can be simplified. Is that right?

letFunny:
> I suspect the list of slices explicitly installed in the spread test in tests/spread/integration/rabbitmq-server can be simplified. Is that right?

For the test to work in prior versions they need the explicit dependencies. For the new version we don't. I would suggest grepping (yq or Python or grep) for the format version and then using one list or the other.

As we discussed, for the test to be meaningful with v3-essential we need to remove the list of hardcoded dependencies to know if they are installed (in fact, they aren't!).

PR #736, opened 2025-10-22
feat(25.10): Add kill-timeout to integration test suite

clay-lake:
PRs with more extensive changes sometimes fail integration tests as the duration of the tests exceed the (kill-)timeout limit of spread. This series of PRs adjusts the timeout from 15m (default) to 1h. 

I added a trivial change to the coreutils slice to trigger integration tests; that SDF is known to cause a timeout. However it appears it may be broken since that release is still in devel. Tests were triggered on 25.04 (#741 ) instead.

Example failure:
- https://github.com/canonical/chisel-releases/pull/693
- https://github.com/canonical/chisel-releases/actions/runs/18382290120/job/53013645257?pr=693

PR in series:
- #737 
- #738 
- #739 
- #740 
- #741 (see checks for example test of `coreutils`)

PR #737, opened 2025-10-22
feat(20.04): Add kill-timeout to integration test suite

clay-lake:
Part of a series of PRs to mitigate integration test timeout. 
See description in #736


lczyk:
let's merge this, but i'd like to remove this in the future (aka go back to 1/2h task timeout) and split large tests like this into multiple tasks.

PR #741, opened 2025-10-22
feat(25.04): Add kill-timeout to integration test suite

clay-lake:
Part of a series of PRs to mitigate integration test timeout.
See description in https://github.com/canonical/chisel-releases/pull/736

### DIFF
--- a/slices/coreutils.yaml
+++ b/slices/coreutils.yaml
@@ -254,8 +254,7 @@ slices:
     contents:
       /usr/bin/tee:
 
-  # File name manipulation -- This slice contains commands that manipulate file
-  # names.
+  # File name manipulation -- This slice contains commands that manipulate file names.
   # https://www.gnu.org/software/coreutils/manual/html_node/File-name-manipulation.html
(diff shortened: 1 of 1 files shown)


spread.yaml
```diff
@@ -101,3 +101,5 @@ prepare-each: chisel version
 suites:
   tests/spread/integration/:
     summary: Tests common scenarios
+
+kill-timeout: 1.5h
```
lengau: This is fine, but maybe we should also consider a default per-test timeout. Most tests shouldn't take more than a minute or two, so it would be nice to know if a specific test is expected to take longer (by having it override its timeout)

clay-lake: @zhijie-yang @lczyk What do you think? should we bump this to the per task timeout? Makes sense for scalability. How long should we set it to?

Also it is not clear from docs if the timeout then must then be set on both the task and project level.

https://github.com/canonical/spread?tab=readme-ov-file#timeouts

lczyk (approved):

PR #813, opened 2025-12-15
slices(26.04): add cloud-init and its dependencies

Meulengracht:
# Proposed changes

## Checklist

* [ ] I have read the [contributing guidelines](
* [ ] I have tested my changes ([see how](https://github.com/canonical/chisel-releases/blob/main/CONTRIBUTING.md#7-test-your-slices-before-opening-a-pr))
* [ ] I have already submitted the [CLA form](

### DIFF
--- a/slices/cloud-guest-utils.yaml
+++ b/slices/cloud-guest-utils.yaml
@@ -1,14 +1,14 @@
 package: cloud-guest-utils
 
 essential:
-  - cloud-guest-utils_copyright
+  cloud-guest-utils_copyright:
 
 slices:
   bins:
     essential:
-      - e2fsprogs_standard
-      - fdisk_bins
-      - python3_standard
+      e2fsprogs_standard:
+      fdisk_bins:
+      python3_standard:
     contents:
       /usr/bin/ec2metadata:
       /usr/bin/growpart:
--- a/slices/cloud-init-base.yaml
+++ b/slices/cloud-init-base.yaml
@@ -1,18 +1,18 @@
 package: cloud-init-base
 
 essential:
-  - cloud-init-base_copyright
+  cloud-init-base_copyright:
 
 slices:
   # Standard installation - includes all components
   standard:
     essential:
-      - cloud-init-base_config
-      - cloud-init-base_network
-      - cloud-init-base_services
-      - cloud-init-base_templates
-      - cloud-init-base_udev
-      - cloud-init-base_utils
+      cloud-init-base_config:
+      cloud-init-base_network:
+      cloud-init-base_services:
+      cloud-init-base_templates:
+      cloud-init-base_udev:
+      cloud-init-base_utils:
 
   # Configuration files
   config:
@@ -26,7 +26,7 @@ slices:
 
   cli-tools:
     essential:
-      - libc6_libs
+      libc6_libs:
     contents:
       /usr/bin/cloud-init:
       /usr/bin/cloud-init-per:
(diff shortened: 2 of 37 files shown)


slices/cloud-guest-utils.yaml
```diff
@@ -0,0 +1,19 @@
...
+
+essential:
+  - cloud-guest-utils_copyright
+
+slices:
+  bins:
+    essential:
+      - e2fsprogs_standard
+      - fdisk_bins
+      - python3_standard
```
alesancor1: Do we need the entire `python3_standard` here? Or could we use `python3_core` with other slices from `libpython3-stdlib`? Check the [python3_standard slice](https://github.com/canonical/chisel-releases/blob/1d12028b0744a2f5396343bd78cfd798a619fe3f/slices/python3.yaml#L13) for reference

slices/cloud-init-base.yaml
```diff
@@ -0,0 +1,120 @@
+package: cloud-init-base
+
+essential:
+  - cloud-init-base_copyright
+
+slices:
+  # Core cloud-init binaries
+  bins:
+    essential:
+      - libc6_libs
```
alesancor1: There is no harm in explicitly specify `libc6_libs` here but it's probably already included by some of the slices below, so it can be removed.

slices/cloud-init-base.yaml
```diff
@@ -0,0 +1,120 @@
...
+    contents:
+      /usr/lib/cloud-init/ds-identify:
+      /usr/lib/cloud-init/hook-hotplug:
+      /usr/lib/cloud-init/uncloud-init:
+      /usr/lib/cloud-init/write-ssh-key-fingerprints:
+
+  # Python libraries
+  python-lib:
+    essential:
+      - python3_standard
```
alesancor1: Same here, maybe we can use `python3_core` instead with just the required modules from the stdlib

slices/cloud-init.yaml
```diff
@@ -0,0 +1,15 @@
+package: cloud-init
+
+essential:
+  - cloud-init_copyright
+
+slices:
+  bins:
```
alesancor1: This should be `standard` since you are essentially installing `cloud-init-base_standard`.

Also consider your use case, if you believe a `core` slice (which wraps `cloud-init-base_core` + other deps) would fulfill your purpose then add it, otherwise leave it as is with `standard` only

slices/dhcpcd-base.yaml
```diff
@@ -0,0 +1,59 @@
...
+      /usr/lib/dhcpcd/dhcpcd-hooks/20-resolv.conf:
+      /usr/lib/dhcpcd/dhcpcd-hooks/30-hostname:
+      /usr/lib/dhcpcd/dhcpcd-run-hooks:
+
+  hooks-network:
+    essential:
+      - dhcpcd-base_hooks-core
+    contents:
+      /usr/share/dhcpcd/hooks/10-wpa_supplicant:
+      /usr/share/dhcpcd/hooks/29-lookup-hostname:
```
alesancor1: This feels off, why are some hooks in `/usr/share` and others in `/usr/lib`? They seem to be following some priority order in its naming, so I would expect them to be symlinked somewhere but I cannot see any links being created in maintainer scripts.

slices/e2fsprogs.yaml
```diff
@@ -0,0 +1,131 @@
...
+  # Standard installation - includes all common tools
+  standard:
+    essential:
+      - e2fsprogs_fsck
+      - e2fsprogs_mkfs
+      - e2fsprogs_tune
+      - e2fsprogs_attr
+      - e2fsprogs_debug
+      - e2fsprogs_advanced
+      - e2fsprogs_services
```
alesancor1: Is it needed?

alesancor1 (changes requested):
Hey, thanks for the contribution! Completed a first pass, leaving some general comments here:

* I can see you are including `standard` in many of your SDFs to kinda replicate the standard package installation. However, our convention is usually to slice for the specific use case. So if your use case doesn't really need `standard` then there's no need to include it, we can always group into new slices or further split the existing ones in the future, when there's a use case for it.

* Regarding testing, I think we could further test the python packages at least

Slice specific comments + suggestions below:

slices/dhcpcd-base.yaml
```diff
@@ -0,0 +1,59 @@
...
+
+  hooks-system:
+    essential:
+      - dhcpcd-base_hooks-core
+    contents:
+      /usr/share/dhcpcd/hooks/15-timezone:
+      /usr/lib/dhcpcd/dhcpcd-hooks/50-timesyncd.conf:
+
+  libs:
```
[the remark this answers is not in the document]
Meulengracht: this is more in line with libs imho, since it's a library under /usr/lib, not a udev rule or something else.

alesancor1:
Marking as blocked - Pending other PRs that were split out of this one.

Additionally, we should also split the python modules out of this PR so we can address each one separately 👍

alesancor1:
> fdisk, logsave and part of systemd have been merged. does this want a rebase / merge of canonical/ubuntu-26.04 ?

There are stil some bits that should be split out of this PR, python libs mostly

PR #815, opened 2025-12-16
fix(24.04): add go-env

zhijie-yang:
# Proposed changes

This fixes the missing default go env when installing the go slices.

### Forward porting

See status in #797. FP PRs will have this included.

## Checklist


lczyk:
for posterity: this fixes a bug in #797 which causes `golang:all` test to fail with:

```
internal/deb/extract.go:16:2: GOPROXY list is not the empty string, but contains no entries
internal/deb/extract.go:17:2: GOPROXY list is not the empty string, but contains no entries
...
```

PR #830, opened 2026-01-12
feat(25.10): cargo-1.85 slices

lczyk:
# Proposed changes

slices for `cargo-1.85`

this is the default cargo for 25.10

## Related issues/PRs

this PR is part of the following stack:

- ~~https://github.com/canonical/chisel-releases/pull/786~~
  - https://github.com/canonical/chisel-releases/pull/830 **(this PR)**

### Forward porting

the default rust versions for each of the current versions of Ubuntu are:

ubuntu | rustc | gcc
-- | -- | --
24.04 | 1.75 | 13
25.04 | 1.84 | 14
25.10 | 1.85 | 15
26.04 | 1.88 | 15

**1.75**:
  - https://github.com/canonical/chisel-releases/pull/814 

**1.84**:
  - ~~https://github.com/canonical/chisel-releases/pull/634~~ _(EOL)_ 

**1.85**:
  - https://github.com/canonical/chisel-releases/pull/845 
  - https://github.com/canonical/chisel-releases/pull/830 **(this PR)**
  - https://github.com/canonical/chisel-releases/pull/831

**1.88**:
  - https://github.com/canonical/chisel-releases/pull/833
  - https://github.com/canonical/chisel-releases/pull/834

## Checklist

🦀

### DIFF
--- a/slices/fonts-ipafont-gothic.yaml
+++ b/slices/fonts-ipafont-gothic.yaml
@@ -1,0 +1,17 @@
+package: fonts-ipafont-gothic
+
+essential:
+  - fonts-ipafont-gothic_copyright
+
+slices:
+  fonts:
+    contents:
+      /usr/share/fonts/opentype/ipafont-gothic/*.ttf:
+
+  config:
+    contents:
+      /etc/fonts/conf.avail/65-fonts-ipafont-gothic.conf:
+
+  copyright:
+    contents:
+      /usr/share/doc/fonts-ipafont-gothic/copyright:
--- a/slices/fonts-noto-color-emoji.yaml
+++ b/slices/fonts-noto-color-emoji.yaml
@@ -1,0 +1,13 @@
+package: fonts-noto-color-emoji
+
+essential:
+  - fonts-noto-color-emoji_copyright
+
+slices:
+  fonts:
+    contents:
+      /usr/share/fonts/truetype/noto/NotoColorEmoji.ttf:
+
+  copyright:
+    contents:
+      /usr/share/doc/fonts-noto-color-emoji/copyright:
--- a/slices/fonts-tlwg-loma-otf.yaml
+++ b/slices/fonts-tlwg-loma-otf.yaml
@@ -1,0 +1,13 @@
+package: fonts-tlwg-loma-otf
+
+essential:
+  - fonts-tlwg-loma-otf_copyright
+
+slices:
+  fonts:
+    contents:
+      /usr/share/fonts/opentype/tlwg/Loma-*.otf:
+
+  copyright:
+    contents:
+      /usr/share/doc/fonts-tlwg-loma-otf/copyright:
--- a/slices/fonts-unifont.yaml
+++ b/slices/fonts-unifont.yaml
@@ -1,0 +1,13 @@
+package: fonts-unifont
+
+essential:
+  - fonts-unifont_copyright
+
+slices:
+  fonts:
+    contents:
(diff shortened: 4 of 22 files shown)


slices/libpam0g-dev.yaml
```diff
@@ -0,0 +1,23 @@
...
+  libs:
+    essential:
+      - libpam0g_libs
+    contents:
+      /usr/lib/*-linux-*/libpam.a:
+      /usr/lib/*-linux-*/libpam.so:  # Symlink to libpam.so.0
+      /usr/lib/*-linux-*/libpam_misc.so:  # Symlink to libpam_misc.so.0
+      /usr/lib/*-linux-*/libpamc.so:  # Symlink to libpamc.so.0
+
+  headers:
```
cjdcordeiro, line 16: doesn't seem to be used or tested anywhere. So do we want to include them? The slice seems accurate, since these are the only headers, but still...

lczyk: no, good point. i do thing that headers are part of the core functionality of a `-dev` package so they should be included, but also then we should test them. hence see https://github.com/canonical/chisel-releases/pull/956 (and friends)

tests/spread/integration/cargo/task.yaml
alesancor1, line 1: `cargo-1.85` is the default for `25.04`, so we can take a similar approach to `python` here and avoid duplicating all the tests:

https://github.com/canonical/chisel-releases/blob/0cc06452905c9c13067af6ff042a95ce8e9c6f04/tests/spread/integration/python3/task.yaml#L8-L14

lczyk: i dont like the cross-test cp :grimacing: reason being is that if you end up editing the test for `python3.12` you don't know that you're also editing the test for `python3`. i'd like to move to some much more explicit way of sharing test files, which we did discuss a bit in the testing meeting with @lengau at some point, but have not really progressed on that since. if you insist, i will change the tests here to the python style, but i think its an anti-pattern, which is why i have not done it this way here. (yes, copying the same file in a few places is bad too, but its a lesser evil imo)

alesancor1: I agree on it not being an ideal and we should probably find a way to share test files. Though, I think for now we should stay consistent with the same approach and avoid having different testing strategies for the same use-case.

 --- 

On a different note, I'm happy to resume the testing discussion, let's keep the scope small and focus on addressing this situation, we can then build on top of that.

lczyk: > I think for now we should stay consistent with the same approach and avoid having different testing strategies for the same use-case.

aka? this approach of the python one? note that the approach here is already what we do in e.g. `cpp`, `gcc` and `binutils`.

> On a different note, I'm happy to resume the testing discussion, let's keep the scope small and focus on addressing this situation, we can then build on top of that.

yep, very much :+1: and yeah, lets try to focus on one thing, and i don't mind this being the one

alesancor1: > aka? this approach of the python one? note that the approach here is already what we do in e.g. cpp, gcc and binutils.

Being that the case I'll not block this any longer. Do we have an issue to track this already? If not, let's create one and keep track of the tests that will require to be modified when we find a way to share the files for them :+1:

lczyk: yep. tracked here:

- https://github.com/canonical/chisel-releases/issues/838

PR #832, opened 2026-01-13
feat(26.04): add frr slices

zhijie-yang:
# Proposed changes

This PR adds the slice definition of FRR on 26.04

### Forward porting

- https://github.com/canonical/chisel-releases/pull/1120
- https://github.com/canonical/chisel-releases/pull/832 **(this PR)**

## Checklist


tests/spread/integration/frr/configs/staticd.conf
lczyk, line 1: i'd name the folder `testfiles` for consistency with other tests. no strong feelings and we don't have a strong convention anyway

slices/frr-pythontools.yaml
```diff
@@ -0,0 +1,21 @@
...
+
+slices:
+  scripts:
+    essential:
+      - frr_bins
+      - libpython3.13-stdlib_data-types
+      - python3-minimal_python3
+      - python3.13_core
+    contents:
+      /usr/lib/frr/frr-reload.py:
```
lczyk, line 14: has `argparser` so can be tested. at least `--help` to check all the python imports fire.

slices/frr-pythontools.yaml
```diff
@@ -0,0 +1,21 @@
...
+      /usr/lib/frr/frr_babeltrace.py:
```
lczyk, line 15: at present non-functioning with import error on `babeltrace`. worth splitting it into a separate slice and adding `python3-babeltrace_libs` ??

NOTE: installing `frr-pythontools` from debs *also* installs this script in a non-functioning state, so idk whether we want to deviate from that or not. maybe split it into a separate slice, add a comment that this will not work without `babeltrace` and then only in tests install `python3-babeltrace` to test the script?

slices/frr-pythontools.yaml
lczyk: same for `generate_support_bundle.py` and `ospfclient.py`

slices/iproute2.yaml
lczyk, line 1: we should backport this fix too. i've checked only 25.10 but it should be there too

slices/frr.yaml
lczyk, line 1: note: i found that installing frr in 26.04 container failed due to `chown syslog:adm /var/log/frr`. i had to add the user `useradd --uid 104 syslog && usermod --append --groups adm syslog` and reconfigure `dpkg --configure frr`.

this obv won't happen in this PR, but i wanted to take a note and then chase with upstream later

lczyk: oo.. actually, this package includes the `var/log/frr` dir. i think we should have some more tests in which we enable verbose mode and file logging to check 1) whether all of that works fine and 2) whether we need to add the `var/log/frr` from the deb. 

frr demons are not very verbose by design, so they may not have logged anything during the tests, and not find that particular case / failure mode.

slices/frr.yaml
```diff
@@ -0,0 +1,119 @@
...
+  # Note: FRR expects the 'frr' user and 'frrvty' group to exist.
+  # You may need to manually add them in the chiseled rootfs using the following commands.
+  # FRR_UID, FRR_VTY_GID, and FRR_GID can be adjusted as needed.
+  #    FRR_UID=999
+  #    FRR_VTY_GID=998
+  #    FRR_GID=999
+  #    echo -e "frr:x:$FRR_UID:\nfrrvty:x:$FRR_VTY_GID:frr" >> "${rootfs}/etc/group"
+  #    echo -e "frr:x:$FRR_UID:$FRR_GID:Frr routing suite,,,:/nonexistent:/usr/sbin/nologin" \
+  #        >> "${rootfs}/etc/passwd"
+  #    mkdir -p "${rootfs}/var/lib/frr"
```
lczyk, line 53: why here if its in the deb? see this comment: https://github.com/canonical/chisel-releases/pull/832/changes#r2782347602

slices/frr.yaml
```diff
@@ -0,0 +1,119 @@
...
+      - frr_daemons
+      - grep_bins
+    contents:
+      /usr/lib/frr/frrcommon.sh:  # Requires /bin/bash in shebang
+      /usr/lib/frr/frrinit.sh:  # Requires /bin/bash in shebang
+      /usr/lib/frr/watchfrr.sh:  # Requires /bin/bash in shebang
+
+  frr-reload:
+    essential:
+      - frr-pythontools_scripts
```
lczyk:
```suggestion
      - base-files_bin
      - dash_bins
      - frr-pythontools_scripts
```

i don't think it has dash otherwise

also, adding `base-files` since shebang is `#!/bin/sh`

slices/frr.yaml
```diff
@@ -0,0 +1,119 @@
...
+      /usr/lib/*-linux-*/frr/modules/*.so:
+      /usr/share/yang/frr-*.yang:
+      /usr/share/yang/ietf-bgp-types.yang:
+      /usr/share/yang/ietf-interfaces.yang:
+      /usr/share/yang/ietf-key-chain.yang:
+      /usr/share/yang/ietf-netconf-acm.yang:
+      /usr/share/yang/ietf-netconf-with-defaults.yang:
+      /usr/share/yang/ietf-netconf.yang:
+      /usr/share/yang/ietf-routing-types.yang:
+      /usr/share/yang/ietf-srv6-types.yang:
```
lczyk, line 107: in libyang you kept the modules in a separate slice. i'd do that here too conf consistency. ( and then make them an essential of libs ofc )

slices/frr.yaml
```diff
@@ -0,0 +1,119 @@
...
+      /usr/lib/frr/ripd:
+      /usr/lib/frr/ripngd:
+      /usr/lib/frr/staticd:
+      /usr/lib/frr/vrrpd:
+      /usr/lib/frr/watchfrr:
+      /usr/lib/frr/zebra:
+
+  libs:
+    essential:
+      - base-files_var
```
lczyk: why?

PR #878, opened 2026-02-02
feat(26.04): improve systemd slices

Meulengracht:
# Proposed changes

Improvements to systemd slices, including more segregation of functionality. These were made as a part of making Ubuntu Core 26 work.

## Checklist


slices/systemd.yaml
```diff
@@ -4,7 +4,13 @@ essential:
   - systemd_copyright
 
 slices:
+  # now a placeholder, just redirects to core
   bins:
+    essential:
+      - systemd_core
```
alesancor1: Considering that `bins` was including all the binaries previously, it should point to `standard` rather than `core`

```suggestion
      - systemd_standard
```

slices/systemd.yaml
```diff
@@ -4,7 +4,13 @@ essential:
   - systemd_copyright
 
 slices:
+  # now a placeholder, just redirects to core
   bins:
```
alesancor1: Well since we are moving to a new base, regressions are acceptable to some extent. We can consider deprecating `bins` in favour of `standard`  here - lmk what you think @canonical/rocks

slices/systemd.yaml
```diff
@@ -14,93 +20,238 @@ slices:
...
+      - systemd_core
+      - systemd_dbus-services
+      - systemd_hostname
+      - systemd_journal
+      - systemd_locale
+      - systemd_login
+      - systemd_network
+      - systemd_services
+      - systemd_timedate
+      - systemd_utils
```
alesancor1: These should also include the ones moved outside core:

* `systemd_detect-virt`
* `systemd_kernel-parameters`
* `systemd_pcrlock`
* `systemd_user-services`

slices/systemd.yaml
```diff
@@ -14,93 +20,238 @@ slices:
...
+
+  # Journal management
+  journal:
+    essential:
+      - libc6_libs
+      - libsystemd-shared_libs
+      - libsystemd0_libs
+      - systemd_config
       - systemd_system-services
-      - systemd_user-services
```
alesancor1: Same here, it has been removed and not present neither in `core` or `standard`

tests/spread/integration/systemd/task.yaml
```diff
@@ -3,7 +3,7 @@ summary: Integration tests for systemd
 execute: |
   # Chisel a minimum number of slices to give us a runnable system that we can
   # test in.
-  rootfs="$(install-slices bash_bins coreutils_bins passwd_config base-files_base systemd_bins)"
```
alesancor1: If we are keeping `_bins` there's no need to modify this, as long as `_bins` points to `_standard`

alesancor1 (changes requested):
It's a big PR but looks pretty good already, thanks for the effort! Some comments below:

slices/systemd.yaml
```diff
@@ -513,14 +607,6 @@ slices:
...
       /usr/share/systemd/language-fallback-map:
 
-  # Generated services slice is to emulate symlinks or initial state
-  # config generated by systemd upon install.
-  # Systemd performs a lot of initialization upon install, however since systemd
-  # does actually not rely on these install actions (since all of these are run as
-  # a part of the first install / early setup upon first boot), a lot of this is
-  # left out. It's done like that because systemd needs to be usable after install
-  # on a running system, but this is not the case here.
-  # These are the few that seems to be not emulated
```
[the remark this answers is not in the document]
Meulengracht: This got lost in the update :-) added back!

slices/systemd.yaml
lczyk: yep. can remove `bins` from here before 26.04 gets out of devel

lczyk: https://github.com/canonical/chisel-releases/pull/878/commits/8736134df22b958104d6d9b2e46df24e2e9462e5

tests/spread/integration/systemd/task.yaml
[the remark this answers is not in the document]
lczyk: https://github.com/canonical/chisel-releases/pull/878/commits/0335439d48b4d05c31ee1b936dd98ea919845286

slices/systemd.yaml
```diff
@@ -4,7 +4,8 @@ essential:
   systemd_copyright:
 
 slices:
-  bins:
+  # Core systemd daemon
```
zhijie-yang: Do we have a reference supporting this split? If so, does a ' minimal ' slice exist, which provides even more fundamental functionalities than `core`?

alesancor1 (approved):
LGTM, thanks!

Please @Meulengracht verify that the latest changes do not break your use-case(s). Afaict, the major breaking change is the removal of `bins` in favor of `core`/`standard`

Meulengracht (commented):
+1, lgtm

slices/systemd.yaml
lczyk: hmmm... so, i want to avoid too too much back and forth because we do have a strict deadline to get this in within next two weeks, and this PR is already a massive improvement. having said that, here are my 2c on this:

yes, there can be a more minimal `minimal` slice. for that i would evict `systemd-analyze`, `kernel-install` and `systemd-mute-console` from the current core. `systemd-analyze` could be in standard, and so would `kernel-install`. then `core` -- the one in between could have `minimal` + `systemd-journal`, `systemd-tempfiles` and `systemd-sysusers`.

@Meulengracht ^ ? thoughts on that. its primarily your PR and you are much more of an expert on systemd than me

lczyk: but, if im correct, that's not actually what you're asking about @zhijie-yang (?) the question is: are there sources ( e.g. systemd docs / manpages ) which support this split into `core` / `standard` etc (see [this](https://documentation.ubuntu.com/chisel/latest/explanation/slice-design-approaches/#slice-design-approaches) link), or is this an ad-hock / experience based. we'd ideally prefer the former, but it's alas not always possible

lczyk: after speaking with @zhijie-yang, we've decided that since we're not quite sure about the structure of `minimal`/`core`, they're better to be left out for now. hence i've inlined `core` into `standard` and renamed `utils` slice into `dev`, according to the [slice design approaches](https://documentation.ubuntu.com/chisel/latest/explanation/slice-design-approaches/#slice-design-approaches) doc. i've updated `test_core.sh` to be `test_standard.sh`. the test can be expanded more in the future.

@Meulengracht confirmed that only `standard` slice is required for their work. @alesancor1 i've re-requested review due to non-trivial changes.

alesancor1 (approved):
Happy with the latest changes.

Please open an issue regarding further testing if it will not be addressed now.

lczyk (approved):
approve from me, but i'm one of two authors so let's get one more +1

lczyk:
[a3d556b](https://github.com/canonical/chisel-releases/pull/878/commits/a3d556b3f91e5816571a7d029cede8a6bfa712d9) patched `init-system-helpers` following merge of https://github.com/canonical/chisel-releases/pull/935

slices/systemd.yaml
```diff
@@ -14,114 +14,276 @@ slices:
...
+      /usr/lib/systemd/network/80-container-ve.link:
+      /usr/lib/systemd/network/80-container-ve.network:
+      /usr/lib/systemd/network/80-container-vz.link:
+      /usr/lib/systemd/network/80-container-vz.network:
+      /usr/lib/systemd/network/80-vm-vt.link:
+      /usr/lib/systemd/network/80-vm-vt.network:
+      /usr/lib/systemd/network/80-wifi-adhoc.network:
+
+  # System services and daemons
+  services:
```
zhijie-yang: we need to put a hint here to indicate this slice contains the service (backend|executor|provider)s of the systemd early-boot services.

maybe we should even name it as `service-(handlers|daemons)`, as we already have `(user|system)-services` slices, where the `.service` and alike files are included.

zhijie-yang (approved):

lczyk:
phew! finally there was a break in archive updated and CI managed to succeed. with that, it's ready to merge :partying_face:

PR #923, opened 2026-02-21
feat(24.04): add test for chrome-headless-shell

mus65:
# Proposed changes

As discussed in https://github.com/canonical/chisel-releases/pull/905#discussion_r2829036150 this adds the test that was originally part of #905 with some changes:

- added sha256 checksum validation of the chrome binary
- added umount of /proc and /dev
- moved script to its own file
- removed prepare/execute/restore
- moved test from integration/ to extra/ since chrome-headless-shell is not actually an ubuntu package

Since only the tests in integration/ are executed, the test is not currently running in CI. It's to be discussed whether extra/ is the correct approach and what would need to be changed to run these in CI as well. From my understanding this would require changes to spread.yaml on the main branch? So it would have to be changed in another PR.

## Related issues/PRs

will open forward port PRs when this is out of draft.

## Checklist

### DIFF
--- a/tests/spread/extra/chrome-headless-shell/task.yaml
+++ b/tests/spread/extra/chrome-headless-shell/task.yaml
@@ -1,0 +1,2 @@
+summary: Integration tests for chrome-headless-shell
+execute: bash -ex ./test.sh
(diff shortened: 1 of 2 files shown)


cjdcordeiro (approved):

mus65:
Note that this test would have caught #1002 . As mentioned in the description, this PR is still in draft since it's not clear to me what changes would have to be made to actually run this in CI.

PR #954, opened 2026-03-11
feat(26.04): add memcached slices and integration tests

asanvaq:
# Proposed changes
This PR introduces Chisel slice definitions for `memcached` targeting Ubuntu 26.04. It includes the necessary binary and configuration slices, along with a functional integration test.

- Created `memcached_bins`, `memcached_config` and `memcached_copyright`  slices.

I have verified the slices using the included Spread integration test. 
The test confirms:
1. The binary executes and reports the correct version.
2. The daemon starts and listens on port `11211`.
3. Successful `SET` and `GET` operations via netcat.

**Spread Output:**
`Successful tasks: 1`

## Related issues/PRs
N/A

### Forward porting
N/A

## Checklist

## Additional Context
N/A


slices/memcached.yaml
```diff
@@ -0,0 +1,25 @@
...
+      libsasl2-2_libs:
+      libssl3t64_libs:
+      memcached_config:
+    contents:
+      /usr/bin/memcached:
+
+  config:
+    contents:
+      /etc/default/memcached:
+      /usr/lib/tmpfiles.d/memcached.conf:
```
alesancor1: This is a `systemd` configuration file. We don't have `systemd` in this slices so should not be required here.

If we at some point have a use case that requires systemd then we can add it back


```suggestion
```

tests/spread/integration/memcached/task.yaml
```diff
@@ -0,0 +1,40 @@
...
+    apt-get update && apt-get install -y iproute2 netcat-openbsd
+
+execute: |
+    rootfs="$(install-slices memcached_bins)"
+
+    chroot "${rootfs}/" memcached -V
+
+    # Mock the 'root' user so the binary can verify its -u flag inside the chroot
+    mkdir -p "${rootfs}/etc"
+    echo "root:x:0:0:root:/root:/bin/sh" > "${rootfs}/etc/passwd"
```
alesancor1: This is fine, but you can probably just install the `base-passwd_data` slice instead.

tests/spread/integration/memcached/task.yaml
```diff
@@ -0,0 +1,40 @@
...
+
+    # Verify it can start and listen on the default port (11211)
+    chroot "${rootfs}/" memcached -u root -d
+    
+    # Wait for the socket to settle
+    sleep 2
```
alesancor1: From my experience, this is not really reliable... Ideally you'd want to add some polling logic with a timeout instead of just sleeping.

tests/spread/integration/memcached/task.yaml
```diff
@@ -0,0 +1,40 @@
...
+    
+    # Check if the port is open using the host's networking tools
+    # pointing at the chroot's process
+    ss -tulpn | grep :11211
```
alesancor1: In line with the comment above: You can put this check in a `while` loop and `break` when the port is open, or fail if certain number of attempts is exceeded.

alesancor1 (changes requested):
Looks good overall. Just have a couple of comments (see below).

I've also noticed there are some scripts under `usr/share/memcached/scripts`, afaiu these are not needed for this use-case (i.e creating a rock) so there's no need to slice them.

tests/spread/integration/memcached/task.yaml
```diff
@@ -0,0 +1,40 @@
+summary: Integration tests for memcached
+
+prepare: |
+    # Ensure the host has the networking tools required for the test
+    apt-get update && apt-get install -y iproute2 netcat-openbsd
+
+execute: |
+    rootfs="$(install-slices memcached_bins)"
+
+    chroot "${rootfs}/" memcached -V
```
lczyk:
```suggestion
    chroot "$rootfs" memcached -V
```

tests/spread/integration/memcached/task.yaml
```diff
@@ -0,0 +1,40 @@
...
+
+    # Mock the 'root' user so the binary can verify its -u flag inside the chroot
+    mkdir -p "${rootfs}/etc"
```
lczyk:
```suggestion
    mkdir -p "$rootfs/etc"
```

tests/spread/integration/memcached/task.yaml
lczyk:
```suggestion
    echo "root:x:0:0:root:/root:/bin/sh" > "$rootfs/etc/passwd"
```

tests/spread/integration/memcached/task.yaml
```diff
@@ -0,0 +1,40 @@
...
+    echo "root:x:0:0:root:/root:/bin/sh" > "${rootfs}/etc/passwd"
+
+    # Verify it can start and listen on the default port (11211)
+    chroot "${rootfs}/" memcached -u root -d
```
lczyk:
```suggestion
    chroot "$rootfs" memcached -u root -d
```

tests/spread/integration/memcached/task.yaml
```diff
@@ -0,0 +1,40 @@
...
+    # Protocol: get <key>\r\n
+    echo "Reading from memcached..."
+    
+    RESPONSE=$(printf "get mykey\r\n" | nc -q 1 127.0.0.1 11211)
+
+    echo "Response: $RESPONSE"
+    echo "$RESPONSE" | grep -q "hello"
+
+    # Cleanup
+    pkill memcached || true
```
lczyk:
```suggestion
    pkill memcached || true

```

nit: _(adding a newline at the end of the file)_

lczyk (commented):
it looks really good! some nits from me, and there are some very valid comments from @alesancor1, so i will only 'comment' as opposed to 'approve' for now

edit: and yeah, i'd personally lay out the tests like [this](https://github.com/canonical/chisel-releases/tree/ubuntu-26.04/tests/spread/integration/bash) (for example). task.yaml for you would then look like this:

```yaml
summary: Integration tests for memcached

variants:
    - memcached

execute: bash -ex ./test_${SPREAD_VARIANT}.sh
```

and then the tests would live in `test_memcached.sh`. it's not really that important here beyond nice syntax highlighting, but it makes it much saner when there are more tests.

i won't let 'perfect' get in the way of 'done' though, and it's not a blocker on my side. just a personal preference / recommendation.

slices/memcached.yaml
alesancor1, line 1: NOTE:

The maintainer scripts copy `/usr/share/memcached/memcached.conf.default` to `/etc/memcached.conf`:

```bash
if [ ! -e /etc/memcached.conf ]
        then
            mkdir -p /etc
            cp /usr/share/memcached/memcached.conf.default /etc/memcached.conf
    fi
```

Though, this config file is only used by the `usr/share/memcached/scripts/start-memcached` helper script, which is neither needed for memcached to work, nor used by our current `memcached` docker image.

tests/spread/integration/memcached/test_memcached.sh
```diff
@@ -0,0 +1,50 @@
...
+    fi
+    attempts=$((attempts + 1))
+    echo "Attempt $attempts/$max_attempts: waiting for port 11211..."
+    sleep 1
+done
+
+if [ $attempts -eq $max_attempts ]; then
+    echo "Failed to detect memcached listening on port 11211 after $max_attempts attempts"
+    exit 1
+fi
```
alesancor1, line 31: You can simplify this logic to something similar to what we do for `apache2` and `php`:

```bash
until ss -tulpn | grep -q :11211; do
    sleep 1
done
```

Or you still want the attempts logic:

```bash
for i in {1..10}; do
    ss -tulpn | grep -q :11211 && break
    echo "Attempt $i/10: waiting for port 11211..."
    sleep 1
done

if ! ss -tulpn | grep -q :11211; then
    echo "Failed to detect memcached on port 11211"
    exit 1
fi
```

Still, there's nothing wrong with your current approach, so all good :+1:

alesancor1 (approved):
LGTM, thanks!

Just couple non-blocking notes/comments below:

alesancor1 (approved):

lczyk:
merged libsasl and merged `ubuntu-26.04` here. you might want to do `git pull --rebase` on your side to get the latest changes

lczyk (approved):
+1 from me

i second @alesancor1's comments but i also second the 'nothing wrong with the current approach' part ;)

will mark as ready and give is a grace period for any other comments before merging 👍

slices/memcached.yaml
lczyk: mmm.. can be added much more easily than removed. let's slice what we need for the image and then improve later if needed

Issue #838, opened 2026-01-14
feat(testing): sharing files between tests


lczyk:
there are many cases when we want to share a bunch of files between tests. at present the files are copied, but instead we should have one common place where we can have such files and then symlink them.

considerations:

- this woudl affect how the changes in test files are detected in https://github.com/canonical/chisel-releases/issues/837
- this issue is mainly pertinent to ~files in `tests/spread/integration/foo/testfiles/...`~~ large/lots of files.
- further thought has to be given to sharing the tests themselves, and care must be taken such that the tests remain simple and readable

PR #963, opened 2026-03-13
feat(26.04): add slices for bind9

alesancor1:
# Proposed changes
Slices for `bind9`

## Checklist

### DIFF
--- a/slices/bind9.yaml
+++ b/slices/bind9.yaml
@@ -9,7 +9,7 @@ slices:
       bind9_config:
       bind9_libs:
     contents:
-      /usr/bin/arpname:
+      /usr/bin/arpaname:
       /usr/bin/dnssec-importkey:
       /usr/bin/named-journalprint:
       /usr/bin/named-nzd2nzf:
--- a/slices/bind9.yaml
+++ b/slices/bind9.yaml
@@ -34,12 +34,12 @@ slices:
       libcap2_libs:
       libjemalloc2_libs:
       libjson-c5_libs:
-      libmldb0_libs:
+      liblmdb0_libs:
       libmaxminddb0_libs:
       libnghttp2-14_libs:
-      libssl3_libs:
+      libssl3t64_libs:
       liburcu8t64_libs:
-      libuv1_libs:
+      libuv1t64_libs:
       libxml2-16_libs:
       zlib1g_libs:
     contents:
--- a/slices/bind9.yaml
+++ b/slices/bind9.yaml
@@ -29,6 +29,8 @@ slices:
       /etc/default/named:
 
   data:
+    essentials:
+      dns-root-data_data:
     contents:
       /var/cache/bind/: {make: true, mode: 02775}
       /var/lib/bind/: {make: true, mode: 02775}
--- a/slices/bind9.yaml
+++ b/slices/bind9.yaml
@@ -29,7 +29,7 @@ slices:
       /etc/default/named:
 
   data:
-    essentials:
+    essential:
       dns-root-data_data:
     contents:
       /var/cache/bind/: {make: true, mode: 02775}
--- a/slices/bind9.yaml
+++ b/slices/bind9.yaml
@@ -38,7 +38,7 @@ slices:
       libmaxminddb0_libs:
       libnghttp2-14_libs:
       libssl3_libs:
-      liburcut8t64_libs:
+      liburcu8t64_libs:
       libuv1_libs:
       libxml2-16_libs:
       zlib1g_libs:
--- a/slices/bind9.yaml
+++ b/slices/bind9.yaml
@@ -30,8 +30,8 @@ slices:
 
   data:
     contents:
-      /var/cache/bind: {make: true, mode: 02775}
-      /var/lib/bind: {make: true, mode: 02775}
+      /var/cache/bind/: {make: true, mode: 02775}
+      /var/lib/bind/: {make: true, mode: 02775}
 
   libs:
     essential:
--- a/slices/bind9.yaml
+++ b/slices/bind9.yaml
@@ -48,4 +48,4 @@ slices:
 
   copyright:
     contents:
-      /usr/share/doc/bind9-libs/copyright:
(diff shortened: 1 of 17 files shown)


slices/bind9.yaml
```diff
@@ -0,0 +1,59 @@
+package: bind9
+
+essential:
+  bind9_copyright:
+
+slices:
+  bins:
```
alesancor1: I'm considering renaming this to `core` and add a `standard` slice that includes `bind9-utils` since they come with the standard installation of `bind9` and will be part of the `bind9` rock.

Thoughts? @canonical/slice-reviewers-guild

alesancor1: I added `bind9-utils` slices in https://github.com/canonical/chisel-releases/pull/963/commits/d83782de1f7385f8534beb2fa547d3cbd4ff6f4a

bind9-utils` bring the following:
* `rndc` command for managing dns server.
* `dnssec-*` for enabling dnssec
* some checks and validations for zone config (i.e `checkzone`, `checkconf`, `compilezone`)

None of them are a requirement for `bind9` to run as intended. A `standard` slice would essentially be an alias of `chisel cut bind9_bins bind9-utils_bins`, I don't have a strong opinion here, whether to have `core`/`standard` instead of `bins`.

cjdcordeiro: are they in the current bind9 ubuntu image? are they also present in other upstream bind9 images? 

If yes, then we should include it too. `bind9` has a dependency on `bind9-utils`, and this is not currently the case in this SDF. So maybe the `core` and `standard` slices make sense, as a way to introduce `bind9-utils` as you propose.

are all `bins` of bind9-utils needed?

slices/bind9.yaml
```diff
@@ -0,0 +1,59 @@
...
+      /usr/bin/dnssec-importkey:
+      /usr/bin/named-journalprint:
+      /usr/bin/named-nzd2nzf:
+      /usr/bin/named-rrchecker:
+      /usr/bin/nsec3hash:
+      /usr/sbin/ddns-confgen:
+      /usr/sbin/named:
+      /usr/sbin/tsig-keygen:
+
+  config:
```
cjdcordeiro, line 33: what about all the other files in this dir?

cjdcordeiro (changes requested):

slices/bind9.yaml
alesancor1: As discussed in our 1:1, I've split `bind9` into `core` and `standard`, where:

* `core` contains the essential `named` binary + the `bind9-utils` dependencies to manage it (`rndc`) and  verify its config (`named-*`)
* `standard` contains the `dnssec` features: config and key generators, plus `iproute2` and `netbase`, as specified in the `.deb` dependencies.

`bind9-utils` has been split accordingly as well. See the commit: https://github.com/canonical/chisel-releases/pull/963/commits/efb2a4f2222b9c7dc21eb2f02185a8a917484c63

slices/bind9.yaml
alesancor1: The other config files are:

* `/etc/apparmor.d/` - `apparmor` specific config. `apparmor` is not installed
* `/etc/insserv.conf.d/` - `insserv` specific configs. `insserv` is not installed
* `/etc/network/if-down.d/` and `/etc/network/if-up.d/` - hooks for `netscript-2.4` binaries. `netscript-2.4` is not installed
* `/etc/ppp` - hooks for `ppp` binaries. `ppp` is not installed
* `/etc/ufw` - configs for `ufw`. `ufw` is not installed

cjdcordeiro (approved):

slices/bind9.yaml
```diff
@@ -0,0 +1,69 @@
+package: bind9
+
+essential:
+  bind9_copyright:
+
+slices:
+  core:
```
zhijie-yang, line 7: This seems more like a `named` slice, instead of a `core` slice... I'd suggest adding a hint like: provides named, the basic component.

slices/bind9.yaml
```diff
@@ -0,0 +1,69 @@
...
+
+  config:
+    contents:
+      /etc/bind/named.conf:
+      /etc/bind/named.conf.local:
+      /etc/bind/named.conf.options:
+      /etc/bind/named.conf.root-hints:
+      /etc/default/named:
+
+  data:
```
zhijie-yang, line 41: Should this be named as `var`, similar to what we have in `libpam-runtime`, `libnss-extrausers`, `rabbitmq`, `dpkg`, and `apt`?

tests/spread/integration/bind9/db.test.local
zhijie-yang, line 1: You can copy this test file from `bind9-utils`' in the task.yaml during the `prepare` to avoid duplication.

slices/bind9.yaml
alesancor1: It also provides `rndc` to manage the dns server, appart from the check utilities. A slice containing **just** `named` would rather be `minimal`

slices/bind9.yaml
alesancor1: This is a bit different, since the `data` slice here also brings `dns-root-data_data`. I think `var` could be misleading because the slice is not only creating the `var/*/bind` directories but also adding data files.

tests/spread/integration/bind9/db.test.local
alesancor1: I share a similar opinion to @lczyk regarding cross-test copies. See the discussion [here](https://github.com/canonical/chisel-releases/pull/830#discussion_r2924908240)

We have an issue to track this situation: https://github.com/canonical/chisel-releases/issues/838

slices/bind9.yaml
zhijie-yang: I'd say the `data` is not from `bind9` itself, but from `dns-root-data`. Feel free to keep the name as `data` if you find the data from `dns-root-data` is tightly coupled with `bind9`.

zhijie-yang (approved):

PR #968, opened 2026-03-30
fix(24.04): base-files as essential of apparmor package

upils:
# Proposed changes

`apparmor_extras` does not declare `base-files_lib` as an essential and install content under `/lib`. Given the way Chisel resolves dependencies and orders packages extraction, this lead to `apparmor` being extract before `base-files`, creating a `/lib` directory, preventing from successfully creating a symlink to `/usr/lib/` when `base-files` is extracted. A similar issue exists with `/usr/sbin`.

This results in `apparmor_extras` being unusable.

This is currently not detected by `chisel debug check-release-archives`. Fixes in Chisel are under considerations.

Several other packages are in a similar situation, except that with their name they are sorted **after** `base-files`, so the issue is not blocking their installation. 

This fix enforce `base-files_lib` and `base-files_bin` as essentials at the **package level**, forcing Chisel to extract `base-files` before `apparmor`.

### Forward porting

Later releases are not affected.

## Checklist

### DIFF
--- a/slices/apparmor.yaml
+++ b/slices/apparmor.yaml
@@ -2,6 +2,10 @@ package: apparmor
 
 essential:
   - apparmor_copyright
+  # Make sure base-files is extracted before any other slices of apparmor to
+  # to avoid creating /lib and /sbin directories instead of symlinks.
+  - base-files_lib
+  - base-files_bin
 
 slices:
   # Also relies on debconf, however it is solely used by install hook


upils:
It looks like the CI failure is unrelated to this change but still related to armor. I suspect the kernel of the testing environment is not compatible with the version of armor tested.

lczyk:
yeah, CI failures are unrealted. they are due to missing profiles. i've made a fix here: https://github.com/canonical/chisel-releases/pull/987. i have also, in fact, run into the issue this PR is fixing *many* times when doing the fix above, and i confirm this PR fixes it.

btw, this fix is using multiple top-level essentials. i thought that was not supported based on this discussion?? :raised_eyebrow: https://github.com/canonical/chisel/issues/271 is is because they're foreign essentials?

lczyk (approved):
approving since this works, but let's wait for the apparmor fix to come in + let's do forward ports of this fix too

lczyk:
removing the 'forward port missing' label since, indeed

> Later releases are not affected.

i should learn to read more carefully :sweat_smile:

upils:
> btw, this fix is using multiple top-level essentials. i thought that was not supported based on this discussion?? 🤨 https://github.com/canonical/chisel/issues/271 is is because they're foreign essentials?

Multiple top-level essentials is working, it just appears that is was never (?) used to far (I did not do an exhaustive check). As mentioned by @letFunny in the other issue, the thing we do not support dealing with **cycles**, which is not the case here. This is indeed due to that fact that the essentials are from another package, but this is not how we detect it, so it is more a consequence of the rule.

lengau (approved):

lczyk:
since https://github.com/canonical/chisel-releases/pull/987 is merged, and the tests are passing, merging this with 2 approvals :+1:

PR #970, opened 2026-04-01
feat(26.04): update `cargo`/`rust`

lczyk:
# Proposed changes

update `cargo` and `rust` to 1.93 which is now the new default version on 26.04.

the update is mostly a find-and-replace of version numbers, but there are a few things to point out:

- the shas in `libstd-rust-1.93` and `libstd-rust-1.93-dev` needed to be updated
- `eza` test was failing due to missing `zlib1g-dev` which this PR therefore also slices. tests in previous releases did not depend on this package.
- this PR incorporated the changes from https://github.com/canonical/chisel-releases/pull/969 since the missing headers are required for the `eza` test

## Related issues/PRs

this PR is part of a series of PRs to fix recent 26.04 breakages. most of these PRs are expected to, in some way, affect one-another's ci runs:

- https://github.com/canonical/chisel-releases/pull/973
- https://github.com/canonical/chisel-releases/pull/971
- https://github.com/canonical/chisel-releases/pull/970 **(this PR)**
- https://github.com/canonical/chisel-releases/pull/969
- https://github.com/canonical/chisel-releases/pull/959
- https://github.com/canonical/chisel-releases/pull/958

### Forward porting

n/a

## Checklist

* [x] I have read the [contributing guidelines](
https://github.com/canonical/chisel-releases/blob/main/CONTRIBUTING.md)
* [x] I have tested my changes ([see how](https://github.com/canonical/chisel-releases/blob/main/CONTRIBUTING.md#7-test-your-slices-before-opening-a-pr))
* [x] I have already submitted the [CLA form](
https://ubuntu.com/legal/contributors/agreement)

PR #969, opened 2026-04-01
fix(26.04): add missing `libc6-dev` headers

lczyk:
# Proposed changes

add missing headers which are new in 26.04

note: `libc6` itself if not missing any files in 26.04

## Related issues/PRs

this PR is part of a series of PRs to fix recent 26.04 breakages. most of these PRs are expected to, in some way, affect one-another's ci runs:

- https://github.com/canonical/chisel-releases/pull/973
- https://github.com/canonical/chisel-releases/pull/971
- https://github.com/canonical/chisel-releases/pull/970
- https://github.com/canonical/chisel-releases/pull/969 **(this PR)**
- https://github.com/canonical/chisel-releases/pull/959
- https://github.com/canonical/chisel-releases/pull/958

### Forward porting

n/a, although libBrokenLocale.a is missing in previous releases too, so will make a patch for that

## Checklist

* [x] I have read the [contributing guidelines](
https://github.com/canonical/chisel-releases/blob/main/CONTRIBUTING.md)
* [x] I have tested my changes ([see how](https://github.com/canonical/chisel-releases/blob/main/CONTRIBUTING.md#7-test-your-slices-before-opening-a-pr))
* [x] I have already submitted the [CLA form](
https://ubuntu.com/legal/contributors/agreement)

PR #959, opened 2026-03-13
feat(26.04): update `php8.4` to `php8.5`

lczyk:
# Proposed changes

update of `php8.4` to `php8.5`

- version replace 8.4 -> 8.5
- remove `opcache` which is now a builtin https://wiki.php.net/rfc/make_opcache_required
- add `libgcc-s1_libs` req on `amd64`
- add `setsid` to php server in php:core tests to allow the server to get killed correctly (we should probably do backport this and just generally use this whenever we run the test app in the background)

## Related issues/PRs

- https://github.com/canonical/chisel-releases/issues/960

this PR is part of a series of PRs to fix recent 26.04 breakages. most of these PRs are expected to, in some way, affect one-another's ci runs:

- https://github.com/canonical/chisel-releases/pull/973
- https://github.com/canonical/chisel-releases/pull/971
- https://github.com/canonical/chisel-releases/pull/970
- https://github.com/canonical/chisel-releases/pull/969
- https://github.com/canonical/chisel-releases/pull/959 **(this PR)**
- https://github.com/canonical/chisel-releases/pull/958

### Forward porting
n/a

## Checklist

* [x] I have read the [contributing guidelines](
https://github.com/canonical/chisel-releases/blob/main/CONTRIBUTING.md)
* [x] I have tested my changes ([see how](https://github.com/canonical/chisel-releases/blob/main/CONTRIBUTING.md#7-test-your-slices-before-opening-a-pr))
* [x] I have already submitted the [CLA form](
https://ubuntu.com/legal/contributors/agreement)

PR #958, opened 2026-03-13
fix(26.04): update `libsimdutf29` to `libsimdutf31`

lczyk:
# Proposed changes

`libsimdutf29` is missing in 26.04. this PR updates the slices to point at `libsimdutf31`

installability tests are expected to fail for this PR

## Related issues/PRs

this PR is part of a series of PRs to fix recent 26.04 breakages. most of these PRs are expected to, in some way, affect one-another's ci runs:

- https://github.com/canonical/chisel-releases/pull/973
- https://github.com/canonical/chisel-releases/pull/971
- https://github.com/canonical/chisel-releases/pull/970
- https://github.com/canonical/chisel-releases/pull/969
- https://github.com/canonical/chisel-releases/pull/959
- https://github.com/canonical/chisel-releases/pull/958 **(this PR)**

### Forward porting
n/a

## Checklist

* [x] I have read the [contributing guidelines](
https://github.com/canonical/chisel-releases/blob/main/CONTRIBUTING.md)
* [x] I have tested my changes ([see how](https://github.com/canonical/chisel-releases/blob/main/CONTRIBUTING.md#7-test-your-slices-before-opening-a-pr))
* [x] I have already submitted the [CLA form](
https://ubuntu.com/legal/contributors/agreement)

PR #971, opened 2026-04-01
fix(26.04): remove missing libs

lczyk:
# Proposed changes

remove libs which are now missing in 26.04 and not used by any other slices

installability tests are expected to fail for this PR

## Related issues/PRs

this PR is part of a series of PRs to fix recent 26.04 breakages. most of these PRs are expected to, in some way, affect one-another's ci runs:

- https://github.com/canonical/chisel-releases/pull/973
- https://github.com/canonical/chisel-releases/pull/971 **(this PR)**
- https://github.com/canonical/chisel-releases/pull/970
- https://github.com/canonical/chisel-releases/pull/969
- https://github.com/canonical/chisel-releases/pull/959
- https://github.com/canonical/chisel-releases/pull/958

### Forward porting
n/a

## Checklist


slices/libavcodec61.yaml
alesancor1, line 1: I am a bit conflicted here:

On one hand there's an update for this library - `libavcodec62` so we should follow the upstream and update here as well.

On the other hand, I believe this used to be  a dependency of another package, which dropped it, so `libavcodec61` went orphan. Since there is not a use case for it anymore we can probably just remove it and reintroduce it again when needed.

I kinda lean more towards the latter, since it's been our approach when it comes to slicing. Still, leaving this comment as food for thought, rather than a request for change.

---

Same applies for the other libs in this PR.

alesancor1 (approved):
LGTM, but please read the comment below.

slices/libavcodec61.yaml
lczyk: yep. agreed with the chain of thought, and i agree with the conclusion. let's drop for now, but let's ponder on general approach. for these libs specifically i've added a ticket to the backlog (ROCKS-3026)

lczyk:
merging with 1 review since its a simple change targeting devel

PR #993, opened 2026-04-16
feat: add slice definitions for squid

alesancor1:
# Proposed changes

Slices for `squid`

## Checklist

## Additional Context
Please check the [comment](https://github.com/canonical/chisel-releases/pull/993#pullrequestreview-4122132514) below.


slices/squid.yaml
```diff
@@ -0,0 +1,111 @@
+package: squid
+
+essential:
+  squid_copyright:
+
+slices:
+  core:
```
alesancor1, line 7: I'm considering the possibility to add `minimal` here, since there are many libraries that would not be required for a truly minimal squid. Though, this would require to tamper with the config in a mutation script to, for example, disable `pinger`

slices/squid.yaml
```diff
@@ -0,0 +1,111 @@
...
+
+  # These are part of core and should be private slices
+  runtime:
+    essential:
+      perl_bins:
+    contents:
+      /usr/lib/squid/diskd:
+      /usr/lib/squid/helper-mux:
+      /usr/lib/squid/pinger:
+      /usr/lib/squid/unlinkd:
```
alesancor1, line 41: This is purely internal to `core` so it does not require testing since it does not add a new functionality. (it is tested as part of core itself)

Though, I'd say we can potentially remove this slice, and inline parts of it into `minimal` and `core` - i.e `unlinkd` is required so it'd be inside `minimal` while `pinger` can be disabled via config

slices/squid.yaml
```diff
@@ -0,0 +1,111 @@
...
+    
+  logging:
+    essential:
+      perl_bins:
+    contents:
+      /usr/lib/squid/log_db_daemon:
+      /usr/lib/squid/log_file_daemon:
```
alesancor1, line 48: just like with `runtime`, these can be inlined in `core` / `minimal`. Afaict you can completely disable logging via config so they would go to `core`

slices/squid.yaml
```diff
@@ -0,0 +1,111 @@
...
+
+  auth:
+    essential:
+      libcrypt1_libs:
+      libdbi-perl_libs:
+      libldap2_libs:
+      libpam0g_libs:
+      libsasl2-2_libs:
+      perl_bins:
+      squid_core:
```
alesancor1, line 58: I'm conflicted here:

* `squid_core` is not strictly necessary for this scripts to run, they can be tested standalone, but they won't add any value unless used within `squid`, which is what they were desingned for. Hence why `squid_core` is here, but could potentially be removed (and tests modified)

slices/squid.yaml
```diff
@@ -0,0 +1,111 @@
...
+      /usr/lib/squid/basic_pop3_auth:
+      /usr/lib/squid/basic_radius_auth:
+      /usr/lib/squid/basic_sasl_auth:
+      /usr/lib/squid/basic_smb_auth:
+      /usr/lib/squid/digest_file_auth:
+      /usr/lib/squid/digest_ldap_auth:
+      /usr/lib/squid/negotiate_kerberos_auth:
+      /usr/lib/squid/negotiate_wrapper_auth:
+  
+  acl:
```
alesancor1, line 74: `squid` already contains acls internally, these are just extra external ones. Should it be renamed to `ext_acls`? Leaning more towards doing so

slices/squid.yaml
```diff
@@ -0,0 +1,111 @@
...
+    essential:
+      perl_bins:
+      squid_auth: # Required for group ACLs
```
alesancor1: Jus like before, `squid_auth` is required here only because when integrating with squid, certain ACLs require `%LOGIN` which requires authentication.

However, the scripts themselves can be tested and used standalone, but they don't really add value outside `squid`. So for the sake of making the slice functional, I included `squid_auth` here

slices/squid.yaml
```diff
@@ -0,0 +1,111 @@
...
+      /usr/lib/squid/ext_session_acl:
+      /usr/lib/squid/ext_sql_session_acl:
+      /usr/lib/squid/ext_time_quota_acl:
+      /usr/lib/squid/ext_unix_group_acl:
+      /usr/lib/squid/ext_wbinfo_group_acl:
+  
+  caching:
+    essential:
+      perl_bins:
+      squid_core:
```
alesancor1, line 91: Same as above, can be tested standalone, but doesn't make sense outside of squid

slices/squid.yaml
```diff
@@ -0,0 +1,111 @@
...
+    contents:
+      /usr/lib/squid/storeid_file_rewrite:
+  
+  cert:
```
alesancor1, line 95: It's just a bash script, should it be renamed to `cert_tool`? I'm leaning towards doing so

tests/spread/integration/squid/test_acl.sh
alesancor1, line 1: All the ACLs (perl scripts) here can be tested standalone by just running:

```
echo "<input>" | chroot "${rootfs}" <script> <options> | grep "OK"
```

However, I implemented an actual squid test which is far more complete and actually tests the functionality of the scripts within `squid`, which is the real use case.

tests/spread/integration/squid/test_acl.sh
```diff
@@ -0,0 +1,99 @@
...
+source "$(dirname "$0")/helpers.sh"
+
+# sqlite3_bins is only required for the SQL session ACL test
+# passwd_bins is required to create users for the ACL tests
+# libpam-runtime_config is required for the chpasswd to work
+rootfs="$(install-slices \
+    squid_acl \
+    base-files_base \
+    base-passwd_data \
+    libc-bin_nsswitch \
```
alesancor1, line 12: I think `libc-bin_nsswitch` here should be part of `core` (even `minimal`) as it is required to resolve the `proxy` username when `squid` drops privileges upon start. Otherwise it won't work.

tests/spread/integration/squid/test_acl.sh
```diff
@@ -0,0 +1,99 @@
...
+CREATE TABLE sessions (
+  id TEXT PRIMARY KEY,
+  user TEXT,
+  enabled INTEGER
+);
+INSERT INTO sessions (id, user, enabled)
+VALUES ('127.0.0.1 -', 'testuser', 1);
+INSERT INTO sessions (id, user, enabled)
+VALUES ('::1 -', 'testuser', 1);
+EOF
```
alesancor1, line 96: This can probably be done outside the `chroot` and copied inside so we avoid introducing `sqlite3` which is purely used for testing purposes.

tests/spread/integration/squid/test_auth.sh
```diff
@@ -0,0 +1,164 @@
...
+echo "auth_param basic program /usr/lib/squid/basic_pam_auth -n squid" >> "${rootfs}/etc/squid/squid.conf"
+echo "acl auth_users proxy_auth REQUIRED" >> "${rootfs}/etc/squid/squid.conf"
+echo "http_access allow auth_users" >> "${rootfs}/etc/squid/squid.conf"
+echo "http_access deny all" >> "${rootfs}/etc/squid/squid.conf"
+
+# PAM service file
+mkdir -p "${rootfs}/etc/pam.d"
+cat > "${rootfs}/etc/pam.d/squid" <<EOF
+auth    required pam_unix.so debug
+account required pam_unix.so debug
```
alesancor1, line 124: `pam_unix.so` is brought in by `libpam-modules_libs`. I'm not sure to what extent this is really required or just for testing:

* On one hand, it's a similar case to a database, where a developer is free to use `sqlite3` or any other, so they are free to create/install a different module from a different package, so `libpam-modules_libs` is not **really** required

* On the other hand, unlike a database which can be `sqlite3` or any other, to me this ones seems like a very basic dependency that is constraining the PAM authentication, so it does make sense to include it.

Considering that adding is always safer than removing slices, I've left it out for now.

alesancor1 (commented):
Bunch of comments, open questions and reasoning explanations below:

slices/squid.yaml
alesancor1: Also, I think `squid_cache` would be a better name for this slice

slices/libdbi-perl.yaml
```diff
@@ -0,0 +1,31 @@
+package: libdbi-perl
+
+essential:
+  libdbi-perl_copyright:
+
+slices:
+
+  # If these bins are ever needed, uncomment the following lines
```
lczyk: appreciate the future note 👍 

just to add, there is a couple more slices like this which could be added here: a bunch of headers at `/usr/lib/*-linux-*/perl5/5.40/auto/DBI/*.h`, what looks like a perl script at `/usr/lib/*-linux-*/perl5/5.40/dbixs_rev.pl` etc. really do not worry about adding those though. no point putting in the work now if they never end up being used.

slices/libecap3.yaml
```diff
@@ -0,0 +1,15 @@
...
+
+essential:
+  libecap3_copyright:
+
+slices:
+  libs:
+    essential:
+      libc6_libs:
+    contents:
+      /usr/lib/*-linux-*/libecap.so*:
```
lczyk: looks like there might be some arch-dependent linking here. on arm64 i get:

```bash
$ lddtree libecap.so.3
libecap.so.3 (interpreter => None)
    libstdc++.so.6 => /usr/lib/aarch64-linux-gnu/libstdc++.so.6
        libm.so.6 => /usr/lib/aarch64-linux-gnu/libm.so.6
    libc.so.6 => /usr/lib/aarch64-linux-gnu/libc.so.6
    ld-linux-aarch64.so.1 => /usr/lib/aarch64-linux-gnu/ld-linux-aarch64.so.1
    libgcc_s.so.1 => /usr/lib/aarch64-linux-gnu/libgcc_s.so.1
```

which would amount to the following deps:

```
libc6_libs:
libgcc-s1_libs:
libstdc++6_libs:
```

lczyk: alright. i run `lddtree` on all arches (emulated) and i get the same deps on all (although slightly different libs from libstdc++6 for each, but that should not be a problem here since we're not subslicing then libs)

slices/libdbi-perl.yaml
```diff
@@ -0,0 +1,31 @@
...
+  #   essential:
+  #     libdbi-perl_libs:
+  #     perl_bins:
+  #   contents:
+  #     /usr/bin/dbilogstrip:
+  #     /usr/bin/dbiprof:
+  #     /usr/bin/dbiproxy:
+  #     /usr/bin/dh_perl_dbi:
+
+  libs:
```
lczyk, line 46: unsure about the structure here... 

all the `*.pm`s are perl modules, the only `*.so` file is at `/usr/lib/aarch64-linux-gnu/perl5/5.40/auto/DBI/DBI.so` and yet is sliced together with those, and then `*.xst` looks to be some kind of C source. at the very least i think this slice deserves more explanations, and potentially a restructuring

slices/libnss3-tools.yaml
```diff
@@ -0,0 +1,17 @@
+package: libnss3-tools
+
+essential:
+  libnss3-tools_copyright:
+
```
lczyk, line 5: a note here saying that this is only one of the binaries in `libnss3-tools` + commented out `bins` as a breadcrumb comment for future slicers? just an idea. obv non-blocking.

tests/spread/integration/squid/helpers.sh
```diff
@@ -0,0 +1,78 @@
...
+        if [ $retries -ge 5 ]; then
+            echo "FAILED: $test_case"
+            return 1
+        fi
+        retries=$((retries + 1))
+        sleep 1
+    done
+
+    echo "OK: $test_case"
+}
```
lczyk: nit: no newline at the end of the file

tests/spread/integration/squid/test_caching.sh
```diff
@@ -0,0 +1,31 @@
...
+mkdir -p "${rootfs}/etc/squid"
+printf "^http://example.com/test/.*$\thttp://example.com/storeid/test/\n" > "${rootfs}/etc/squid/storeid.map"
+
+# Add logging to grep for it later
+echo "logformat storeid_test %ru -> %note" >> "${rootfs}/etc/squid/squid.conf"
+echo "access_log stdio:/var/log/squid/storeid.log storeid_test" >> "${rootfs}/etc/squid/squid.conf"
+
+restart_squid
+curl -x "http://localhost:3128" "http://example.com/test/12345"
+cat "${rootfs}/var/log/squid/storeid.log" | grep "http://example.com/storeid/test"
```
lczyk: nit: no newline at the end of the file

tests/spread/integration/squid/helpers.sh
```diff
@@ -0,0 +1,78 @@
+#!/usr/bin/env bash
+# spellchecker: ignore rootfs
+
+setup_squid() {
+    # Mount dev into the chroot
+    mkdir -p "${rootfs}/dev" "${rootfs}/proc" "${rootfs}/sys"
```
lczyk:
```suggestion
    mkdir -p "$rootfs/dev" "$rootfs/proc" "$rootfs/sys"
```

nit: all the `rootfs`'s dont need curlies. just makes the code a bit cleaner.

tests/spread/integration/squid/helpers.sh
```diff
@@ -0,0 +1,78 @@
...
+        pkill -f "chroot ${rootfs}/ /usr/sbin/squid-gnutls -N" || true
+        pkill -f "/usr/sbin/squid-gnutls -N" || true
+    done
+}
+
+restart_squid() {
+    cleanup
+
+    # Start squid in the background
+    chroot "${rootfs}/" /usr/sbin/squid-gnutls -N &
```
lczyk: was this where `setsid` was not working?

slices/squid.yaml
```diff
@@ -0,0 +1,111 @@
...
+      libgnutls30t64_libs:
+      libgssapi-krb5-2_libs:
+      libkrb5-3_libs:
+      libltdl7_libs:
+      libnetfilter-conntrack3_libs:
+      libnettle8t64_libs:
+      libstdc++6_libs:
+      libsystemd0_libs:
+      squid-common_config:
+      squid-common_icons:
```
lczyk: do we need the icons? i mean, its not a major security hole, i don't think, but just wondering whether or not we could trim them..

slices/squid.yaml
lczyk: i'd say leave that for future improvement. it feels like a potentially larger piece of work, especially be config changes.

reminds me a bit of the situation with `systemd` where we left it at just `standard` for now. tbd i guess

slices/squid.yaml
lczyk: on one hand side, since chisel does not support private slices (sad / tbd again maybe..?), i'd preffer these to be inlined. on the other hand it would make this file very unreadable. i think let's leave it like this, test to the best of our ability and maybe resume the discussion on the private slices? @upils 

as for the tests for this slice specifically, id love to see at least a skeleton test which tests the binaries run. for example, i can do

```
/usr/lib/squid/diskd foo bar baz 2>&1 | grep -Fiq "invalid argument"
```

which at least checks the dynamic linker is not complaining  (which, coincidentally, i think it would here since the slice is also missing `libc6_libs` and `libstdc++6_libs` from essentials)

slices/squid.yaml
```diff
@@ -0,0 +1,111 @@
...
+    contents:
+      /etc/squid/conf.d/debian.conf:
+      /etc/squid/squid.conf:
+      /usr/sbin/squid-gnutls:
+      /var/log/squid/:
+      /var/spool/squid/:
+
+  # These are part of core and should be private slices
+  runtime:
+    essential:
```
lczyk, line 73: missing at least:

```
libc6_libs
libcap2_libs
libgcc-s1_libs
libstdc++6_libs
```

from essentials

slices/squid.yaml
lczyk: and, if this slice is less trimmed and more of just a standard installation of squid, i'd call it `standard` not `core` maybe?

slices/squid.yaml
lczyk: i'm not gonna repeat the same comment on all the slices below for brevity, but it does apply.

slices/squid.yaml
lczyk: mmmm... i've come across cases like this before and i've leaned on the side of not including the slice (aka, in this case `squid_core`) and just leaving a comment next to the slice -- along the lines of _"if in doubt, don't include it"_. this it what i would do here too, especially since you say the auth scripts can be tested w/out squid. it's just an opinion though and not a hill i would die on, nor really a hill i would block this PR with if the prevailing opinion is to do otherwise.

slices/squid.yaml
lczyk: `ext-acls` (aka `-` instead of `_`), but yeah, i think so too

slices/squid.yaml
lczyk: [same answer](https://github.com/canonical/chisel-releases/pull/993/changes#r3118951777)

lczyk: > Also, I think `squid_cache` would be a better name for this slice

why introduce a double-barrel name though? or did you mean for the slice to be called `cache` such that you then `chisel cut squid_cache ...` ?

slices/squid.yaml
```diff
@@ -0,0 +1,111 @@
...
+      /usr/lib/squid/ext_wbinfo_group_acl:
+
+  caching:
+    essential:
+      perl_bins:
+      squid_core:
+    contents:
+      /usr/lib/squid/storeid_file_rewrite:
+
+  cert:
```
lczyk:
```suggestion
  cert_tool:
```

lczyk: snap. i made the suggestion before reading https://github.com/canonical/chisel-releases/pull/993/changes#r3094408084 :D

slices/squid.yaml
```diff
@@ -0,0 +1,111 @@
...
+      /usr/lib/squid/unlinkd:
+
+  logging:
+    essential:
+      perl_bins:
+    contents:
+      /usr/lib/squid/log_db_daemon:
+      /usr/lib/squid/log_file_daemon:
+
+  auth:
```
lczyk, line 104: the following files are uncovered and feel like they might belong here? :

```
/usr/lib/squid/basic_fake_auth
/usr/lib/squid/basic_smb_auth.sh
/usr/lib/squid/negotiate_kerberos_auth_test
/usr/lib/squid/ntlm_fake_auth
/usr/lib/squid/security_fake_certverify
/usr/lib/squid/url_fake_rewrite
/usr/lib/squid/url_fake_rewrite.sh
```

just a question though. don't include things which are not needed / we can always add them later 👍

tests/spread/integration/squid/helpers.sh
```diff
@@ -0,0 +1,78 @@
...
+    # Wait for squid to be ready
+    until ss -ltn | grep -q ':3128'; do
+        sleep 0.1
+    done
+}
+
+test_proxy() {
+    local test_case="$1"
+    shift
+    local curl_opts="$@"
```
lczyk: nit: assigning array to a string which will break down on any arg to `test_proxy` with spaces.

`local curl_opts=("$@")` and then `"${curl_opts[@]}"` would be safer i think

tests/spread/integration/squid/helpers.sh
```diff
@@ -0,0 +1,78 @@
...
+
+    echo "Testing: $test_case"
+    retries=0
```
lczyk:
```suggestion
    local retries=0
```

tests/spread/integration/squid/helpers.sh
```diff
@@ -0,0 +1,78 @@
...
+
+reset_squid_conf() {
+    # Keep original base config (everything before auth)
+    sed -i '/# TEST START/,$d' "${rootfs}/etc/squid/squid.conf"
+    echo "# TEST START" >> "${rootfs}/etc/squid/squid.conf"
+}
+
+cleanup() {
+    # Kill existing squid process
+    while ps -aux | grep -q "[s]quid-gnutls"; do
```
lczyk, line 58: nit: this could attempt to clean up forever (aka until killed by spread). maybe timeout or retries?

tests/spread/integration/squid/helpers.sh
```diff
@@ -0,0 +1,78 @@
...
+    cleanup
+
+    # Start squid in the background
+    chroot "${rootfs}/" /usr/sbin/squid-gnutls -N &
+    trap cleanup EXIT
+
+    # Wait for squid to be ready
+    until ss -ltn | grep -q ':3128'; do
+        sleep 0.1
+    done
```
lczyk, line 93: 2 nits / small issues:

1) another infinite wait if squid fails to restart
2) potential race with old squid still listening on 3128

something like this maybe??

```bash
restart_squid() {
    cleanup

    # Wait for port to be free (old squid fully gone) before starting new
    for _ in $(seq 1 10); do
        ss -ltn | grep -q ':3128 ' || break
        sleep 1
    done
    if ss -ltn | grep -q ':3128 '; then
        echo "port 3128 still in use after cleanup" >&2
        return 1
    fi

    # Start squid in the background
    chroot "${rootfs}/" /usr/sbin/squid-gnutls -N &
    squid_pid=$!
    trap cleanup EXIT

    # Wait for this squid to listen, bounded (~10s)
    for _ in $(seq 1 10); do
        if ! kill -0 "$squid_pid" 2>/dev/null; then
            echo "squid exited before opening :3128" >&2
            return 1
        fi
        ss -ltn | grep -q ':3128 ' && return 0
        sleep 1
    done

    echo "squid did not open :3128 within 10s" >&2
    kill "$squid_pid" 2>/dev/null || true
    return 1
}
```

or maybe move the loop which waits for cleanup to be done inside of `cleanup` ?

tests/spread/integration/squid/helpers.sh
```diff
@@ -0,0 +1,78 @@
...
+
+setup_squid() {
+    # Mount dev into the chroot
+    mkdir -p "${rootfs}/dev" "${rootfs}/proc" "${rootfs}/sys"
+    mount --rbind /dev "${rootfs}/dev"
+
+    # Get the uid:gid for the proxy user
+    proxy_uid_gid=$(grep '^proxy:' "${rootfs}/etc/passwd" | cut -d':' -f3,4)
+
+    # Set permissions for squid directories
```
lczyk: i know we talked about this on 1-1 and why it's just `chown`ing the entire root. i would leave more of a comment here acknowledging that this is a bit of a pain in tests + how it fails if its not set up like this. even just something like _"squid really seems to need to own the root of the chroot to not complain about permissions here..."_ ;)

tests/spread/integration/squid/helpers.sh
lczyk, line 1: nit: could not get the tests to run in my setup until i installed `iproute2` in the test env (for `ss`). can we maybe add and `apt install ...` for it? i know lxd base we use has it, hence the `nit`

tests/spread/integration/squid/helpers.sh
```diff
@@ -0,0 +1,78 @@
+#!/usr/bin/env bash
+# spellchecker: ignore rootfs
+
+setup_squid() {
+    # Mount dev into the chroot
+    mkdir -p "${rootfs}/dev" "${rootfs}/proc" "${rootfs}/sys"
+    mount --rbind /dev "${rootfs}/dev"
```
lczyk: never unmounted

tests/spread/integration/squid/test_cert-tool.sh
```diff
@@ -0,0 +1,17 @@
+#!/usr/bin/env bash
+# spellchecker: ignore rootfs
+rootfs="$(install-slices squid_cert base-files_base)"
+
+# Setup
+mkdir -p "${rootfs}/dev"
+mount --rbind /dev "${rootfs}/dev"
+mknod -m 666 "$rootfs/dev/random" c 1 8 || true
+mknod -m 666 "$rootfs/dev/urandom" c 1 9 || true
```
lczyk: surely these are redundant after mounting dev? also, can we just mock `/dev/random` by just echoing a kb of entropy from `/dev/random` instead of mounting `/dev`? i have a slight aversion against mounting things from test host ever since i've nuked my `/dev`... 😒

tests/spread/integration/squid/test_auth.sh
```diff
@@ -0,0 +1,164 @@
...
+echo "auth_param digest program /usr/lib/squid/digest_file_auth -c /etc/squid/auth/digest" >> "${rootfs}/etc/squid/squid.conf"
+echo "auth_param digest realm testrealm" >> "${rootfs}/etc/squid/squid.conf"
+
+echo "acl auth_users proxy_auth REQUIRED" >> "${rootfs}/etc/squid/squid.conf"
+echo "http_access allow auth_users" >> "${rootfs}/etc/squid/squid.conf"
+echo "http_access deny all" >> "${rootfs}/etc/squid/squid.conf"
+
+# Create digest password file manually
+mkdir -p "${rootfs}/etc/squid/auth"
+echo testuser:testrealm:ebff225e1ceb73e026fcc645af3e84f6 > "${rootfs}/etc/squid/auth/digest"
```
lczyk: hardcoded digest. can we generate it inline?


```suggestion
digest_ha1=$(printf '%s' "testuser:testrealm:testpass" | md5sum | cut -d' ' -f1)
echo "testuser:testrealm:${digest_ha1}" > "$rootfs/etc/squid/auth/digest"
```

tests/spread/integration/squid/test_auth.sh
```diff
@@ -0,0 +1,164 @@
+#!/usr/bin/env bash
+# spellchecker: ignore rootfs
+source "$(dirname "$0")/helpers.sh"
+
+# sqlite3_bins is only required for the DB auth test
+# libpam-modules_libs is only required for the PAM auth test (provide pam_unix.so)
```
lczyk: can we split these tests up into separate rootfs's? otherwise there is a pile of slices from one test in rootfs of another and the tests might not catch some missing dep

tests/spread/integration/squid/test_auth.sh
lczyk, line 1: looks like:

```
basic_ldap_auth
basic_pop3_auth
basic_radius_auth
basic_smb_auth
digest_ldap_auth
negotiate_kerberos_auth
negotiate_wrapper_auth
```

are untested. worth adding a comment that this is only a test of the basic auth?

if there any skeleton testing we could do for the above? (like calling `--help` or even `--invalid-flag` 2>&1 | ...` ?)

tests/spread/integration/squid/test_ext-acls.sh
lczyk, line 1: looks like the following are not tested:

```
ext_kerberos_ldap_group_acl
ext_ldap_group_acl
ext_session_acl
ext_time_quota_acl
ext_wbinfo_group_acl
```

<insert same considerations as in [this](https://github.com/canonical/chisel-releases/pull/993/changes#r3119284258) comment>

tests/spread/integration/squid/task.yaml
lczyk, line 1: any point in `logging` tests? even skeleton ones? ( 'not really' is a fine answer. just want to poke though ;) )

tests/spread/integration/squid/helpers.sh
```diff
@@ -0,0 +1,78 @@
...
+}
+
+test_proxy() {
+    local test_case="$1"
+    shift
+    local curl_opts="$@"
+
+    echo "Testing: $test_case"
+    retries=0
+    until curl -s $curl_opts --proxy http://localhost:3128 https://ubuntu.com/ >/dev/null; do
```
lczyk: so, this is going to be really picky, but can we not make this more of a hermetic test? atm it depends on ubuntu.com being up. it feels like a lot of work / redoing things so maybe just as a general food for thought: test-local python `http.server` ? unsure how it would interact with the chroot.. 🤔

tests/spread/integration/squid/helpers.sh
```diff
@@ -0,0 +1,78 @@
...
+    chroot "${rootfs}/" /usr/sbin/squid-gnutls -N &
+    trap cleanup EXIT
+
+    # Wait for squid to be ready
+    until ss -ltn | grep -q ':3128'; do
+        sleep 0.1
+    done
+}
+
+test_proxy() {
```
lczyk, line 96: enhancement idea:

could we add something like

```bash
test_proxy_deny() {
    local test_case="$1"
    shift
    local curl_opts=("$@")

    echo "Testing deny: $test_case"
    if curl -fsS --max-time 5 "${curl_opts[@]}" --proxy http://localhost:3128 https://ubuntu.com/ >/dev/null 2>&1; then
        echo "FAILED (request succeeded, expected deny): $test_case"
        return 1
    fi
    echo "OK (denied): $test_case"
}
```

and then in test_auth:

```
test_proxy "basic_ncsa_auth" --proxy-user testuser:testpass
# but also:
test_proxy_deny "basic_ncsa_auth (no creds)"
test_proxy_deny "basic_ncsa_auth (wrong pass)" --proxy-user testuser:wrongpass
```

??

i've tried this locally and seems to work just fine

lczyk (changes requested):
alright. 1st review done! lots of comments and some missing deps so definitely 'request changes' but its much more 'request lots of small changes' rather than 'request that you change everything' ;)

great work on this. squid is not eazy! 🐙 😬

slices/squid.yaml
alesancor1: Ended up going for it - check https://github.com/canonical/chisel-releases/pull/993/commits/246309e34003ac6a5543369d98009470bdada2cd

slices/squid.yaml
alesancor1: these binaries are ran internally by squid, I added a check for them in the tests for `core` and `standard`, where they are being used. https://github.com/canonical/chisel-releases/pull/993/commits/246309e34003ac6a5543369d98009470bdada2cd

slices/squid.yaml
alesancor1: Having minimal now I think we can include it as an `essential` for the sake of making them minimally functional. The same logic is being followed in our python slices IIRC.

slices/libecap3.yaml
alesancor1: Added https://github.com/canonical/chisel-releases/pull/993/commits/009938d2868bdf8d8ccfb91d5de686567819c73d

slices/libdbi-perl.yaml
alesancor1: https://github.com/canonical/chisel-releases/commit/43282a338e69cb865ae5c33a1214fc88a03943ec

slices/libnss3-tools.yaml
alesancor1: Imo it's all right if there's no comment. When there's a use case for a new binary the slicer will see it missing and simply add it. I don't think there's a need to bloat SDF with comments for this cases.

tests/spread/integration/squid/helpers.sh
alesancor1: https://github.com/canonical/chisel-releases/pull/993/commits/26a597b72cb4a0f8fdd397ca33ab201aa82a2b43

tests/spread/integration/squid/test_caching.sh
alesancor1: https://github.com/canonical/chisel-releases/pull/993/commits/26a597b72cb4a0f8fdd397ca33ab201aa82a2b43

tests/spread/integration/squid/helpers.sh
alesancor1: https://github.com/canonical/chisel-releases/pull/993/commits/26a597b72cb4a0f8fdd397ca33ab201aa82a2b43

tests/spread/integration/squid/helpers.sh
alesancor1: Yep, I found that trying to kill the process in a loop was more reliable than setsid. Not sure why, to be investigated

slices/squid.yaml
alesancor1: Addressed as part of https://github.com/canonical/chisel-releases/pull/993/commits/246309e34003ac6a5543369d98009470bdada2cd

slices/squid.yaml
alesancor1: Removed the slice entirely, now inlined in `core` and `standard` https://github.com/canonical/chisel-releases/pull/993/commits/246309e34003ac6a5543369d98009470bdada2cd

slices/squid.yaml
alesancor1: Afaict they are test/dev files that don't add actual functionality to squid.

tests/spread/integration/squid/helpers.sh
alesancor1: added umount in the cleanup

tests/spread/integration/squid/test_cert-tool.sh
alesancor1: For some reason when running the tests on my end they (`random` and `urandom`) were not being mounted so I enforced the creation.

Will investigate whether we can just mock other `/dev` contents so we don't need to mount it here.

tests/spread/integration/squid/test_auth.sh
alesancor1: See https://github.com/canonical/chisel-releases/pull/993#discussion_r3213853947

tests/spread/integration/squid/test_ext-acls.sh
alesancor1: See https://github.com/canonical/chisel-releases/pull/993#discussion_r3213853947

tests/spread/integration/squid/task.yaml
alesancor1: Slice removed, tested as part of standard/core

slices/squid.yaml
alesancor1: They have been inlined in `core` and `standard`

slices/squid.yaml
alesancor1: https://github.com/canonical/chisel-releases/pull/993/commits/8d59e251c4be0fa4a67980f4a829a2865318046e

slices/squid.yaml
alesancor1: https://github.com/canonical/chisel-releases/pull/993/commits/8d59e251c4be0fa4a67980f4a829a2865318046e

slices/squid.yaml
alesancor1: https://github.com/canonical/chisel-releases/pull/993/commits/8d59e251c4be0fa4a67980f4a829a2865318046e

tests/spread/integration/squid/helpers.sh
alesancor1: https://github.com/canonical/chisel-releases/pull/993/commits/70484686b05d5ae62fcde76972db1d20d208db06

tests/spread/integration/squid/helpers.sh
alesancor1: https://github.com/canonical/chisel-releases/pull/993/commits/dddc6661d5fc0b3d0909a3a2a3836cb6bbc4845c

tests/spread/integration/squid/helpers.sh
alesancor1: fixed https://github.com/canonical/chisel-releases/pull/993/commits/93606e21e110533c49989d9eedc180cef64bfbc3

tests/spread/integration/squid/helpers.sh
alesancor1: fixed in https://github.com/canonical/chisel-releases/pull/993/commits/93606e21e110533c49989d9eedc180cef64bfbc3

tests/spread/integration/squid/helpers.sh
alesancor1: https://github.com/canonical/chisel-releases/pull/993/commits/0dfcc7feb629fcd9a8862a2e33baa16227157120

tests/spread/integration/squid/helpers.sh
alesancor1: I'm a bit reluctant to add deps for other testing environments that are not the one(s) it is supposed to be running, Imo everything installed should be what's required to make it run within spread.

tests/spread/integration/squid/test_auth.sh
alesancor1: https://github.com/canonical/chisel-releases/pull/993/commits/ca0d1767922e7f788288a8b8d2b772ed84235fc9

tests/spread/integration/squid/test_auth.sh
alesancor1: https://github.com/canonical/chisel-releases/pull/993/commits/a6a85d39b97907bdad9cd85e85017afb1aac6378

tests/spread/integration/squid/helpers.sh
alesancor1: Let's leave this for a future refactor

tests/spread/integration/squid/helpers.sh
alesancor1: Added - https://github.com/canonical/chisel-releases/pull/993/commits/3c5a5285365aefeec2a3d16cbcfc191bbf118194

slices/libdbi-perl.yaml
```diff
@@ -0,0 +1,29 @@
...
+  libdbi-perl_copyright:
+
+slices:
+
+  modules:
+    essential:
+      libdbi-perl_driver:
+      libdbi-perl_libs:
+    contents:
+      /usr/lib/*-linux-*/perl5/5.40/**/*.pm:
```
vpa1977: This location is defined by the perl default : `perl5/5.40`, consider changing to `perl5/*/` to protect against perl upgrades.

alesancor1: Good catch! Changes applied in https://github.com/canonical/chisel-releases/pull/993/commits/bd8b05685caa525e159987f63d3b2218eefc60b5. 

Thanks!

lczyk:
- fix for hint validation issue: https://github.com/canonical/chisel-releases/pull/1018

slices/libdbi-perl.yaml
```diff
@@ -0,0 +1,28 @@
...
+      libdbi-perl_driver:
+      libdbi-perl_libs:
+    contents:
+      /usr/lib/*-linux-*/perl5/**/*.pm:
+
+  libs:
+    essential:
+      libc6_libs:
+    contents:
+      /usr/lib/*-linux-*/perl5/**/*.so:
```
lczyk:
```suggestion
      /usr/lib/*-linux-*/perl5/*/auto/DBI/DBI.so:
```

i think this is better, no? 1) it does not use `**` and 2) if `libdbpi-perl` decides to sneak in additional so which is liked against more than libc, then we silently get a half-broken slice.

slices/libdbi-perl.yaml
```diff
@@ -0,0 +1,28 @@
...
+  libdbi-perl_copyright:
+
+slices:
+
+  modules:
+    essential:
+      libdbi-perl_driver:
+      libdbi-perl_libs:
+    contents:
+      /usr/lib/*-linux-*/perl5/**/*.pm:
```
lczyk: same comment as https://github.com/canonical/chisel-releases/pull/993/changes#r3334208806 -- i'd do `perl5/*/...` and you only need to list `DBI`, `DBD`, `Bundle` and `Win32`, as well as `/usr/lib/*-linux-*/perl5/5.40/DBI.pm` itself (and, arguably, can drop the `win32` one).

also, your glob currently will collide with *a lot* of stuff:

```bash
$apt-file search --regexp '/usr/lib/[^/]*-linux-[^/]*/perl5/[^/]*/.*\.pm' | cut -d: -f1 | sort -u | wc -l
612
```

with explicitly listing one more level it collides much much less, and its only really an issue for 10 packages:

```bash
$ apt-file search --regexp '/usr/lib/[^/]*-linux-[^/]*/perl5/[^/]*/(DBI|DBD|Bundle|Win32)/.*pm' | cut -d: -f1 | sort -u | tee /dev/fd/2 | wc -l
libapache2-mod-perl2
libdbd-firebird-perl
libdbd-mariadb-perl
libdbd-mysql-perl
libdbd-odbc-perl
libdbd-pg-perl
libdbd-sqlite3-perl
libdbd-sybase-perl
libdbi-perl
libsnmp-perl
10
```

slices/libdbi-perl.yaml
lczyk, line 1: just noting that this slice is currently not slicing any bins, headers and the `dbixs_rev.pl` file ( which should probably go with the headers since its build-time accessible version info )

tests/spread/integration/squid/test_standard.sh
```diff
@@ -0,0 +1,75 @@
...
+
+# Setup mysql for testing
+apt install -y mysql-server
+apt download libdbd-mysql-perl && dpkg -x libdbd-mysql-perl_*.deb "$rootfs/" && rm libdbd-mysql-perl_*.deb
+trap "pkill mysqld; wait; cleanup" EXIT
+
+# Enable mysql_native_password plugin
+echo "[mysqld]" > /etc/mysql/mysql.conf.d/mysql.cnf
+echo "mysql_native_password=ON" >> /etc/mysql/mysql.conf.d/mysql.cnf
+service mysql restart
```
lczyk, line 30:
```suggestion
# equivalent to `service mysql restart` but does not rely on systemd
pkill mysqld 2>/dev/null || true
mysqld --user=mysql &
until mysqladmin ping --silent 2>/dev/null; do sleep 1; done
```

minor, and ff to ignore since lxd tests *are* the king, but with this patch i get 100% test success when testing with (adhoc) docker backend ;)

slices/squid.yaml
```diff
@@ -0,0 +1,166 @@
...
+
+          # Disable pid file
+          pid_filename none
+      /etc/squid/squid.conf:
+        mutable: true
+        text: ""
+      /usr/sbin/squid-gnutls:
+    mutate: |
+      files = content.list("/etc/squid")
+      if "default.conf" not in files:
```
lczyk:
```suggestion
      if "default.conf" not in files:  # if squid_core is not installed
```

also, nit but this is an implicit soft dependency on default.conf. i'd add a regression test in case someone moves the default.conf and does not see that they need to update minimal too. i've tested the following and it works fine:

```bash
#!/usr/bin/env bash
# spellchecker: ignore rootfs
source "$(dirname "$0")/helpers.sh"

# regression test:
# squid_minimal's mutate checks for "default.conf" by name to pick
# between the minimal config and the deb default. If squid_core ever renames
# that file, or the file gets moved to a different slice,
# the check in mutate can silently misbehave. Both
# branches are exercised here to check.

# Branch 1: squid_core co-installed -> default.conf present -> deb config used.
rootfs="$(install-slices squid_core)"
! grep -qF "visible_hostname squid.minimal" "$rootfs/etc/squid/squid.conf"
cleanup

# Branch 2: squid_minimal alone -> no default.conf -> minimal config used.
rootfs="$(install-slices squid_minimal)"
grep -qF "visible_hostname squid.minimal" "$rootfs/etc/squid/squid.conf"
cleanup
```

slices/squid.yaml
```diff
@@ -0,0 +1,166 @@
...
+    contents:
+      /etc/squid/conf.d/debian.conf:
+      /etc/squid/default.conf:
+        until: mutate
+        copy: /etc/squid/squid.conf
+      /usr/lib/squid/log_file_daemon:
+      /usr/lib/squid/unlinkd:
+      # Icons are not strictly needed, but the
+      # directory is required to exist.
+      /usr/share/squid/icons/: { make: true }
```
lczyk, line 85: even though https://github.com/canonical/chisel-releases/pull/993/changes#diff-57dbce637af6a749e8b61fd31a25895ce0021432e673b18e251eab7dde1e7205R50 ?? squids be weird 🤷‍♀️

slices/squid.yaml
```diff
@@ -0,0 +1,166 @@
...
+      /usr/lib/squid/basic_pop3_auth:
+      /usr/lib/squid/basic_radius_auth:
+      /usr/lib/squid/basic_sasl_auth:
+      /usr/lib/squid/basic_smb_auth:
+      /usr/lib/squid/digest_file_auth:
+      /usr/lib/squid/digest_ldap_auth:
+      /usr/lib/squid/negotiate_kerberos_auth:
+      /usr/lib/squid/negotiate_wrapper_auth:
+
+  ext-acls:
```
lczyk, line 127: ok, so, as weird as it sounds, i dont think this slice actually needs `squid_auth`. squid communicates between its auth subsystem and ext-acl helpers purely through plain string arguments via stdin. the ext-acl helper never invokes auth helper binary. so, while *squid* needs the auth helpers to figure out what to pass to ext-acl helpers, the helpers themselsves dont care. i've managed to get the following to work:

```diff
   ext-acls:
     essential:
+      libdbi-perl_modules:
+      libldap2_libs:
+      libsasl2-2_libs:
+      libtdb1_libs:
       perl_bins:
-      squid_auth:   # Required for group ACLs
+      squid_minimal:
     contents:
       /usr/lib/squid/ext_file_userip_acl:
       /usr/lib/squid/ext_kerberos_ldap_group_acl:
```

(note additionally sliced (trivial) `libtdb1_libs` for `ext_session_acl` and `ext_time_quota_acl` which is also missing here)

and a small auth perl shim in the ext-acls test:

```diff
 chroot "$rootfs/" usermod -aG squidgrp testuser
 echo "testuser:testpass" | chroot "$rootfs/" chpasswd

-# Add basic authentication (required for %LOGIN usage in ACLs)
-echo "auth_param basic program /usr/lib/squid/basic_getpwnam_auth" >> "$rootfs/etc/squid/squid.conf"
+# inimal auth shim: accepts all credentials. use perl which is already in rootfs.
+cat > "$rootfs/usr/lib/squid/basic_test_auth" <<'EOF'
+#!/usr/bin/perl
+$| = 1;
+while (<STDIN>) { print "OK\n" }
+EOF
+chmod +x "$rootfs/usr/lib/squid/basic_test_auth"
+
+echo "auth_param basic program /usr/lib/squid/basic_test_auth" >> "$rootfs/etc/squid/squid.conf"
 echo "acl auth_users proxy_auth REQUIRED" >> "$rootfs/etc/squid/squid.conf"

 # FILE USER-IP ACL
```

lczyk: also, id test all the exts: https://pastebin.ubuntu.com/p/qGycpBDSMV/

tests/spread/integration/squid/test_ext-acls.sh
```diff
@@ -0,0 +1,101 @@
...
+
+restart_squid
+test_proxy "ext_unix_group_acl" --proxy-user testuser:testpass
+
+
+# SQL SESSION ACL
+# ------------------------------------------------
+reset_squid_conf
+
+# Manually add the sqlite perl module (Only for tests)
```
lczyk: mmm.... i guess ok for tests but i'd slice and update this afterwards to use the slices.

lczyk (changes requested):
a pile of minors and one major-ish (the `ext-acls` rework). with all of that though, its looking really nice and sound! :) (and really hard 😬 well done!)

slices/libdbi-perl.yaml
alesancor1: Done as part of: ea5ce412e5db594e55b02efc17625bec05277c2b

slices/libdbi-perl.yaml
alesancor1: Ended up expanding the contents to prevent such conflicts: ea5ce412e5db594e55b02efc17625bec05277c2b

tests/spread/integration/squid/test_standard.sh
alesancor1: Couldn't get it to work in spread, hangs indefinitely.

I'm sure there might be a hacky way to run the daemon directly, but they seem to be a bit flaky so I'm sticking to systemd for now

slices/squid.yaml
alesancor1: That's a nice addition, thanks! Works like a charm! 8843bec5ea63ebcecaa8f47ce6ea2dbaf89a9d93

I've put it there with the rest of scripts for now (also seems that was the intention, since you are sourcing the `helpers.sh` file) but maybe we could consider `tests/spread/regression/<package>` directories if we have more of these atp

slices/squid.yaml
alesancor1: This is core, so it's using the default config, which means it points to `/usr/share/squid/icons/`. Even when the contents are not really needed, the directory still needs to exists.

Although this may suggest that icons should then be included, strictly speaking they would not belong to the "core use cases", hence why I left them out of this slice.

Happy to know your thoughts on this :+1:

slices/libdbi-perl.yaml
alesancor1: Let's add headers in a future iteration, when needed

slices/squid.yaml
alesancor1: Made the changes in: https://github.com/canonical/chisel-releases/pull/993/commits/78c4bc16272c55927ebbf3b36f847c06b2b96253

Still I couldn't manage to get ext_session_acl and time_quota tests to work, so I've left them commented for now. Let's review this 1:1

lczyk (approved):
🐙🌊 what a journey! looks great. ty!

slices/squid.yaml
alesancor1: Out of our latest 1:1

the session_acl doesn't work due to a bug in `libtdb1` that causes an error when the helper tries to open the session db file. The test has been commented out for the time being.

The `ext_quota_acl_test` has been fixed and is passing now.

See https://github.com/canonical/chisel-releases/pull/993/commits/f8b360f7fbf99ec84c4f5fe311912d15822bf035

slices/libdbi-perl.yaml
```diff
@@ -0,0 +1,58 @@
+package: libdbi-perl
+
+essential:
+  libdbi-perl_copyright:
+
+slices:
+
+  modules:
```
zhijie-yang, line 8: This slice should at least depend on a `perl-base_bins` or `perl_bins`. Otherwise, these `.pm` files won't work.

slices/libecap3.yaml
```diff
@@ -0,0 +1,17 @@
...
+  libecap3_copyright:
+
+slices:
+  libs:
+    essential:
+      libc6_libs:
+      libgcc-s1_libs:
+      libstdc++6_libs:
+    contents:
+      /usr/lib/*-linux-*/libecap.so*:
```
zhijie-yang:
```suggestion
      /usr/lib/*-linux-*/libecap.so.3*:
```

The lib name contains `3`

slices/libtdb1.yaml
```diff
@@ -0,0 +1,15 @@
...
+
+essential:
+  libtdb1_copyright:
+
+slices:
+  libs:
+    essential:
+      libc6_libs:
+    contents:
+      /usr/lib/*-linux-*/libtdb.so*:
```
zhijie-yang:
```suggestion
      /usr/lib/*-linux-*/libtdb.so.1*:
```

Note the `1` in the package name.

slices/squid-common.yaml
```diff
@@ -0,0 +1,24 @@
...
+
+slices:
+  data:
+    essential:
+      squid-langpack_data:
+    contents:
+      /usr/share/squid/errors:  # Symlink to /usr/share/squid-langpack
+      /usr/share/squid/mib.txt:
+
+  config:
```
zhijie-yang: I'd use `mime` or `mime-config`, as this is more like a mime lookup table instead of a normal "config".

slices/squid-common.yaml
```diff
@@ -0,0 +1,24 @@
...
+    contents:
+      /usr/share/squid/mime.conf:
+
+  icons:
+    contents:
+      /usr/share/squid/**/*.png:
```
zhijie-yang:
```suggestion
      /usr/share/squid/icons/**.png:
```

nit

slices/squid-langpack.yaml
```diff
@@ -0,0 +1,13 @@
+package: squid-langpack
+
+essential:
+  squid-langpack_copyright:
+
+slices:
+  data:
+    contents:
+      /usr/share/squid-langpack/**/*:
```
zhijie-yang, line 9: Q (non-blocking): Shall we make a further strip-down on different languages?

slices/squid.yaml
```diff
@@ -0,0 +1,170 @@
...
+        text: ""
+      /usr/sbin/squid-gnutls:
+    mutate: |
+      files = content.list("/etc/squid")
+      if "default.conf" not in files:  # If squid_core not installed
+        data = content.read("/etc/squid/minimal.conf")
+        content.write("/etc/squid/squid.conf", data)
+      else:
+        data = content.read("/etc/squid/default.conf")
+        content.write("/etc/squid/squid.conf", data)
```
zhijie-yang, line 65: Nice!

tests/spread/integration/squid/test_regression.sh
```diff
@@ -0,0 +1,20 @@
...
+
+# Branch 1: squid_core co-installed -> default.conf present -> deb config used.
+rootfs="$(install-slices squid_core)"
+! grep -qF "visible_hostname squid.minimal" "$rootfs/etc/squid/squid.conf"
+cleanup
+
+# Branch 2: squid_minimal alone -> no default.conf -> minimal config used.
+rootfs="$(install-slices squid_minimal)"
+grep -qF "visible_hostname squid.minimal" "$rootfs/etc/squid/squid.conf"
+cleanup
```
zhijie-yang:
```suggestion
cleanup

```

zhijie-yang (commented):
Well done. Thanks for the code @alesancor1 and the first round review @lczyk ! Some nitpicks need to be addressed before the PR is ready to go.

slices/libdbi-perl.yaml
alesancor1: Good call! I included `perl_bins` here and removed it from some slices in `squid`, where is not required anymore.

Also tried `perl-base_bins` instead, but seems the entire `perl_bins` are required.

See https://github.com/canonical/chisel-releases/pull/993/commits/0e175f4ebdf145b90ec6976f5ad4175184d27bd6

slices/libecap3.yaml
alesancor1: d4c7ad8c95e76707993bc41de447b0eae80939e2

slices/libtdb1.yaml
alesancor1: d4c7ad8c95e76707993bc41de447b0eae80939e2

slices/squid-common.yaml
alesancor1: 8e7c85da234061f3e63238dcf04d24ba1dda70bf

slices/squid-langpack.yaml
alesancor1: Potentially, but not right now, as we don't have a use case for it at the moment.

tests/spread/integration/squid/test_regression.sh
alesancor1: c2376c945ead2bce3218ff03f18b7f666757de4d :P

slices/squid-common.yaml
alesancor1: `mime-config` seems appropiate: 92aa21bcb45a32a6e3b2cf9f19c620fb6e41370b

lczyk:
blocking:

- https://github.com/canonical/squid-rock/pull/4
- https://github.com/canonical/squid-rock/pull/5

zhijie-yang (approved):
No more nits from my side, and good to go!

PR #1001, opened 2026-04-23
ci(26.04): lxd launch fallback

lczyk:
# Proposed changes

add fallback logic to `lxc launch` in `spread.yaml`. when launching the `distro-info --latest` release fails, it's possible that the image has just been released and isn't available in the LXD `ubuntu:` channel yet. so fall back to the `ubuntu-daily:` channel, which should have the new release.

this also:
- force-deletes any pre-existing `$SPREAD_SYSTEM` container before launching, so a stale instance from a previous run does not block allocation
- uses spread's [`FATAL`](https://github.com/canonical/spread#Functions) (instead of `exit 1`) when launch definitively fails, so spread does not retry a broken allocation
- bails out of the `is-system-running` wait loop with `FATAL` if systemd reports `degraded`, instead of looping forever

this unblocks 26.04 spread runs when the image lags behind the release announcement.

### Forward porting

- https://github.com/canonical/chisel-releases/pull/1001 **(this PR)**
- https://github.com/canonical/chisel-releases/pull/1008
- https://github.com/canonical/chisel-releases/pull/1012
- https://github.com/canonical/chisel-releases/pull/1013
- https://github.com/canonical/chisel-releases/pull/1014

## Checklist


lczyk:
tests are failing atm due to the `degraded` status of the LXD machine using the new `FATAL` exit, so they're actually proving the PR, at least partially, correct

spread.yaml
```diff
@@ -29,10 +29,28 @@ backends:
...
+      fi
+
+      if [ -n "${launch_rc-}" ]; then
+        FATAL "Failed to launch LXD instance for release $release"
+      fi
+
       while true; do
         status=$(lxc exec $SPREAD_SYSTEM -- systemctl is-system-running || true)
         [ "$status" = "running" ] && break
+        if [ "$status" = "degraded" ]; then
```
lczyk, line 57: to investigate: can we ever get out of degraded?

alesancor1 (approved):
LGTM, thanks!

lczyk:
@alesancor1 requesting rereview since i've bolted on the 26.04 console-getty bug skip (https://github.com/canonical/lxd/issues/16860#issuecomment-4450647779). it's proactively gated behind `if [ "$release" = "resolute" ]; then` in case we forget to remove this when opening 26.10 when, hopefully, the bug will not be there anymore

spread.yaml
lczyk: answer: yes, if the unit has a restart its possible to get a degraded status which then recovers. in this case (see my comment below + https://github.com/canonical/lxd/issues/16860#issuecomment-4450647779) the issue is the `console-getty.service`. it does have a retry but never succeeds and gets culled by the systemd restart limit.

lczyk: re FATAL here, i guess this means that we sometimes could actually wait a bit and see if "degraded" clears but i'd revisit this only is the currently proposed somewhat overly aggresive solution bites us too much -- kills too many lxds in actual tests when they woudld have recovered. in practise i've never seen `starting` -> `degraded` -> `running`. it's always been `starting` -> `running` in healthy instances.

lczyk: another consideration is whether we should also check that the machine is not in `maintenence` or `offline` -- two more terminal states, but again, in practise does not really happen and easy to add if we ever see tests hanged on this.

slices/bash.yaml
```diff
@@ -3,6 +3,8 @@ package: bash
 essential:
   bash_copyright:
 
+# hello
+
```
alesancor1: remember to rollback this

spread.yaml
```diff
@@ -29,10 +29,35 @@ backends:
       # but that would effectively change the host's LXC configurations.
 
       echo "Allocating $SPREAD_SYSTEM..."
-      lxc launch --ephemeral ubuntu:$release $SPREAD_SYSTEM
+      lxc delete --force $SPREAD_SYSTEM 2>/dev/null || true
```
alesancor1: Why deleting beforehand? I mean, it's all good but seems unrelated to the issue being fixed, is it a drive-by?

spread.yaml
```diff
@@ -29,10 +29,35 @@ backends:
...
+        # *just* been released and isn't available in the 'ubuntu' channel yet.
+        # fall back to the daily channel, which should have the new release.
+        echo "Release $release failed to launch. Trying the daily channel..."
+        launch_rc=""
+        lxc launch --ephemeral ubuntu-daily:$release $SPREAD_SYSTEM || launch_rc=$?
+      fi
+
+      if [ -n "${launch_rc-}" ]; then
+        FATAL "Failed to launch LXD instance for release $release"
+      fi
```
alesancor1, line 45: This is a drive-by as well isn't it? Looks good but it's increasing the scope of this PR

spread.yaml
```diff
@@ -29,10 +29,35 @@ backends:
...
+      if [ "$release" = "resolute" ]; then
+        lxc exec $SPREAD_SYSTEM -q- systemctl mask --now console-getty.service || true
+      fi
+
       while true; do
         status=$(lxc exec $SPREAD_SYSTEM -- systemctl is-system-running || true)
         [ "$status" = "running" ] && break
+        if [ "$status" = "degraded" ]; then
+          FATAL "System is degraded, aborting"
+        fi
```
alesancor1, line 59: Same here

alesancor1 (commented):
Looks good, though there are few things being addressed aside from the issue to be fixed.

The main issue here is gated to resolute, but the other changes would need to be backported for consistency, so either:

* Backport this PR (except for resolut's changes, ofc)
* Split this PR so we can quickly merge the systemd fix, and address the rest of the changes + backports separately

In any of the cases, it LGTM, so I'l leave the choice to you and  I'll happy to approve it then :+1:

spread.yaml
lczyk: yeah, but it's a minor one. removed in https://github.com/canonical/chisel-releases/pull/1001/commits/af9c06ab98fea99d1afd4702de2943bf820a508e to keep PR focused

spread.yaml
lczyk: you mean

```
if [ -n "${launch_rc-}" ]; then
  FATAL "Failed to launch LXD instance for release $release"
fi
```

bit?

if so, then i guess, yes, but it feels very precient and obvious to put here -- if we failed to launch then we should not go into the status-checking loop.

re backporting: i agree. will make appropriate PRs

alesancor1 (approved):
LGTM, please backport the changes to the other active releases

lczyk:
@alesancor1, see: :+1: 

- https://github.com/canonical/chisel-releases/pull/1008
- https://github.com/canonical/chisel-releases/pull/1012
- https://github.com/canonical/chisel-releases/pull/1013
- https://github.com/canonical/chisel-releases/pull/1014

lczyk:
blocking https://github.com/canonical/chisel-releases/actions/runs/26237981209/job/77519707380?pr=1005

example cli ouptut:

```
...
-----
(... 338 lines above ...)
++ true
+ status=degraded
+ '[' degraded = running ']'
+ sleep 5
+ true
++ lxc exec spread-chisel-releases-resolute -- systemctl is-system-running
++ true
+ status=degraded
+ '[' degraded = running ']'
+ sleep 5
-----
2026-05-23 15:21:10 WARNING: local script running late. Current output:
-----
(... 338 lines above ...)
++ true
+ status=degraded
+ '[' degraded = running ']'
+ sleep 5
+ true
++ lxc exec spread-chisel-releases-resolute -- systemctl is-system-running
++ true
+ status=degraded
+ '[' degraded = running ']'
+ sleep 5
-----
2026-05-23 15:26:10 WARNING: local script running late. Current output:
...
```

spread.yaml
```diff
@@ -29,10 +29,34 @@ backends:
...
+
+      if [ -n "${launch_rc-}" ]; then
+        FATAL "Failed to launch LXD instance for release $release"
+      fi
+
+      # Workaround for systemd 258+ regression where console-getty.service
+      # fails on container boot, leaving the system in 'degraded' state.
+      # See https://github.com/canonical/lxd/issues/16860
+      if [ "$release" = "resolute" ]; then
+        lxc exec $SPREAD_SYSTEM -q- systemctl mask --now console-getty.service || true
```
alesancor1:
```suggestion
        lxc exec $SPREAD_SYSTEM -q -- systemctl mask --now console-getty.service || true
```

alesancor1 (changes requested):

alesancor1 (approved):

cjdcordeiro (approved):

Issue #1031, opened 2026-06-22
bug: flocal extraction broken in pkg-deps


lczyk:
in `.github/scripts/pkg-deps/pkg-deps`

```bash
yq '.slices.[].essential[]' "$f"
```

essential blocks are maps (libc6_libs: -> null). `[]` splats to *values* (null), not keys -> flocal always empty -> spurious "-<dep>" diff vs upstream apt deps on every dep-bearing slice PR (e.g. PR #1029 shows -libc6 / +).

fix:

```bash
yq '.slices.[].essential | keys | .[]' "$f"
```

related to #1030

youneedgreg:
Hi, if this issue is stilll open I would like to take a crack at it, will link a PR soon

PR #1033, opened 2026-06-25
fix(ci): handle list and map essential in pkg-deps

saaicasm:
# Proposed changes

The pkg-deps script builds the list of a slice's dependencies with `yq '.slices.[].essential[]'`. The trailing `[]` returns a node's values, which only works when `essential` is a YAML list. 
On ubuntu-26.04 the `essential` blocks are written as maps instead, so the values are null and the list ends up empty. 
The bot then reports a dependency that isn't really missing (e.g. `-libc6` / `+`).

### slice difference across releases

ubuntu-24.04:
```yaml
core:
  essential:
    - golang-go_core
```

ubuntu-26.04:
```yaml
core:
  essential:
    golang-go_core:
```

`keys` fixes the map case but then breaks the list case (and errors on slices that have no essential at all). 
So the fix checks the node type and reads list items or map keys accordingly:

```
yq '.slices.[].essential | ( (select(tag == "!!seq") | .[]), (select(tag == "!!map") | keys | .[]) )' "$f"
```

### Forward porting

n/a, the script lives only on main.

## Checklist

* [x] I have read the [contributing guidelines](https://github.com/canonical/chisel-releases/blob/main/CONTRIBUTING.md)
* [x] I have already submitted the [CLA form](https://ubuntu.com/legal/contributors/agreement)

## Additional Context

Tested locally with the same yq version CI uses.

### DIFF
--- a/.github/scripts/pkg-deps/pkg-deps
+++ b/.github/scripts/pkg-deps/pkg-deps
@@ -40,7 +40,7 @@ for f in $@; do
   sort | uniq > "$fupstream"
 
   flocal="$(mktemp)"
-  yq '.slices.[].essential[]' "$f" | \
+  yq '.slices.[].essential | ( (select(tag == "!!seq") | .[]), (select(tag == "!!map") | keys | .[]) )' "$f" | \
   sed "s/_.*//; /^$pkg$/d" | sort | uniq > "$flocal"
 
   fdiff="$(mktemp)"


lczyk (approved):
good fix. cheers :)

i confirm this works for all the cases mentioned [here](https://github.com/canonical/chisel-releases/pull/1035#issuecomment-4886801750)

lczyk:
fixes:

- https://github.com/canonical/chisel-releases/issues/1031

alesancor1 (approved):
LGTM, thanks for the fix!

Issue #752, opened 2025-11-19
ci: networking error preparing lxd in tests (1)


lczyk:
Starting some day between 12th and 17th, lots of Spread tests started to fail:

<img width="2071" height="1347" alt="Image" src="https://github.com/user-attachments/assets/e6328e44-e79e-412e-8fce-85885b6bfbab" />

example: https://github.com/canonical/chisel-releases/actions/runs/19497991833/job/55804854143

```
-----
++ dpkg --print-architecture
++ sed -e s/armhf/arm/g -e s/ppc64el/ppc64le/g
+ arch=amd64
+ apt install -y curl wget

WARNING: apt does not have a stable CLI interface. Use with caution in scripts.

Reading package lists...
Building dependency tree...
Reading state information...
curl is already the newest version (8.12.1-3ubuntu1).
curl set to manually installed.
wget is already the newest version (1.24.5-2ubuntu1).
wget set to manually installed.
Solving dependencies...
Summary:
  Upgrading: 0, Installing: 0, Removing: 0, Not Upgrading: 0
+ curl -s https://api.github.com/repos/canonical/chisel/releases/latest
+ awk '/browser_download_url/ && /chisel_v/ && /_amd64\./'
+ cut -d : -f 2,3
+ tr -d '"'
+ xargs wget
wget: missing URL
Usage: wget [OPTION]... [URL]...

Try `wget --help' for more options.
-----
```

looks like we might not be getting the correct link from `api.github.com` although doing

```bash
curl -s https://api.github.com/repos/canonical/chisel/releases/latest | awk
'/browser_download_url/ && /chisel_v/ && /_amd64\./' | cut -d : -f 2,3 | tr -d '"'
```

locally looks as expected

lczyk:
renamed to deduplicate from https://github.com/canonical/chisel-releases/issues/757

lczyk:
<img width="412" height="140" alt="Image" src="https://github.com/user-attachments/assets/b2b1bc2d-5a70-41ec-96b1-33e2250547df" />

this can get bad...

one of the workflows from https://github.com/canonical/chisel-releases/pull/1044

lczyk:
closed by:

- https://github.com/canonical/chisel-releases/pull/1048
- https://github.com/canonical/chisel-releases/pull/1038
- https://github.com/canonical/chisel-releases/pull/1039
- https://github.com/canonical/chisel-releases/pull/1040
- https://github.com/canonical/chisel-releases/pull/1041
[...]

PR #1039, opened 2026-07-05
fix(25.10): use 'latest' endpoint

lczyk:
# Proposed changes

fixes https://github.com/canonical/chisel-releases/issues/752, backport of https://github.com/canonical/chisel-releases/pull/1038

## Related issues/PRs

- https://github.com/canonical/chisel-releases/issues/752

### Forward porting

- https://github.com/canonical/chisel-releases/pull/1048
- https://github.com/canonical/chisel-releases/pull/1038
- https://github.com/canonical/chisel-releases/pull/1039 **(this PR)**
- https://github.com/canonical/chisel-releases/pull/1040
- https://github.com/canonical/chisel-releases/pull/1041
- https://github.com/canonical/chisel-releases/pull/1042

## Checklist

* [x] I have read the [contributing guidelines](
https://github.com/canonical/chisel-releases/blob/main/CONTRIBUTING.md)
* [x] I have tested my changes ([see how](https://github.com/canonical/chisel-releases/blob/main/CONTRIBUTING.md#7-test-your-slices-before-opening-a-pr))
* [x] I have already submitted the [CLA form](
https://ubuntu.com/legal/contributors/agreement)

PR #1021, opened 2026-06-05
feat(24.04): add cron slice definitions

Soundarya03:
# Proposed changes

Add slice definition file for cron, containing daemon, crontab, config and copyright slices.

Context: I'm working on building a rock for an upstream project, and while traditional cron, even sliced, isn't ideal for containers, I would like to have a like-for-like translation of the upstream Dockerfile to a chiseled rock to begin with. 

### Forward porting
I'll be raising a PR for 25.10 and 26.04 with the same changes once this is approved. Would that be fine?

## Checklist

* [x] I have read the [contributing guidelines](
https://github.com/canonical/chisel-releases/blob/main/CONTRIBUTING.md)
* [x] I have tested my changes ([see how](https://github.com/canonical/chisel-releases/blob/main/CONTRIBUTING.md#7-test-your-slices-before-opening-a-pr))
* [x] I have already submitted the [CLA form](
https://ubuntu.com/legal/contributors/agreement)

PR #1048, opened 2026-07-06
fix(26.10): use 'latest' endpoint

lczyk:
# Proposed changes

fixes https://github.com/canonical/chisel-releases/issues/752

this was stalled for awhile since i used to think we'll have to thread the GITHUB_TOKEN into the lxd setup, which would bring security implications (not impossible, but much more fiddly). i've had another crack at this since it started to come up again in https://github.com/canonical/chisel-releases/pull/1021, and found a solution. turns out we can grab the latest tag "for free" from the redirect of https://github.com/canonical/chisel/releases/latest. the reason we need a version tag first, and can't just grab the artefact is because the tarball name includes a version in its filename and we can only download from https://github.com/canonical/chisel/releases/latest if we already know the asset name (and we don't, since we don't at the time know what's `latest`)

## Related issues/PRs

- https://github.com/canonical/chisel-releases/issues/752
- tested here: https://github.com/canonical/chisel-releases/pull/1056

### Forward porting

- https://github.com/canonical/chisel-releases/pull/1048 **(this PR)**
- https://github.com/canonical/chisel-releases/pull/1038
- https://github.com/canonical/chisel-releases/pull/1039
- https://github.com/canonical/chisel-releases/pull/1040
- https://github.com/canonical/chisel-releases/pull/1041
- https://github.com/canonical/chisel-releases/pull/1042

## Checklist

## Additional information

proof:

```bash
$ curl -fsS -D - https://github.com/canonical/chisel/releases/latest | sed -n '/content-security-policy:/d;/set-cookie:/d;p'
HTTP/2 302
date: Sun, 05 Jul 2026 15:08:14 GMT
content-type: text/html; charset=utf-8
vary: X-PJAX, X-PJAX-Container, Turbo-Visit, Turbo-Frame, X-Requested-With, Sec-Fetch-Site,Accept-Encoding, Accept, X-Requested-With
location: https://github.com/canonical/chisel/releases/tag/v1.4.1
cache-control: no-cache
strict-transport-security: max-age=31536000; includeSubdomains; preload
x-frame-options: deny
x-content-type-options: nosniff
x-xss-protection: 0
referrer-policy: no-referrer-when-downgrade
server: github.com
content-length: 0
x-github-request-id: FFCD:1A2C2D:F66BC5:129F58D:6A4A7381
```

no `x-ratelimit-limit` headers and tag in `location:` (i've dropped `two v long fields with sed). in contrast:

```bash
$ curl -fsS -D - -o /dev/null https://api.github.com/repos/canonical/chisel/releases/latest
HTTP/2 200
date: Sun, 05 Jul 2026 15:12:34 GMT
content-type: application/json; charset=utf-8
...
server: github.com
accept-ranges: bytes
x-ratelimit-limit: 60
x-ratelimit-remaining: 54
x-ratelimit-used: 6
x-ratelimit-resource: core
x-ratelimit-reset: 1783267230
content-length: 25391
x-github-request-id: C02F:384F98:C7C5DA:F1D1BF:6A4A7462
```

### DIFF
--- a/spread.yaml
+++ b/spread.yaml
@@ -128,14 +128,26 @@ prepare: |
   arch="$(dpkg --print-architecture | sed -e 's/armhf/arm/g' -e 's/ppc64el/ppc64le/g')"
 
   apt install -y curl wget
-  curl -s https://api.github.com/repos/canonical/chisel/releases/latest \
-    | awk "/browser_download_url/ && /chisel_v/ && /_$arch\./" \
-    | cut -d : -f 2,3 \
-    | tr -d \" \
-    | xargs wget
 
-  sha384sum -c chisel_v*sha384
-  tar -xf chisel_v*tar.gz -C /usr/local/bin
+  # Resolve the latest chisel release without calling api.github.com, whose
+  # unauthenticated REST quota (60 req/hr, shared per source IP) the CI runners
+  # routinely exhaust. An exhausted quota returns no asset URL, and the old
+  # `curl ... | xargs wget` pipeline then died with a cryptic `wget: missing URL`.
+  # The github.com web endpoint 302-redirects to the tagged release and is not
+  # subject to that quota, so we read the tag from the redirect target. See #752.
+  tag="$(curl -fsS -o /dev/null -w '%{redirect_url}' \
+    https://github.com/canonical/chisel/releases/latest | sed 's#.*/tag/##')"
+  if [ -z "$tag" ]; then
+    echo "FATAL: could not resolve latest chisel release tag (github.com unreachable?)" >&2
+    exit 1
+  fi
+
+  base="https://github.com/canonical/chisel/releases/download/$tag"
+  tarball="chisel_${tag}_linux_${arch}.tar.gz"
+  wget "$base/$tarball" "$base/$tarball.sha384"
+
+  sha384sum -c "$tarball.sha384"
+  tar -xf "$tarball" -C /usr/local/bin
 
 prepare-each: chisel version
 


spread.yaml
```diff
@@ -128,14 +128,26 @@ prepare: |
...
-  tar -xf chisel_v*tar.gz -C /usr/local/bin
+
+  # Resolve the latest chisel release without calling api.github.com, whose
+  # unauthenticated REST quota (60 req/hr, shared per source IP) the CI runners
+  # routinely exhaust. An exhausted quota returns no asset URL, and the old
+  # `curl ... | xargs wget` pipeline then died with a cryptic `wget: missing URL`.
+  # The github.com web endpoint 302-redirects to the tagged release and is not
+  # subject to that quota, so we read the tag from the redirect target. See #752.
+  tag="$(curl -fsS -o /dev/null -w '%{redirect_url}' \
+    https://github.com/canonical/chisel/releases/latest | sed 's#.*/tag/##')"
```
alesancor1, line 139: As much as I'd like to get this fixed I'm concerned by the reliability of this... GitHub tends to change their pages without prior notice.

While I don't think that the release page is going to change anytime soon, this still feels like a step back in reliability compared to using the API.

I'm sure there should be another approach, would we run into rate limits with GraphQL? Is it worth it to use a token maybe?

lczyk: this does not parse the page but gets the info from 302 redirect header of `/releases/latest` (documented [here](https://docs.github.com/en/repositories/releasing-projects-on-github/linking-to-releases) and goes back at least to 2021 in [the wayback machine](https://web.archive.org/web/20211103015500/https://docs.github.com/en/repositories/releasing-projects-on-github/linking-to-releases). edit: [2015](http://web.archive.org/web/20150405022905/https://help.github.com/articles/linking-to-releases/) with some digging) note though that this does not prove that 1) it will not change tomorrow and 2) that it always was a 302 redirect, so you do have a point.

re graphql/other options: they all need a token and sure, we can thread that into the LXD prepare, but it feels like more code for not much gain + it exposed the token to tests. sure we would scope the token etc etc, but it feels excessive.

3rd option would be to resolve chisel version in test host (where we have GITHUB_TOKEN) and then just pass it into the spread. the thing i dont like about this is that it breaks just running `spread` on the clone. it woudl need to be `env CHISEL_VERSION=v9.9.9 spread` instead, which i guess is a feature, but then i would expect just `spread` (w/out setting the env) to default to `latest` and we're back to the same thing.

alesancor1: After 1:1 discussion - Either the current approach or using the authenticated API with the read-only GITHUB_TOKEN

alesancor1 (approved):
Approving this, though I'd like a second review.

After some discussion, I don't have a strong opinion on whether keep the current approach or use the authenticated API with the read-only token.

ping @canonical/rocks

zhijie-yang (approved):

PR #1052, opened 2026-07-07
feat(26.10): add tomcat11 slices

vpa1977:
# Proposed changes
Tomcat11 package layout is the same as tomcat10.
Copy the tomcat10 slices.

Added refactoring to improve slice naming:
 - scripts renamed to catalina
 - setclasspath split into separate slice
 - extra-scripts renamed to scripts

## Related issues/PRs

https://github.com/canonical/chisel-releases/pull/1055

## Checklist

* [ ] I have read the [contributing guidelines](
https://github.com/canonical/chisel-releases/blob/main/CONTRIBUTING.md)
* [ ] I have tested my changes ([see how](https://github.com/canonical/chisel-releases/blob/main/CONTRIBUTING.md#7-test-your-slices-before-opening-a-pr))
* [ ] I have already submitted the [CLA form](
https://ubuntu.com/legal/contributors/agreement)

PR #1055, opened 2026-07-07
feat(26.04): add tomcat11 slices

vpa1977:
Tomcat11 package layout is the same as tomcat10.

Copy the tomcat10 slices.

# Proposed changes
Tomcat11 package layout is the same as tomcat10.
Copy the tomcat10 slices.

Added refactoring to improve slice naming:
 - scripts renamed to catalina
 - setclasspath split into separate slice
 - extra-scripts renamed to scripts

## Related issues/PRs

Migrate tomcat10 in 26.10 to the new slice structure https://github.com/canonical/chisel-releases/pull/1141

### Forward porting
https://github.com/canonical/chisel-releases/pull/1052

## Checklist

### DIFF
--- a/slices/tomcat11-common.yaml
+++ b/slices/tomcat11-common.yaml
@@ -1,0 +1,108 @@
+package: tomcat11-common
+
+essential:
+  tomcat11-common_copyright:
+
+# NOTE: all the jar files in are symlinks
+# to their corresponding jars from libtomcat11-java except
+# /usr/share/tomcat11/bin/bootstrap.jar
+
+slices:
+  scripts:
+    essential:
+      coreutils_delaying:
+      coreutils_directory-listing:
+      coreutils_dirname:
+      coreutils_echo:
+      coreutils_expr:
+      coreutils_output-of-entire-files:
+      coreutils_rm-bin:
+      coreutils_special-file-types:
+      coreutils_system-context:
+      coreutils_touch:
+      coreutils_working-context:
+      dash_bins:
+      debianutils_which:
+      grep_bins:
+      procps_bins:
+      tomcat11-common_jars:
+    contents:
+      /usr/share/tomcat11/bin/catalina.sh:
+      /usr/share/tomcat11/bin/setclasspath.sh:
+
+  jars:
+    essential:
+      libtomcat11-java_jars:
+      openjdk-21-jre-headless_jfr:
+      openjdk-21-jre-headless_management:
+      openjdk-21-jre-headless_security:
+    contents:
+      /usr/share/tomcat11/bin/bootstrap.jar:
+      /usr/share/tomcat11/bin/tomcat-juli.jar:
+      /usr/share/tomcat11/lib/annotations-api.jar:
+      /usr/share/tomcat11/lib/catalina-ha.jar:
+      /usr/share/tomcat11/lib/catalina-storeconfig.jar:
+      /usr/share/tomcat11/lib/catalina-tribes.jar:
+      /usr/share/tomcat11/lib/catalina.jar:
+      /usr/share/tomcat11/lib/el-api.jar:
+      /usr/share/tomcat11/lib/jasper-el.jar:
+      /usr/share/tomcat11/lib/jasper.jar:
+      /usr/share/tomcat11/lib/jaspic-api.jar:
+      /usr/share/tomcat11/lib/jsp-api.jar:
+      /usr/share/tomcat11/lib/servlet-api.jar:
+      /usr/share/tomcat11/lib/tomcat-api.jar:
+      /usr/share/tomcat11/lib/tomcat-coyote.jar:
+      /usr/share/tomcat11/lib/tomcat-dbcp.jar:
+      /usr/share/tomcat11/lib/tomcat-jdbc.jar:
+      /usr/share/tomcat11/lib/tomcat-jni.jar:
+      /usr/share/tomcat11/lib/tomcat-util-scan.jar:
+      /usr/share/tomcat11/lib/tomcat-util.jar:
+      /usr/share/tomcat11/lib/tomcat-websocket.jar:
+      /usr/share/tomcat11/lib/websocket-api.jar:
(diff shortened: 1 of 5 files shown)


rroessler (commented):
Looks good to me!

alesancor1 (approved):
LGTM, thanks!

slices/tomcat11-common.yaml
```diff
@@ -0,0 +1,108 @@
...
+      libtomcat11-java_i18n:
+      tomcat11-common_jars:
+    contents:
+      /usr/share/tomcat11/lib/tomcat-i18n-es.jar:
+      /usr/share/tomcat11/lib/tomcat-i18n-fr.jar:
+      /usr/share/tomcat11/lib/tomcat-i18n-ja.jar:
+      /usr/share/tomcat11/lib/tomcat-i18n-ru.jar:
+
+  # additional shell scripts
+  extra-scripts:
```
cjdcordeiro: what was the rational for calling it "extra-scripts"? We typically have 2 schools of though - by file type, or by functionality. This conveys neither cause "extra" introduces vagueness (_what goes into extra vs what stays out?_)

vpa1977: Those are self test, configuration queries (ciphers, version) and service startup/shutdown scripts and new catalina-base creation. 

We can rename this one to `utility` if it better conveys its purpose.

cjdcordeiro: seems like a mixed pot of things that are, in some cases, uncorrelated. 

It doesn't help that `scripts` isn't actually all the scripts. Which is typically the case: the `<type>s` slice puts together all paths of the same type (in this case, would've been ALL scripts), and then there may be sub slices for compartmentalizing said scripts by scope.

IoW, I'd be less worried if `scripts` was everything, and then you had something like `core-scripts` (which is actually what you are using `scripts` for right now -> for the `core` slice)

rroessler (approved):

slices/tomcat11-common.yaml
vpa1977: So just to confirm: `scripts` are renamed to `core-scripts` and `extra-scripts` become `scripts` in tomcat11?

It will introduce a bit of mismatch vs tomcat10 that already has `scripts` and `extra-scripts`.

cjdcordeiro: asking for @lczyk 's input on this.

TLDR: given tomcat10 already exists, I think we may need a migration strategy. What about:

```yaml

slices:
  core:
    essential:
      tomcat11-common_jars:              # Redundant
      tomcat11-common_scripts:

  scripts:
     <keep current>

  extra-scripts:
     <keep current>
```

Then for 26.10, let's have:

```yaml

slices:
  core:
    essential:
       tomcat11-common_jars:              # Redundant
    contents:
       <the current paths from `scripts`>

  scripts:
     <current scripts + extra scripts>
```

Wdyt?

vpa1977: We can drop extra-scripts slice from tomcat10 in 26.10.  Something like this: https://github.com/canonical/chisel-releases/pull/1141/changes

vpa1977: Ah, I remember why we did not [do core slice in tomcat10-common](https://github.com/canonical/chisel-releases/pull/637#discussion_r2394003997) - the reasoning was is that core slice is supposed to provide some functionality on its own, but it is impossible with tomcat10-common package because all the required configuration is place in tomcat10 package. 

let's wait for @lczyk opinion then before making PRs =)

vpa1977: As discussed offline with @cjdcordeiro , I've split out setclasspath slice, renamed scripts -> catalina and extra-scripts -> scripts.

cjdcordeiro: wdyt @lczyk ?

Looks like a good split.

Given that `tomcat10` already exists, I wonder if 26.04 should preserve the old slices, and this new pattern should only be applied to 26.10 onward.

We can also have a mix here, in 26.04:

 - keep the catalina and setclasspath
 - keep the 26.04 tomcat10-like extra-scripts
 - keep the 26.04 tomcat10-like scripts

cjdcordeiro (commented):
⬆️

slices/tomcat11-common.yaml
vpa1977: Updated slices

slices/tomcat11.yaml
```diff
@@ -0,0 +1,47 @@
...
+      /var/log/tomcat11/:
+
+  standard:
+    essential:
+      tomcat11-common_i18n:
+      tomcat11_core:
+
+  dev:
+    essential:
+      tomcat11-common_scripts:
```
lczyk:
```suggestion
      tomcat11-common_extra-scripts:
```

following [this](https://github.com/canonical/chisel-releases/pull/1055/changes#r3713284204) discussion, i'm guessing this should be extra-scripts ?? ( otherwise dev is the same as standard, which already has scripts through standard -> tomcat11_core -> tomcat11-common_catalina, which itself essentials _setclasspath and _jars. tomcat11-common_scripts here would add  nothing standard doesn't already have.

lczyk (changes requested):
found a couple of small and obvious typos so instead of fiddling with review comments i've pushed them. pls have a look and let me know if anything is not correct. other than that it looks good. i'll leave it as 'req changes' though for the one comment i've made since its more of a slice-design decision.

vpa1977:
> found a couple of small and obvious typos so instead of fiddling with review comments i've pushed them. pls have a look and let me know if anything is not correct. other than that it looks good. i'll leave it as 'req changes' though for the one comment i've made since its more of a slice-design decision.

Thank you!!!! 

I've also added PRs to fix the same copyright issue in tomcat10 
- https://github.com/canonical/chisel-releases/pull/1183
- https://github.com/canonical/chisel-releases/pull/1184

lczyk (approved):

PR #1069, opened 2026-07-13
test(24.04): add bash coverage

Samielakkad:
Part of the Bash slice coverage work for #1010.

This PR applies the integration test to `ubuntu-24.04`. The project maintains each Ubuntu release on a separate branch, so the same change needs three linked PRs:

- #1072 — Ubuntu 20.04
- #1070 — Ubuntu 22.04
- #1069 — Ubuntu 24.04

The test installs the `bash_bins` slice in a clean root filesystem, runs Bash inside the chroot, and checks that Bash also starts with an empty environment.

Checked locally with `bash -n` and a YAML parse. The full Spread suite needs Linux with LXD and was not run locally. The Canonical contributor agreement is signed.

Issue #1010, opened 2026-05-21
feat(bash): add tests


lczyk:

Samielakkad:
I opened the maintained-release test chain:

- #1072 for ubuntu-20.04
- #1070 for ubuntu-22.04
- #1069 for ubuntu-24.04
- #1071 for ubuntu-25.10

Each PR installs bash_bins, verifies Bash can execute inside the chroot, and checks BASH_VERSION with an empty environment. ubuntu-26.04 and ubuntu-26.10 already contain equivalent coverage.

PR #1072, opened 2026-07-13
test(20.04): add bash coverage

Samielakkad:
Part of the Bash slice coverage work for #1010.

This PR applies the integration test to `ubuntu-20.04`. The project maintains each Ubuntu release on a separate branch, so the same change needs three linked PRs:

- #1072 — Ubuntu 20.04
- #1070 — Ubuntu 22.04
- #1069 — Ubuntu 24.04

The test installs the `bash_bins` slice in a clean root filesystem, runs Bash inside the chroot, and checks that Bash also starts with an empty environment.

Checked locally with `bash -n` and a YAML parse. The full Spread suite needs Linux with LXD and was not run locally. The Canonical contributor agreement is signed.


Samielakkad:
The current 20.04 integration failure looks like runner infrastructure rather than the test patch: the X64 job stops at Cannot allocate lxd:ubuntu-focal: System is degraded. I do not have permission to rerun the workflow from this side, but the branch still contains the requested direct chroot invocation.

Samielakkad:
I have now signed the Canonical Contributor License Agreement. I pushed a no-file-change commit solely to trigger a fresh CI/CLA evaluation; GitHub created a new CI run for this PR, but it is currently action_required with no jobs, which appears to require repository-side approval for the fork workflow. Could a maintainer approve the run or re-run the CLA check when convenient?

PR #1103, opened 2026-07-29
feat: bin slice builder

upils:
# Proposed changes

Bin packages (.tar.xz from `bincraft pack`) are not Debian debs, so the existing deb-slicing workflow cannot author their Chisel slice definitions, and sd-tools produces them in bulk — hand-authoring does not scale. This adds `slice-builder`, a uv-managed Python CLI that generates one SDF per bin archive by wrapping an `omp` agent guided by an `sdf-generation` skill.

The split of responsibilities is the crux: the agent only makes judgment calls (path→slice classification, picking dependency slices for `essential`), while the builder owns everything deterministic — path-traversal-safe extraction, byte-order sorting, literal-path `prefer` conflict detection, and a custom PyYAML dumper that reproduces the repo's SDF style. Consequently validation is two-pass: the agent is retried only on semantic failures, since style and ordering are re-imposed by `render()` and retrying for them would waste LLM round-trips.

Two things worth knowing for review: the agent is handed the *extracted* archive directory rather than a path list, because file type and content are often needed to classify correctly (a `/usr/lib/` entry may be a shared library, a static archive, or data); and the whole tool is a self-contained `slice-builder/` subtree so it can be lifted to its own repository without rework. Non-determinism is accepted for now — the path table and dep-slice selection are expected to evolve via skill edits rather than code changes.

## Checklist

### DIFF
--- a/slice-builder/README.md
+++ b/slice-builder/README.md
@@ -1,7 +1,7 @@
 # slice-builder
 
 `slice-builder` generates [Chisel](https://canonical-chisel--72.com.readthedocs.build/72/)
-slice definition files (SDFs) for **superdistro bin** packages (`.tar.xz` archives produced by
+slice definition files (SDFs) for **bin** packages (`.tar.xz` archives produced by
 `bincraft pack`). Bin packages are not Debian debs, so the standard deb-slicing workflow does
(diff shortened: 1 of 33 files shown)


upils:
Closing in favor of https://github.com/canonical/sd-tools/pull/3924

PR #1104, opened 2026-07-30
feat(26.04): add rsyslog and rsyslog-relp slices

jy5275:
# Proposed changes

Add slice definitions for `rsyslog`, `rsyslog-relp`, and their dependency chain on `ubuntu-26.04`.

### Packages added

| Package | Slices | Notes |
|---------|--------|-------|
| `libestr0` | `libs`, `copyright` | Leaf library; rsyslog dependency |
| `libfastjson4` | `libs`, `copyright` | Leaf library; rsyslog dependency |
| `librelp0` | `libs`, `copyright` | RELP protocol library; rsyslog-relp dependency |
| `rsyslog` | `bins`, `modules`, `config`, `data`, `copyright` | Daemon + 37 loadable dlopen modules |
| `rsyslog-relp` | `modules`, `copyright` | `imrelp.so` / `omrelp.so` RELP modules |

### Forward porting
#1105

## Checklist
* [x] I have read the [contributing guidelines](https://github.com/canonical/chisel-releases/blob/main/CONTRIBUTING.md)
* [x] I have already submitted the [CLA form](https://ubuntu.com/legal/agreement)

### DIFF
--- a/slices/rsyslog-relp.yaml
+++ b/slices/rsyslog-relp.yaml
@@ -1,0 +1,17 @@
+package: rsyslog-relp
+
+essential:
+  rsyslog-relp_copyright:
+
+slices:
+  modules:
+    essential:
+      libc6_libs:
+      librelp0_libs:
+    contents:
+      /usr/lib/*-linux-*/rsyslog/imrelp.so:
+      /usr/lib/*-linux-*/rsyslog/omrelp.so:
+
+  copyright:
+    contents:
+      /usr/share/doc/rsyslog-relp/copyright:
--- a/slices/rsyslog.yaml
+++ b/slices/rsyslog.yaml
@@ -1,0 +1,84 @@
+package: rsyslog
+
+essential:
+  rsyslog_copyright:
+
+slices:
+  bins:
+    essential:
+      libc6_libs:
+      libestr0_libs:
+      libfastjson4_libs:
+      libsystemd0_libs:
+      libuuid1_libs:
+      rsyslog_config:
+      rsyslog_modules:
+      zlib1g_libs:
+    contents:
+      /usr/sbin/rsyslogd:
+
+  data:
+    contents:
+      /var/spool/rsyslog/:
+
+  config:
+    essential:
+      rsyslog_modules:
+    contents:
+      /etc/logcheck/ignore.d.server/rsyslog:
+      /etc/logrotate.d/rsyslog:
+      /etc/rsyslog.conf:
+      /etc/rsyslog.d/50-default.conf:
+        copy: /usr/share/rsyslog/50-default.conf
+      /etc/systemd/system/multi-user.target.wants/rsyslog.service:
+        symlink: /usr/lib/systemd/system/rsyslog.service
+
+  modules:
+    essential:
+      libc6_libs:
+      libfastjson4_libs:
+      libsystemd0_libs:
+      libzstd1_libs:
+      zlib1g_libs:
+    contents:
+      /usr/lib/*-linux-*/rsyslog/fmhash.so:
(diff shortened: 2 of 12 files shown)


slices/rsyslog-relp.yaml
```diff
@@ -0,0 +1,18 @@
+package: rsyslog-relp
+
+essential:
+  rsyslog-relp_copyright:
+
+slices:
+  modules:
```
cjdcordeiro, line 7: why `modules` and not `libs`?

cjdcordeiro: ok, I understand this is then possibly an extension of `rsyslog_modules`. Does it need those as an `essential` then?

slices/rsyslog.yaml
```diff
@@ -0,0 +1,82 @@
+package: rsyslog
+
+essential:
+  rsyslog_copyright:
+
+slices:
+  bins:
```
cjdcordeiro, line 7: `bins` don't depend on `modules`, and `modules` don't depend on `bins`. Is that normal?

It is at least inconsistent with `rsyslog-relp_modules`, which depends on `rsyslog_bins`...

tests/spread/integration/rsyslog-relp/task.yaml
cjdcordeiro, line 1: is it possible to do a more functional test than this? After all the entire `rsyslog_bins` will be present, with two additional modules (`relp`) that should add the new functionality we wanna test

cjdcordeiro (changes requested):
in the lib slices you opted for explicitly naming the paths instead of something like `...libfoo.0*`. But that's ok. Just noting that for future reference it would be ok to use the glob in these cases.

jy5275:
> in the lib slices you opted for explicitly naming the paths instead of something like `...libfoo.0*`. But that's ok. Just noting that for future reference it would be ok to use the glob in these cases.

The reason for explicitly naming all paths (instead of `/usr/lib/*-linux-*/rsyslog/*.so:`) is that both rsyslog_modules and rsyslog-relp_modules have files that match this glob. I want to separate them explicitly.

slices/rsyslog-relp.yaml
jy5275: These `.so` files are dlopen-loaded plugins (`module(load="imrelp")`), not link-time shared libraries — they're not in the ldconfig index and aren't linked by any binary. So I think `modules` matches the convention used by `libnginx-mod-*`, `php8.5-cli`, and `apache2` in this repo. But if you think `libs` are more proper, I'll do the renaming.

jy5275: The RELP modules (imrelp.so, omrelp.so) are self-contained. They don't require any of the 37 rsyslog modules to function. A user may want a setup where `rsyslogd` loads only `imrelp`/`omrelp` for RELP forwarding without other modules. So `rsyslog-relp_modules` should not depend on `rsyslog_modules`.

slices/rsyslog.yaml
jy5275: Thanks for catching this. I've updated the dependency relationships. Now:
- `rsyslog_bins` depends on `rsyslog_modules` and `rsyslog_config` (consistent with 24.04)
- `rsyslog-relp_modules` doesn't depend on `rsyslog_bins`

So installing bins gives a working daemon out of the box (daemon + core modules + default config)

tests/spread/integration/rsyslog-relp/task.yaml
jy5275: Have added the functional test: start rsyslogd with a config that loads imrelp, send a log message over the RELP protocol, and verify it is received.

PR #1109, opened 2026-07-30
feat(26.04): add rsyslog-relp and dependency slices

jy5275:
## Summary

Add slice definitions for `rsyslog-relp` and its full dependency chain on `ubuntu-26.04`.

### Packages added (bottom-up)

| Package | Slices | Notes |
|---------|--------|-------|
| `libestr0` | `libs`, `copyright` | Leaf library; consumed by rsyslog |
| `libfastjson4` | `libs`, `copyright` | Leaf library; consumed by rsyslog |
| `librelp0` | `libs`, `copyright` | RELP protocol library; consumed by rsyslog-relp |
| `rsyslog` | `bins`, `modules`, `config`, `var`, `copyright` | Daemon + 37 loadable modules; `var` slice uses a `mutate` script to create the `syslog` user/group (normally done by postinst via `adduser`) |
| `rsyslog-relp` | `modules`, `copyright` | `imrelp.so` / `omrelp.so` RELP modules |

### Design decisions

- **`adduser` / `ucf` not sliced**: they are install-time tools (postinst), not runtime dependencies — `rsyslogd` does not link or call them at runtime. The `syslog` user is instead created via a `mutate:` script in `rsyslog_var` (depends on `base-passwd_data`, which runs its mutate first per chisel's dependency-ordered mutation).
- **rsyslog modules listed explicitly** (not globbed) to avoid a cross-package path-conflict with `rsyslog-relp_modules` in the same `/usr/lib/*-linux-*/rsyslog/` directory.
- **Config reproduced from postinst**: `/etc/rsyslog.d/50-default.conf` is created via `copy:` from the deb's `/usr/share/rsyslog/50-default.conf` (postinst normally uses `ucf`).
- **Dropped auxiliary configs**: logrotate/logcheck configs dropped (those tools aren't sliced); apparmor/services/scripts/tmpfiles omitted for a minimal, cron-style slice set — can be added in follow-ups.
- **Format v3**: all `essential:` entries use the map form.

### Verification

- `check-slice.py`: all 5 SDFs clean (format v3)
- `check-diff.py`: no append-only regressions
- `check-test.py`: rsyslog binaries all exercised; library/module packages have no `bins` (info only)
- `chisel cut` (Layer 1 installability): all 8 slices cut successfully
- Manual functional tests: `rsyslogd -v` (8.2512.0), `rsyslogd -N1` config validation passes, all 39 `.so` modules valid ELF, `syslog` user/group created by mutate, `/var/spool/rsyslog` created (mode 0700)
- Spread tests written for all 5 packages (not run locally — `spread` not installed in this environment; CI will execute them)

### Forward-port note

Net-new packages — no existing SDFs on other release branches to forward-port from. If/when these land on 26.04, they can be forward-ported to future release branches.

### DIFF
--- a/slices/libestr0.yaml
+++ b/slices/libestr0.yaml
@@ -1,0 +1,15 @@
+package: libestr0
+
+essential:
+  libestr0_copyright:
+
+slices:
+  libs:
+    essential:
+      libc6_libs:
+    contents:
+      /usr/lib/*-linux-*/libestr.so.0*:
+
+  copyright:
+    contents:
+      /usr/share/doc/libestr0/copyright:
--- a/slices/libfastjson4.yaml
+++ b/slices/libfastjson4.yaml
@@ -1,0 +1,15 @@
+package: libfastjson4
+
+essential:
+  libfastjson4_copyright:
+
+slices:
+  libs:
+    essential:
+      libc6_libs:
+    contents:
+      /usr/lib/*-linux-*/libfastjson.so.4*:
+
+  copyright:
+    contents:
+      /usr/share/doc/libfastjson4/copyright:
--- a/slices/librelp0.yaml
+++ b/slices/librelp0.yaml
@@ -1,0 +1,17 @@
+package: librelp0
+
+essential:
+  librelp0_copyright:
+
+slices:
+  libs:
+    essential:
+      libc6_libs:
+      libgnutls30t64_libs:
+      libssl3t64_libs:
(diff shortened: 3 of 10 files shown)


jy5275:
Closing in favor of #1104, which already covers rsyslog + libestr0 + libfastjson4. The two extra packages (librelp0, rsyslog-relp) have been added to #1104 instead.

PR #857, opened 2026-01-26
fix(26.04): update util-linux slices

alesancor1:
# Proposed changes
Update `util-linux` slices according to upstream changes:

* Some binaries have been moved to `util-linux-extra` and `bsdextrautils`.
  * Slices for them are included

## Checklist

* [x] I have read the [contributing guidelines](
https://github.com/canonical/chisel-releases/blob/main/CONTRIBUTING.md)
* [x] I have tested my changes ([see how](https://github.com/canonical/chisel-releases/blob/main/CONTRIBUTING.md#7-test-your-slices-before-opening-a-pr))
* [x] I have already submitted the [CLA form](
https://ubuntu.com/legal/contributors/agreement)

PR #662, opened 2025-09-19
fix(25.10): updates to util-linux for 2.41

lengau:
# Proposed changes
Based on: https://changelogs.ubuntu.com/changelogs/pool/main/u/util-linux/util-linux_2.41-4ubuntu4/changelog

- resizepart was moved to util-linux-extra in 2.40.4-4
- nsenter requires more arguments now (test fixed, no slice changes)
- lslogins now requires liblastlog2

## Checklist

* [x] I have read the [contributing guidelines](
https://github.com/canonical/chisel-releases/blob/main/CONTRIBUTING.md)
* [x] I have tested my changes ([see how](https://github.com/canonical/chisel-releases/blob/main/CONTRIBUTING.md#7-test-your-slices-before-opening-a-pr))
* [x] I have already submitted the [CLA form](
https://ubuntu.com/legal/contributors/agreement)

PR #1147, opened 2026-08-21
fix(26.04): util-linux no longer needs liblastlog

cjdcordeiro:
# Proposed changes

Removed unneeded dependency from `util-linux_login`.

**IMPORTANT NOTE**: this is an interesting case where the opening of a new release carried a change that cause orphaned dependencies. In https://github.com/canonical/chisel-releases/pull/662, @lengau introduced a missing dependency to `util-linux`, that was needed for the `lslogins` binary. 

In 26.04, that binary is no longer in the pkg and consequently no longer in the `util-linux_login` slice. **However**, the `liblastlog2-2` is left over from https://github.com/canonical/chisel-releases/pull/662, and is getting unnecessarily installed. That's not good, cause it goes against what we try to do with Chisel -> _remove everything that is unnecessary_

I noticed this because `liblastlog2-2` brings in sqlite3 libs too, which completely caught me off guard. 

So what's the takeaway? We're doing a good job at detecting and testing dependencies that are missing, and we don't do enough checks to ensure that we are not shipping unnecessary. `lslogins` was removed in https://github.com/canonical/chisel-releases/pull/857/changes#diff-19d63239e73ac43045433aa7c62758e215be4782ad63accaec2435ddaf8254d0L156. The CI that list the dependency diff did NOT run in https://github.com/canonical/chisel-releases/pull/857, unfortunately. That would have (hopefully) said that there was an unnecessary dependency. BUT, even with that, it would have just been an informative comment, while IMO this should've been a CI error. @lczyk can you pls ACK, weigh in, and if agreeing, create a candidate roadmap item for us to tackle this in future releases?

## Related issues/PRs

https://github.com/canonical/chisel-releases/pull/662

### Forward porting

https://github.com/canonical/chisel-releases/pull/1146

## Checklist

### DIFF
--- a/slices/util-linux.yaml
+++ b/slices/util-linux.yaml
@@ -147,8 +147,6 @@ slices:
       libc6_libs:
       libcrypt1_libs:
       libselinux1_libs:
-      libsmartcols1_libs:
-      libsystemd0_libs:
     contents:
       /usr/sbin/sulogin:
 
--- a/slices/util-linux.yaml
+++ b/slices/util-linux.yaml
@@ -146,7 +146,6 @@ slices:
     essential:
       libc6_libs:
       libcrypt1_libs:
-      liblastlog2-2_libs:
       libselinux1_libs:
       libsmartcols1_libs:
       libsystemd0_libs:


lengau (approved):
Thanks!

lczyk:
@cjdcordeiro ack, and yeah, i agree with the diagnosis and with the testing gap. will have a ponder on what more to do about it and drop an item for the next cycle. in the meantime, found 2 more dpes which could be dropped for the same reason: https://github.com/canonical/chisel-releases/pull/1147/commits/ccef8f08ba571c3a4bce34a446603fdab39d83bc. we do have a [`util-linux:login`](https://github.com/canonical/chisel-releases/blob/ubuntu-26.04/tests/spread/integration/util-linux/test_login.sh) test which tests just that binary and *does* [run here](https://github.com/canonical/chisel-releases/actions/runs/32641625650/job/97199478186?pr=1147#step:10:40) to check that this is another correct exclusion.

edit: tracked internally as ROCKS-3435

lczyk (approved):

PR #1159, opened 2026-08-23
fix(24.04): rootfs'es pile up across spread tasks

lczyk:
# Proposed changes

backport of #1156. partially addresses #1077.

`install-slices` test helper never removed the rootfs'es which it created which led to oom on some low-resources runners like ppc. this PR adds `clean-rootfs` helper which can be used both in the tests themselves and in spread's `restore-each` to clean up the runners between test runs.

the rabbitmq-server tests on this branch had a pre-existing bug where they were valid for amd64/arm64 only but there was no arch check. added in https://github.com/canonical/chisel-releases/pull/1159/commits/269cf5a3ab2005bcf0b2cd2b30aef6e21c919e4c

## Related issues/PRs

- https://github.com/canonical/chisel-releases/issues/1077
- https://github.com/canonical/chisel-releases/pull/1156

### Forward porting

- https://github.com/canonical/chisel-releases/pull/1158
- https://github.com/canonical/chisel-releases/pull/1156
- https://github.com/canonical/chisel-releases/pull/1159 **(this PR)**
- https://github.com/canonical/chisel-releases/pull/1160
- https://github.com/canonical/chisel-releases/pull/1161

## Checklist

* [x] I have read the [contributing guidelines](
https://github.com/canonical/chisel-releases/blob/main/CONTRIBUTING.md)
* [x] I have tested my changes ([see how](https://github.com/canonical/chisel-releases/blob/main/CONTRIBUTING.md#7-test-your-slices-before-opening-a-pr))
* [x] I have already submitted the [CLA form](
https://ubuntu.com/legal/contributors/agreement)

Issue #1077, opened 2026-07-14
ci: runners can run out of memory


lczyk:
spread tests on ppc64el runner fall over with `no space left on device` while cutting slices that pull in `rust-1.93-src_src` (full rustc source tree, ~GB per rootfs).

example: https://github.com/canonical/chisel-releases/actions/runs/29369888590/job/87210440346?pr=1034

```
2026/07/14 21:44:12 Extracting files from package "rust-1.93-src"...
error: cannot extract from package "rust-1.93-src": write /tmp/tmp.4KHdxHFn62/usr/src/rustc-1.93.1/src/tools/clippy/tests/ui/unnecessary_box_returns.rs: no space left on device
Failed to cut slices. See above for details.
```

`install-slices` does `mktemp -d` per call and never removes the rootfs. no `restore-each` in spread.yaml either, so rootfs'es (rootfs'i 🤔 ) pile up in `/tmp` of the lxd container across all tasks in the run. any test cutting big slices (rust-src, gcc-15, llvm, ...) will duffer from this issue, and on any runner (but mos likely ppc since its currently the smallest)

possible fixes, not mutually exclusive:

- clean up rootfses between tasks: suite-level `restore-each: rm -rf /tmp/*` in spread.yaml (cleanup inside `install-slices` obv doesn't work, although we could [defer](https://github.com/lczyk/defer.sh) it...

PR #1160, opened 2026-08-23
fix(22.04): rootfs'es pile up across spread tasks

lczyk:
# Proposed changes

backport of #1156. partially addresses #1077.

`install-slices` test helper never removed the rootfs'es which it created which led to oom on some low-resources runners like ppc. this PR adds `clean-rootfs` helper which can be used both in the tests themselves and in spread's `restore-each` to clean up the runners between test runs.

the `rabbitmq-server` tests on this branch had a pre-existing bug where they were valid for amd64/arm64 only but there was no arch check. added in https://github.com/canonical/chisel-releases/pull/1160/commits/f13e4ffa6b43177c31afc20a7c9cb1f1ae9e191e

## Related issues/PRs

- https://github.com/canonical/chisel-releases/issues/1077
- https://github.com/canonical/chisel-releases/pull/1156

### Forward porting

- https://github.com/canonical/chisel-releases/pull/1158
- https://github.com/canonical/chisel-releases/pull/1156
- https://github.com/canonical/chisel-releases/pull/1159
- https://github.com/canonical/chisel-releases/pull/1160 **(this PR)**
- https://github.com/canonical/chisel-releases/pull/1161

## Checklist

### DIFF
--- a/tests/spread/integration/rabbitmq-server/task.yaml
+++ b/tests/spread/integration/rabbitmq-server/task.yaml
@@ -23,4 +23,4 @@ execute: |
 
 restore: |
   # Kill the RabbitMQ server, which by default runs on port 25672
-  kill `sudo lsof -t -i :25672`
+  kill `sudo lsof -t -i :25672` || true
--- a/tests/spread/integration/rabbitmq-server/task.yaml
+++ b/tests/spread/integration/rabbitmq-server/task.yaml
@@ -1,15 +1,11 @@
 summary: Integration tests for RabbitMQ
 
 execute: |
-  # These tests only run for amd64 and arm64
-  arch="$(uname -m)"
-  case "$arch" in
-    x86_64|aarch64) ;;
-    *)
-      echo "Skipping: RabbitMQ server is supported only on amd64 and arm64"
-      exit 0
-      ;;
-  esac
+  # TODO: skipped on ppc64el, where the detached server never finishes booting
+  if [ "$(uname -m)" = "ppc64le" ]; then
+    echo "Skipping: RabbitMQ server does not come up on ppc64el"
+    exit 0
+  fi
 
   # For this test we also need libc-bin_getconf
   slices="rabbitmq-server_bins libc-bin_getconf"
--- a/tests/spread/integration/rabbitmq-server/task.yaml
+++ b/tests/spread/integration/rabbitmq-server/task.yaml
@@ -1,12 +1,6 @@
 summary: Integration tests for RabbitMQ
 
 execute: |
-  # TODO: skipped on ppc64el, where the detached server never finishes booting
-  if [ "$(uname -m)" = "ppc64le" ]; then
-    echo "Skipping: RabbitMQ server does not come up on ppc64el"
-    exit 0
-  fi
-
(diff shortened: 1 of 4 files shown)


tests/spread/integration/rabbitmq-server/task.yaml
```diff
@@ -2,6 +2,15 @@ summary: Integration tests for RabbitMQ
...
   # These tests only run for amd64 and arm64
+  arch="$(uname -m)"
+  case "$arch" in
+    x86_64|aarch64) ;;
+    *)
+      echo "Skipping: RabbitMQ server is supported only on amd64 and arm64"
+      exit 0
+      ;;
+  esac
+
```
zhijie-yang: The test works for the arches other than amd64 and arm64.

The latest attempt of the workflow at [2ec09f0](https://github.com/canonical/chisel-releases/pull/1160/commits/2ec09f052a6f3776d03d02e651f9a2171373e956) (which is before the gating of rabbitmq) works for s390x already, leaving only the ppc64el as the problematic architecture.

I'd assume it's a result of an unstable infrastructure and the test itself being fragil. The test should not be simply skipped.

zhijie-yang (changes requested):

tests/spread/integration/rabbitmq-server/task.yaml
lczyk: i was following the literal comment in that file saying _"# These tests only run for amd64 and arm64"_, but that's a fair point. i will expand these to work on s390x too

lczyk: https://github.com/canonical/chisel-releases/pull/1160/commits/fe9dbcab6dd9fcf2935e77ed203ca2738792b5b0

lczyk: at the end of the day had to skip both. resolving this and posting normal comment with more details...

lczyk:
ok, at the end of the day had to skip both ppc *and* s390x but for different reasons:

- on `ppc64el`: the node comes up, but not within the `sleep 10`. `rabbitmqctl status` returns `this command requires the 'rabbit' app to be running on the target node`, i.e. node is there, the `rabbit` app just hasn't finished booting.
- `s390x` has a different failure. the node does not register with [epmd](https://www.erlang.org/doc/apps/erts/epmd_cmd.html) at all: `epmd reports: node 'rabbit' not running at all`. `rabbitmq-server -detached` returns 0 and nothing appears. there's also this:

```
warning: Elixir is running in a system with a different endianness than the one its source code was compiled in ... the current one: big
```
i did try fixing it but `rabbitmqctl  wait_startup --timeout 120` didn't work because `await_startup`. leaving this here as work-in-progress breadcrumb trail for humans and agents alike. for now deeming this a deep rabbit hole ([🥁🥁🐍](https://www.youtube.com/watch?v=8eXj97stbG8)) which strays afar from the point of the PR.

zhijie-yang (approved):
thanks. it would be nice to create an issue to track the removal of this test. we want it get restored in the future whenever possible. otherwise, the completeness of the tests will be undermined and left forget.

PR #1161, opened 2026-08-23
fix(20.04): rootfs'es pile up across spread tasks

lczyk:
# Proposed changes

backport of #1156. partially addresses #1077.

`install-slices` test helper never removed the rootfs'es which it created which led to oom on some low-resources runners like ppc. this PR adds `clean-rootfs` helper which can be used both in the tests themselves and in spread's `restore-each` to clean up the runners between test runs.

there is no `rabbitmq-server` test on this branch, so the restore guard which #1156 also carries is not part of this one.

## Related issues/PRs

- https://github.com/canonical/chisel-releases/issues/1077
- https://github.com/canonical/chisel-releases/pull/1156

### Forward porting

- https://github.com/canonical/chisel-releases/pull/1158
- https://github.com/canonical/chisel-releases/pull/1156
- https://github.com/canonical/chisel-releases/pull/1159
- https://github.com/canonical/chisel-releases/pull/1160
- https://github.com/canonical/chisel-releases/pull/1161 **(this PR)**

## Checklist

### DIFF
--- a/tests/spread/lib/clean-rootfs
+++ b/tests/spread/lib/clean-rootfs
@@ -5,10 +5,12 @@
 # take up. Called between tasks by the restore-each hook in spread.yaml; tests
 # that cut several large rootfs'es can also call it on ones they are done with.
 # Usage: clean-rootfs [<rootfs>...]
-# With no arguments, every rootfs in the directory install-slices uses is removed.
+# With no arguments, every rootfs in $ROOTFS_DIR is removed.
 
-# Must match install-slices.
-_ROOTFS_DIR=/tmp/chisel-rootfs
+if [ -z "${ROOTFS_DIR:-}" ]; then
+    echo "Error: ROOTFS_DIR is not set" >&2
+    exit 1
+fi
 
 function main() {
     local rootfs
@@ -20,10 +22,10 @@ function main() {
         return
     fi
 
-    [ -d "$_ROOTFS_DIR" ] || return 0
+    [ -d "$ROOTFS_DIR" ] || return 0
     while IFS= read -r rootfs; do
         remove_rootfs "$rootfs"
-    done < <(find "$_ROOTFS_DIR" -mindepth 1 -maxdepth 1)
+    done < <(find "$ROOTFS_DIR" -mindepth 1 -maxdepth 1)
 }
 
 function remove_rootfs() {
@@ -31,9 +33,9 @@ function remove_rootfs() {
     local rootfs="${1%"${1##*[!/]}"}"
 
     # install-slices only ever creates direct children; anything else is not ours
-    case "${rootfs#"$_ROOTFS_DIR"/}" in
+    case "${rootfs#"$ROOTFS_DIR"/}" in
         "$rootfs" | "" | . | .. | */*)
-            echo "Error: refusing to remove '$1', which is not a rootfs in $_ROOTFS_DIR" >&2
+            echo "Error: refusing to remove '$1', which is not a rootfs in $ROOTFS_DIR" >&2
             exit 1
             ;;
     esac
--- a/tests/spread/lib/clean-rootfs
+++ b/tests/spread/lib/clean-rootfs
@@ -5,7 +5,8 @@
 # take up. Called between tasks by the restore-each hook in spread.yaml; tests
 # that cut several large rootfs'es can also call it on ones they are done with.
 # Usage: clean-rootfs [<rootfs>...]
-# With no arguments, every rootfs in $ROOTFS_DIR is removed.
+# Each argument has to be a path install-slices returned; anything else is
+# ignored. With no arguments, every rootfs in $ROOTFS_DIR is removed.
 
 if [ -z "${ROOTFS_DIR:-}" ]; then
     echo "Error: ROOTFS_DIR is not set" >&2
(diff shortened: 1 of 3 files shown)


tests/spread/lib/clean-rootfs
```diff
@@ -0,0 +1,71 @@
...
+# spellchecker: ignore rootfs
+
+# Removes chiselled rootfs'es created by install-slices, freeing the disk they
+# take up. Called between tasks by the restore-each hook in spread.yaml; tests
+# that cut several large rootfs'es can also call it on ones they are done with.
+# Usage: clean-rootfs [<rootfs>...]
+# With no arguments, every rootfs in the directory install-slices uses is removed.
+
+# Must match install-slices.
+_ROOTFS_DIR=/tmp/chisel-rootfs
```
cjdcordeiro: Thanks for this. I'm a bit worried here though.

This creates a split brain with the other `install-slices` script that assumes the same path. 

What about setting this env globally, in the spread project env?

tests/spread/lib/clean-rootfs
```diff
@@ -0,0 +1,71 @@
...
+
+    [ -d "$_ROOTFS_DIR" ] || return 0
+    while IFS= read -r rootfs; do
+        remove_rootfs "$rootfs"
+    done < <(find "$_ROOTFS_DIR" -mindepth 1 -maxdepth 1)
+}
+
+function remove_rootfs() {
+    # trailing slashes would read as a nested path to the guard below
+    local rootfs="${1%"${1##*[!/]}"}"
```
cjdcordeiro: hum, what is this doing that `"${1%%+(/)}"` can't do?

tests/spread/lib/clean-rootfs
```diff
@@ -0,0 +1,71 @@
...
+
+    # install-slices only ever creates direct children; anything else is not ours
+    case "${rootfs#"$_ROOTFS_DIR"/}" in
+        "$rootfs" | "" | . | .. | */*)
+            echo "Error: refusing to remove '$1', which is not a rootfs in $_ROOTFS_DIR" >&2
+            exit 1
+            ;;
+    esac
+    [ -e "$rootfs" ] || return 0
```
cjdcordeiro: I mean, some of these checks seems a bit overengineered. 

it's doing a lot of sanitization, as if this function was meant to be used on its own. If the following removal lines were placed above in line 25, instead of `remove_rootfs ...`, then you'd simply trust the `find` command I guess, which can return all directories, under _ROOTFS_DIR, with a given depth, and no trailing "/".

So, I guess it's a trade off between modularity and maintainability. I'm not gonna reject the PR based on this alone, but I wonder if we wouldn't gain more for just pursuing the simple route.

cjdcordeiro (changes requested):

tests/spread/lib/clean-rootfs
lczyk: yeah, `+(/)` is cleaner but it requires extglob which i didn't think of. mute after the `clean-rootfs` rework though. ( i'd not mind it either way tbh )

tests/spread/lib/clean-rootfs
lczyk: yep. np. that's a bit better, i agree.

tests/spread/lib/install-slices
```diff
@@ -5,14 +5,28 @@
...
     local slices="$*"
     # add base-files_chisel slice to ensure we generate the manifest
     slices="${slices} base-files_chisel"
 
-    local rootfs="$(mktemp -d)"
+    # chrooted daemons that drop privileges need the path to stay traversable
+    #shellcheck disable=SC2174 # the parent always exists, only the leaf is created
+    mkdir -p -m 755 "$ROOTFS_DIR"
+
+    local rootfs="$(mktemp -d -p "$ROOTFS_DIR")"
```
zhijie-yang: **SC2155: `mktemp` failure is masked by `local`.**

Because `local` is its own command, its exit status (always 0) replaces `mktemp`'s. If `mktemp` fails (e.g. `$ROOTFS_DIR` unwritable, disk full), `$rootfs` ends up empty and execution continues into `chisel_cut "$rootfs" ...`, which runs `chisel cut --root ""` against the current directory instead of failing fast. `set -e` won't catch it.

Since this line is already being modified, splitting declaration and assignment makes it fail loudly:

```sh
local rootfs
rootfs="$(mktemp -d -p "$ROOTFS_DIR")"
```

Pre-existing pattern and low severity, but this is the one place `--root` could silently become empty.

zhijie-yang (changes requested):
One change requested regarding `mktemp` error handling in `install-slices`. The rest of the change looks solid — the path guard, unmount-before-delete, and restore-hook non-failure handling in `clean-rootfs` are all correct.

tests/spread/lib/clean-rootfs
lczyk: > as if this function was meant to be used on its own

it is. its called by `restore-each` but it is designed to also allow test authors to call it in cases where we might want to clean things up mid-test, e.g. when cutting many maaaasive rootfses. that's the whole reason this fun is in lib, and not just a restore-each script.

> some of these checks seems a bit overengineered

fair though. i think `find` might be an overkill though. the simplest formulation i can see is "we allow removing only absolute paths to dirs under ROOTFS_DIR". simplified the whole thing to just that.

lczyk: https://github.com/canonical/chisel-releases/pull/1161/commits/4808d3d9c35934a11eeeb22cd62c1064c132edda

tests/spread/lib/clean-rootfs
lczyk: marking this as resolved since it's mute due to changes for the comment below

tests/spread/lib/clean-rootfs
lczyk: https://github.com/canonical/chisel-releases/pull/1161/commits/49a1a42a537fb9ce73ce74da27daf7d9f95346b6

tests/spread/lib/install-slices
lczyk: yep. that's a good catch and an easy driveby here. added in https://github.com/canonical/chisel-releases/pull/1161/commits/a5914da6a0a2be1914fd98a7a83a2cb941cd4cbe

cjdcordeiro (approved):
thanks for the changes. LGTM. (please wait for the CI to pass) (gonna wait for the CI before approving the FPs)

zhijie-yang (approved):

PR #1187, opened 2026-09-09
feat(ci): reject unknown keys

### LINKED ISSUE #1186: ci: reject unknown keys in slice definition files
Chisel ignores keys it doesn't recognise for forwards-compatibility reasons. This however means that a *misspelled* key is indistinguishable from a *future* key, and chisel-releases ci is the right place to try to catch it. We should add such a ci.

Specific example which just happened: Several `copyright` slices used `content:` instead of `contents:` and silently shipped no copyright. This is now fixed / in-flight:

- https://github.com/canonical/chisel-releases/pull/1141
- https://github.com/canonical/chisel-releases/pull/1183
- https://github.com/canonical/chisel-releases/pull/1185
- https://github.com/canonical/chisel-releases/pull/1184

lczyk:
# Proposed changes

this pr proposes a ci job which checks what chisel *by design* (for forward compatibility) does not -- whether the keys in slice definition files are only the ones used by chisel.

## Related issues/PRs

closes https://github.com/canonical/chisel-releases/issues/1186

### Forward porting

none yet, but if the PR is approved it will need a bunch of simple per-live-branch PRs adding this to ci.yaml's

- ...
- ...

## Checklist


cjdcordeiro (changes requested):
I'm not sure about this. We should put a pin on it and think it through carefully.

Reality is that this schema leniency is there by design, on the Chisel side, as you mention in the python script's docstring.

So failing the CI feels...awkward. Not to say the fact that right now the script packs a bunch of boilerplate schema keys, which is also not ideal - we're always gonna be playing catch (FYI there is a backlog item for the Chisel team to work on a schema file that can be sourced and maintained via the schemaStore).

AFAICT, the motivation for this PR stems from a bad `content` vs `contents`, and I would dare saying that this case should rather be caught by failing tests, and not this schema validation. Because, what if I really want a key called "foo"?

Happy to discuss this further, but I'd put it on hold for now

lczyk:
yep, let's talk more. i'll make it a draft for now. in general i agree it feels slightly awkward, but also i think this is the correct layer for this. not a hill i'll die on though ;)

vpa1977:
> AFAICT, the motivation for this PR stems from a bad `content` vs `contents`, and I would dare saying that this case should rather be caught by failing tests, and not this schema validation. Because, what if I really want a key called "foo"?

Yes, the other good option is to have a default test that validates the copyright installation for every slice.

PR #1188, opened 2026-09-10
chore(26.10): more libssl4

lczyk:
# Proposed changes

26.10 has ossl4 alongside ossl3 and most of the archive has been rebuilt against 4. This PR updates more slices to use ossl4.

A bunch were still deliberately open on ossl4 since the transition is in progress: https://transitions.ubuntu.com/html/openssl4.html#good,bad,partial,unknown,!notintesting

### Forward porting

n/a

## Checklist

### DIFF
--- a/slices/gnu-coreutils.yaml
+++ b/slices/gnu-coreutils.yaml
@@ -88,7 +88,7 @@ slices:
   summarizing-files:
     essential:
       libc6_libs:
-      libssl3t64_libs:
+      libssl4_libs:
     contents:
       /usr/bin/gnub2sum:
       /usr/bin/gnucksum:
@@ -108,7 +108,7 @@ slices:
     essential:
       gnu-coreutils_sort:
       libc6_libs:
-      libssl3t64_libs:
+      libssl4_libs:
     contents:
       /usr/bin/gnucomm:
       /usr/bin/gnuptx:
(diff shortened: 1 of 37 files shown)


lczyk:
blocked by:

- https://github.com/canonical/chisel-releases/pull/1189

but they kindof co-block one another, so this still can be reviewed and merged since 26.10 is devel

slices/gnu-coreutils.yaml
zhijie-yang, line 336: Include `libssl4_libs` here

slices/gnu-coreutils.yaml
```diff
@@ -108,7 +108,7 @@ slices:
     essential:
       gnu-coreutils_sort:
       libc6_libs:
-      libssl3t64_libs:
+      libssl4_libs:
```
zhijie-yang: Not introduced by this PR, but `libssl4` is a dependent of `/usr/bin/gnusort` instead of the bins here. Otherwise, the standalone installation of the `_sort` slice would fail to run because of a missing dynamically loaded library, although `libssl4` is not listed in the `ldd`.

zhijie-yang (changes requested):
One little fix, but otherwise good.

slices/gnu-coreutils.yaml
lczyk: good catch 👍

PR #1189, opened 2026-09-10
fix(26.10): systemd

lczyk:
# Proposed changes

fix and update systemd 261 in 26.10. a bunch of stuff moved to `systemd-tpm` and `udev` which also need updating but this PR is big as is, so out of scope for now.

slicing:

- added `minimal`: PID 1, `systemctl`, the target graph and the shutdown chain -- literally the bare minimal to boot. confirmed with tests.
- added `core`: ... + journal, the `sysusers` and `tmpfiles` appliers, `systemd-notify`, the network and time targets, and PID 1's own bus policy. this is what an ordinary daemon's unit assumes, confirmed with tests (incl. bus activation of a daemon from another package).
- updated `standard`: ... + the daemons, `run0`, generators and the remaining units.
- updated `dev`: ... + the kitchen sink.
- `journal`, `login`, `hostname`, `locale`, `timedate` and `network` carry their own units, bus policy and activation files, and no longer depend on aggregate `system-services`.
- `system-services` builds on `core` and pulls in the programs its units run. no path is listed in two slices any more (bar the bare `systemd-firstboot` in `dev`).
- `dbus-services` is dropped: PID 1's policy is in `core`, the daemons' in their own slices, the interface XMLs in `dev`.
- `tpm2` and `pcrlock` are dropped, their files moved to `systemd-tpm`, which is out of scope for this PR.
- library essentials are what each slice's binaries link: `libc6` and `libsystemd-shared`. `libsystemd-shared` links only `libc6` and dlopens the rest, so it is tiered by dlopen gate (`minimal`, `core`, `standard`, `dev`), each tier matching the systemd slice of that name. a slice whose programs open no gate takes the shared objects alone.
- added standalone `sysusers` and `tmpfiles` (binary alone, for image builds), `firstboot` (opt-in, its unit prompts on the console), `sysinstall` (opt-in, `system-install.target` hands the console to the installer), single-binary `ask-password`, `creds`, `machine-id-setup`, `mute-console` and `sysext`, and `libpam-modules-bin_unix-chkpwd`.

fixes:

- `system-services` enabled `ldconfig.service` and `systemd-machine-id-commit.service` without their binaries (203/EXEC on every boot), and ~40 more units whose binaries were in `service-handlers`, `sysusers`, `tmpfiles` or only in `dev` (the ask-password agents, and the ask-password, `creds` and `sysext` sockets).
- `system-install.target` was enabled in `standard` without `systemd-sysinstall`. it now lives in `sysinstall`.
- `run0` failed every call with 224/EXIT_PAM. it now pulls the "other" PAM stack and `unix_chkpwd`.
- getty units failed without `agetty` wherever a VT or console exists.
- the rescue and emergency shells never started: no `sulogin`, and no `/bin/sh` for it to fall back to.
- `systemd-firstboot.service` was enabled without its binary. the service file now lives in `firstboot`.
- the daemons could not take their bus names without `dbus-services`, and bus activation needs PID 1 on the bus.
- `systemd-creds` only reached `libcrypto` by accident; `creds` now pulls `libssl4` itself.
- dangling links in `service-handlers` and `polkit`.
- `run0` and `detect-virt` shipped openssl's `legacy.so` without `libcrypto`.

more tests:

- a variant for every slice with a binary. the tiers and daemons boot as PID 1 in their own pid, mount, uts and network namespaces, with the rootfs moved onto `/` the way switch_root does, and are driven through their own clients. greps read stdout only, so a binary that can't load fails the check. the suites run under `bash -eux`.
- `unit-programs` checks that every slice shipping units also ships the programs they run.
- `install-slices` makes each rootfs traversable, so daemons that drop privileges can chdir into it. (this should be cherrypicked to a separate PR...)

## Related issues/PRs

- unblocks #1188
- unblocks #1192

### Forward porting

- https://github.com/canonical/chisel-releases/pull/1189 **(this PR)**
- https://github.com/canonical/chisel-releases/pull/1194

## Checklist


slices/systemd.yaml
```diff
@@ -4,21 +4,90 @@ essential:
...
+  # What an ordinary daemon's unit file assumes on top of minimal. The manager's
+  # D-Bus registration and systemd-detect-virt are left out since no unit
+  # depends on either. Users can additionally cut systemd_dbus-services, or
+  # systemd_detect-virt to get these respectively.
+  core:
+    hint: Journal, boot setup and notify
+    essential:
+      libc6_libs:
+      libsystemd-shared_libs:
+      systemd_journal:
```
zhijie-yang, line 61: Maybe a comment over this line to indicate that keeping the `systemd_journal` is for the sake of forward compatibility.

slices/libpam-modules-bin.yaml
```diff
@@ -22,6 +22,16 @@ slices:
       /usr/sbin/unix_chkpwd:
       /usr/sbin/unix_update:
 
+  unix-chkpwd:
+    hint: Password verification helper for PAM
+    essential:
+      libaudit1_libs:
+      libc6_libs:
+      libcrypt1_libs:
+      libselinux1_libs: {arch: [armhf, riscv64]}  # linked only on these
```
zhijie-yang: This also applies to the arch i386. FF to reject this as we don't officially support i386 any more.

tests/spread/integration/systemd/helpers.sh
```diff
@@ -0,0 +1,93 @@
...
+  # onto / rather than a chroot: setns() back into this namespace resets root
+  # to its /, and machine-id-commit reads the id back through that
+  env -i container=lxc SYSTEMD_LOG_TARGET=console \
+    unshare --pid --uts --net --fork --mount-proc="$rootfs/proc" \
+    sh -c 'cd "$1" && mount --move . / && exec chroot . /usr/lib/systemd/systemd' sh "$rootfs" &
+  unshare_pid=$!
+
+  systemd_pid=""
+  for _ in $(seq 1 100); do
+    systemd_pid="$(pgrep -P "$unshare_pid" || true)"
```
zhijie-yang: Finding from an AI review session:

`pgrep -P` can discover the namespace child while it is still executing `sh`, before `mount --move`, `chroot`, and `exec systemd` complete. `nsystemctl` is-system-running can consequently use the worker root and `/run/systemd/private`; if the worker manager reports running, `boot_rootfs` returns before the nested manager has booted.
```suggestion
    systemd_pid="$(pgrep -P "$unshare_pid"  -x "systemd"|| true)"
```

slices/systemd.yaml
```diff
@@ -487,25 +585,18 @@ slices:
...
       /usr/lib/systemd/system/sysinit.target.wants/systemd-tmpfiles-setup-dev.service:
       /usr/lib/systemd/system/sysinit.target.wants/systemd-tmpfiles-setup.service:
-      /usr/lib/systemd/system/sysinit.target.wants/systemd-tpm2-setup-early.service:
-        arch: [amd64, arm64, armhf, riscv64]
-      /usr/lib/systemd/system/sysinit.target.wants/systemd-tpm2-setup.service:
-        arch: [amd64, arm64, armhf, riscv64]
       /usr/lib/systemd/system/sysinit.target.wants/systemd-update-done.service:
       /usr/lib/systemd/system/syslog.socket:
+      /usr/lib/systemd/system/system-install.target:
+      /usr/lib/systemd/system/system-install.target.wants/systemd-sysinstall.service:
```
zhijie-yang: Finding from an AI review session, amended with my comments:

- `systemd_standard` exposes an opt-in target that fails when activated because its executable (`/usr/bin/systemd-sysinstall`) is absent (only available ib the `_dev` slices).
- Preferred fix: create a dedicated `systemd_sysinstall` slice containing the binary, target, wants link, and service. Remove the three unit paths from `system-services`. Keep the bare binary in `_dev` if offline use is desired, matching the existing firstboot pattern.
- Audit whether the new slice needs `libsystemd-shared_standard` or `libsystemd-shared_dev`, then add a functional test that starts `system-install.target`.
This preserves standard-tier minimality while ensuring anyone selecting `sysinstall` gets a complete feature.

zhijie-yang (changes requested):
Thanks for the PR. Overall good with some findings.

slices/systemd.yaml
lczyk: yep. that's a good catch. addressing in progress...

slices/libpam-modules-bin.yaml
lczyk: might as well 🤷‍♀️

slices/systemd.yaml
lczyk: addressed. its now in its own slice and an essential of dev

lczyk:
@zhijie-yang as mentioned this morn: very astute comment which caused a small avalanche of better design. lmk what you think. [here](https://github.com/canonical/chisel-releases/pull/1189/files/04cdc823274d203310b6f0b0081f9a07270ff6db..17a78a7bfdea06060cd310ea7f07a995a12e3100) is the diff since your last review, for your convenience (although i think gh gives that to you anyway)

couple of things to point out:

- `core` now has the dbus policy files. it does not really matter for systemd deamons either way since they would come in through some other slice but, without these files in core, deamons from other packages would not be able to start without also cutting the slice with policy files, which felt a bit counterintuitive. there is a test for this in test_core.sh. lmk if you think they should be in a separate config-ish slice and ppl should just be told to cut it with core to get the external deamons.
- ive noticed a bunch of duplicated inclusions e.g. in standard, which would come in through core anyway, so a bunch of cleanup happened there
- as you suggested in [here](https://github.com/canonical/chisel-releases/pull/1189#discussion_r4062943561), the targets are redistributed correctly + there is now test_unit-programs.sh which checks that every slice with units ships what its supposed to.
- `service-handlers` can definitely be cleaned up more, but scope creep
- there's a bunch of lower importance TODOs for unsliced libs, but again scope creep. happy to add any/all if you wanna ofc.

tests/spread/integration/systemd/test_sysinstall.sh
```diff
@@ -0,0 +1,52 @@
...
+nsystemctl start system-install.target
+for _ in $(seq 1 20); do
+  state="$(nsystemctl show -p ActiveState --value systemd-sysinstall.service)"
+  [ "$state" = "failed" ] && break
+  sleep 0.5
+done
+test "$state" = "failed"
+# it ran, rather than failing to exec (203), and logind came up with it
+test "$(nsystemctl show -p ExecMainStatus --value systemd-sysinstall.service)" = "1"
+nsrun journalctl --no-pager -b -u systemd-sysinstall.service -o cat \
```
zhijie-yang: We are only asserting `sysinstall` to fail. Should we make it a step further to slice the `systemd-repart` and `bootctl` to make it fully functional?

IMO, this slices and this test is broken. Also it works in spread (exiting 0), but never in its real use case. I won't let you dig into the rabbit hole in this PR. Let's add a comment at the slice definition, indicating that this slices is broken, and create an issue in this repo to further slice the other two packages and deliver the full functionality.

lczyk: yeah, thats fair. that's not the right approach to something which we xfail. i've dome some speculative cuts and proposed a (partial) test instead, which i left commented out with a `TODO` note. also, issue for later: https://github.com/canonical/chisel-releases/issues/1225

Meulengracht (approved):
Thank you for this, these are really great improvements

zhijie-yang (approved):
Looks good now, thanks!

lczyk:
merging with 2 approvals 🥳

PR #992, opened 2026-04-15
feat(26.04): shorten `coreutils` slice names

lczyk:
# Proposed changes

this PR shortens the slice names for 2-character binaries from e.g. `ln-utility` to `ln-bin`. i have *not* done a full subslice of `coreutils` to keep this PR small + that can be done later.

suggested by @mohqas. we can only now do this w/out regressions though.

marking as 'priority' since this can only happen before 26.04 release.

## Related issues/PRs
n/a

### Forward porting
n/a

## Checklist

* [x] I have read the [contributing guidelines](
https://github.com/canonical/chisel-releases/blob/main/CONTRIBUTING.md)
* [x] I have tested my changes ([see how](https://github.com/canonical/chisel-releases/blob/main/CONTRIBUTING.md#7-test-your-slices-before-opening-a-pr))
* [x] I have already submitted the [CLA form](
https://ubuntu.com/legal/contributors/agreement)

## Additional context

slice names must be at least 3 characters. this means that some slices need to add some kind of suffix to pad the name. except for the cases in this PR, there are 3 more cases where `'slice =~ ".{1,2}-.*"'`: 

- `{"package": "golang-1.25-go", "slice": "go-env"}` -- actually fine as it, since it contains `go.env` file, not `go`
- `{"package": "libperl5.40", "slice": "io-modules"}` -- no standard naming convention for perl modules so not really something to do anything about here
- `{"package": "util-linux", "slice": "su-support"}` -- contains 2 different binaries so it's more of a function slice. if we subsliced *just* `su` into its own slice it would be `su-bin` but that's outside of the scope here and can be done later w/out regressions
- `{"package": "cpio", "slice": "mt-gnu"}` -- that's just what the binary is called

PR #1193, opened 2026-09-12
feat(26.10): subslice `coreutils`

lczyk:
# Proposed changes

finally gave in and once and for all subsliced all of coreutils. for now just 26.10, can backport later. the tests are now only per-binary. i've had to, somewhat humorously, quote "yes" and "false" slices so that yaml parsers in install-slices worfklows don't confuse them for booleans and thy to cut `gnu-coreutils_True` 😆

## Related issues/PRs

- #992 -- shortened the slice names, deferred the full subslice
- #1126 -- also adds `md5sum` and `seq` subslices, with `libssl3t64_libs`. overlaps with this one; keep this pr's versions (`libssl4_libs`) and drop those commits from #1126
- #1188 -- swaps `libssl3t64` -> `libssl4` in the same two `gnu-coreutils` group slices. this pr removes the lib essentials from the groups altogether, so keep this side on conflict

### Forward porting

n/a

## Checklist


lczyk:
spread tests: `(1/208)...` 😅

slices/coreutils-from-gnu.yaml
```diff
@@ -175,31 +155,26 @@ slices:
...
-      gnu-coreutils_changing-file-attributes:
 
   # File space usage -- These commands report how much storage is in use or
   # available, report other file and file status information, and write buffers
   # to file systems.
   # https://www.gnu.org/software/coreutils/manual/html_node/File-space-usage.html
   file-space-usage:
     essential:
+      coreutils-from-gnu_df-bin:
+      coreutils-from-gnu_du-bin:
```
cjdcordeiro, line 166: uf...the ugly slice names :p

what about `_disk-free` and `_disk-usage`, or `_dfree` and `_dusage`?

I'm just not a big fan of the implicit change of naming convention driven by a technical limitation. `-bin` sounds more conventional than the actual name of the slice IMO. But I don't have a perfect proposal here, so I'm not blocking it on this

cjdcordeiro (approved):
LGTM.

Left a comment about the `-bin` suffix, but it's more of a thought since I know this convention already exists in other slices.

I'm also a bit skeptical about the amount of tests being introduced here - 300+ files. I wonder if there's a way to simplify the amount of duplication throughout all these files (e.g. avoid the recurrent `rootfs=...` ; `chroot ...`). The obvious answer would be to turn the test scripts into command files (e.g. `chroot <path> $(cat testfile)`), but then it would be tricky to run those test scripts that actually have assertions within. So I think this is ok for now.

slices/coreutils-from-gnu.yaml
lczyk: mmm.... i'd keep this to just the bianry name for consistency. the user (slice developer) story here is roughly _"i want a bunch of binaries, let me pick and choose. i want mknod, ls, and du.  ah, i see, they are just the binary names with -bin suffix..."_

basically we'd either rename all the short slices to something else, or keep them as `-bin`. the mix is inconsistent and worst-of-both-worlds imo. in isolation `_dfree` or `_disk-free` do sound better, but taken as a whole i think they're surprising.

lczyk:
will wait on one more +1 on this since its a larger LoC-wise PR, but id argue to merge as is, and, if we dont like the `-bin` for short slices, do a followup in the next couple of weeks before 26.10 comes out.

lczyk:
oh, and yeah, tests are a bit annoying, but there isn't really much i can think of we could do about it, given we want to cut each slice on its own to prove we have all the deps in the cut.. 

btw, all 1248 coreutils tests cut in 15min using [spread-bread](https://github.com/lczyk/spread-bread) 😁🍞 : https://asciinema.org/a/WwY8Y2q2RIcPRS96 in no small part due to faster (as in, completely skipped) chisel validation of each cut. hopefully one day https://github.com/canonical/chisel/pull/302 ( and maybe even https://github.com/letFunny/chisel/pull/29 ) merge and we'll get faster cuts...

lczyk:
ok, ci for this is now impossible to complete on s390x without e.g. https://github.com/canonical/chisel/pull/302 or test batching ( 🤔 ... ). i have run all of these tests locally and i confirm they all work. merging with slightly red ci (as in, only s390x, only because it hits the 6h timeout)

PR #1194, opened 2026-09-14
fix(26.04): systemd

lczyk:
# Proposed changes

backport of #1189.

26.04 is a live release, so no a bunch of concessions had to be made. every change either adds a paths or leaves a slice's closure as it was. checked on all seven arches, i386 included. 

slicing:

- added `minimal`: PID 1, `systemctl`, the target graph and the shutdown chain -- literally the bare minimal to boot. confirmed with tests.
- added `core`: ... + journal, the `sysusers` and `tmpfiles` appliers, `systemd-notify`, the network and time targets. this is what an ordinary daemon's unit assumes, confirmed with tests. on 26.04 `journal` pulls in `system-services`, so `core` lists the journal's paths itself (TODO for future releases).
- `standard` now pulls `core`. its closure is unchanged.
- added standalone `sysusers` and `tmpfiles` (binary alone, for image builds), `firstboot` and `libpam-modules-bin_unix-chkpwd`. unlike 26.10, `system-services` still enables `systemd-firstboot.service`, as it can't lose the unit on a live branch. the boot tests mask it.
- hints for every slice.
- dropped library essentials from `standard`, `login` and `network` that none of their binaries link and that still arrive through `libsystemd-shared_libs`. closures unchanged.
- not backported, since each would remove paths: per-daemon units instead of the aggregate `system-services`, dropping `libapparmor1` / `liblzma5` / `libcap2` / `libsystemd0`, the `libsystemd-shared` reshuffle, and moving `tpm2` / `pcrlock` out.

fixes:

- `system-services` enabled `ldconfig.service` and `systemd-machine-id-commit.service` without their binaries (203/EXEC on every boot).
- `run0` failed every call with 224/EXIT_PAM. it now pulls the "other" PAM stack and `unix_chkpwd`.
- getty units failed without `agetty` wherever a VT or console exists.
- the tpm2 and pcr units in `system-services` ran binaries only `tpm2` ships. `tpm2` is now an essential of it.
- dangling links in `service-handlers` and `polkit`.
- `system-generators` and `user-services` shipped binaries without `libc6` and `libsystemd-shared`. also shipped the gpt-auto `/dev/null` mask and `770-nvpcr-separator.pcrlock`, which no slice had.
- 26.04 only: `systemd-modules-load` dlopens `libkmod` and failed on every boot where PID 1 holds `CAP_SYS_MODULE`. `service-handlers` now pulls `libkmod2`.

more tests:

- ported from #1189: a variant for every slice with a binary. the tiers and daemons boot as PID 1 in their own pid, mount, uts and network namespaces, with the rootfs moved onto `/` the way switch_root does, and are driven through their own clients. greps read stdout only, so a binary that can't load fails the check. the suites run under `bash -eux`.
- `install-slices` makes each rootfs traversable, so daemons that drop privileges can chdir into it. (same as in #1189, should be cherrypicked to a separate PR...)

### Forward porting

- https://github.com/canonical/chisel-releases/pull/1189
- https://github.com/canonical/chisel-releases/pull/1194 **(this PR)**

## Checklist

## Additional Context

You can use this to check that no cut slice has lost any contents:

```bash
#!/usr/bin/env -S bash -eu
# usage: [BASE=origin/ubuntu-26.04] [ARCH=amd64] ./check-lost.sh SLICE
SLICE="$1"
BASE="${BASE:-origin/ubuntu-26.04}"
HEAD_SHA="$(git rev-parse HEAD)"
WT="$(mktemp -d)"
trap 'chmod -R u+w "$WT"; git worktree remove --force "$WT"' EXIT

_cut() {
    mkdir "$WT/$1"
    chisel cut --release "$WT" --root "$WT/$1" ${ARCH:+--arch "$ARCH"} "$SLICE"
    (cd "$WT/$1" && find . | sort) > "$WT/$1.txt"
}

git worktree add -q --detach "$WT" "$(git merge-base "$BASE" "$HEAD_SHA")"
_cut base
git -C "$WT" checkout -q --detach "$HEAD_SHA"
_cut head

lost="$(comm -23 "$WT/base.txt" "$WT/head.txt")"
[ -z "$lost" ] || { printf 'lost from %s:\n%s\n' "$SLICE" "$lost" >&2; exit 1; }
printf '%s: nothing lost\n' "$SLICE"
```


slices/systemd.yaml
```diff
@@ -808,16 +980,19 @@ slices:
       /usr/share/polkit-1/actions/org.freedesktop.timedate1.policy:
       /usr/share/polkit-1/actions/org.freedesktop.timesync1.policy:
       /usr/share/polkit-1/rules.d/10-systemd-logind-root-ignore-inhibitors.rules:
+      /usr/share/polkit-1/rules.d/10-systemd-logind-root-ignore-inhibitors.rules.example:
```
zhijie-yang, line 983:
```suggestion
```

tests/spread/integration/libpam-modules-bin/test_unix_chkpwd.sh
```diff
@@ -0,0 +1,16 @@
...
+
+# an account database with one current and one expired account
+mkdir -p "$rootfs/etc"
+printf 'root:x:0:0:root:/root:/bin/sh\nexpired:x:1000:1000::/nonexistent:/bin/sh\n' > "$rootfs/etc/passwd"
+printf 'root:*:20000:0:99999:7:::\nexpired:*:1:0:99999:7::2:\n' > "$rootfs/etc/shadow"
+chmod 640 "$rootfs/etc/shadow"
+
+# pam_unix runs it without a terminal, naming the user and the check to make
+chroot "$rootfs" /usr/sbin/unix_chkpwd root chkexpiry < /dev/null
+if chroot "$rootfs" /usr/sbin/unix_chkpwd expired chkexpiry < /dev/null; then exit 1; fi
```
zhijie-yang, line 16: Maybe using an assertion on `$?` provides better readability here.

slices/systemd.yaml
lczyk: actually no! 😮 `.rules` is a symlink to this 🤯

lczyk:
going to convert this to draft since its lagging behind the 26.10 version that it's supposed to be a backport of. to be reopened after 26.10. i've made a temporary label to track this

